PR #56, opened 2025-03-15
.github: add a production build to CI

stevekuznetsov:
No idea why we're getting sharp installation errors even though we explicitly opt out of image optimization in Next config. Maybe this is a waste of time, I'll look into setting up Vercel builds instead.


stevekuznetsov:
blah, need to create a branch on the repo, not from a fork, to use secrets

PR #66, opened 2025-03-18
Upgrade payload to 3.28.1

busbyk:
Also includes:
- Removed the leading period in the README instructions. I believe that was unintentional based on the cp command above it but let me know if not!
- Latest versions of pnpm specify a packageManager field in package.json - this should keep us all on the same pnpm version which should be good for our local environments as well as CI. 

I believe pnpm will just pick up/install the new version based on the packageManager field but if it doesn't you should be able to:
```
corepack enable
corepack use pnpm@10.6.4
```


stevekuznetsov:
```
Error: Multiple versions of pnpm specified:
    - version [10](https://github.com/NWACus/web/actions/runs/13935422711/job/39002187956?pr=66#step:3:11) in the GitHub Action config with the key "version"
    - version pnpm@10.6.4 in the package.json with the key "packageManager"
  Remove one of these versions to avoid version mismatch errors like ERR_PNPM_BAD_PM_VERSION
```

Looks like editing the Github Actions YAML should make it pass, then we're good to merge. Rad!

busbyk:
> ```
> Error: Multiple versions of pnpm specified:
>     - version [10](https://github.com/NWACus/web/actions/runs/13935422711/job/39002187956?pr=66#step:3:11) in the GitHub Action config with the key "version"
>     - version pnpm@10.6.4 in the package.json with the key "packageManager"
>   Remove one of these versions to avoid version mismatch errors like ERR_PNPM_BAD_PM_VERSION
> ```
> 
> Looks like editing the Github Actions YAML should make it pass, then we're good to merge. Rad!

Fixed!

PR #68, opened 2025-03-19
Decision docs

busbyk:
I took a pass at writing up some decision docs based on what's in the repo, what I've learned, what you've told me.

@stevekuznetsov @yuliadub Please modify/add to what I've written here and/or add your own decision files.


busbyk:
Please ignore the other changes in this PR. I branched off of https://github.com/NWACus/web/pull/66 and this should be merged after that.

docs/decisions/003-observability-tools.md
```diff
@@ -0,0 +1,31 @@
...
+Sentry and PostHog are great tools for this and we already have accounts used by [Avy](https://github.com/NWACus/avy).
+
+Sentry:
+- Error monitoring
+- Performance monitoring
+
+PostHog:
+- Analytics
+
+___:
```
busbyk, line 26: Log drains do come with a cost on Vercel but it's nice to keep logs x number of days to help with production debugging. I've used Betterstack in the past (and for a few current projects). It's got a solid free tier. But we can evaluate some alternatives as well: DataDog, LogFlare. LogFlare has been hard to use for me in the past but they would be the cheaper option.

docs/decisions/004-rbac.md
busbyk, line 1: @stevekuznetsov Would love for some additional commentary here 😃

docs/decisions/004-rbac.md
```diff
@@ -0,0 +1,35 @@
...
+
+We want:
+- Super admin role: can see all content regardless of tenant
+- Flexible tenant-scoped roles: can perform any of create, read, update, delete, * on any of the collections
+- Users can have many roles
+
+We may want:
+- Users to be able to have roles across multiple tenants
+- Users to be able to have roles that only allow them to authenticate outside of the payload admin panel on a tenant's subdomain.
+  - Users would be able to have a single account across tenants
```
busbyk, line 21: This is out of scope for v1 but I put it here in case we want to take this into consideration as a future need now.

stevekuznetsov:
```
Error: Multiple versions of pnpm specified:
    - version [10](https://github.com/NWACus/web/actions/runs/13935422711/job/39002187956?pr=66#step:3:11) in the GitHub Action config with the key "version"
    - version pnpm@10.6.4 in the package.json with the key "packageManager"
  Remove one of these versions to avoid version mismatch errors like ERR_PNPM_BAD_PM_VERSION
```

Looks like editing the Github Actions YAML should make it pass, then we're good to merge. Rad!

docs/decisions/003-observability-tools.md
stevekuznetsov: Let's do some napkin math - if we're able to pre-render most things and serve from the CDN, are we generating any meaningful volume of logs daily?

docs/decisions/004-rbac.md
stevekuznetsov: I think you're right to call this out, and it might even be in for v1 - when we finalize the Payload contract we get SSO, so that should fulfill this need, right?

docs/decisions/004-rbac.md
```diff
@@ -0,0 +1,35 @@
...
+
+Status: proposed
+
+## Context
+
+Payload provides a very flexible approach to [access control](https://payloadcms.com/docs/access-control/overview). But we still need to design and implement RBAC based on our needs ourselves.
+
+Their multi-tenant plugin helps with access related to multi-tenancy but we still need to implement our custom RBAC requirements.
+
+We want:
```
stevekuznetsov, line 13: In addition to what you've got the other two big wins are:
- roles and bindings are just CMS content, so AC staff / our users can reconfigure these at their discretion and on their own time, so it's at best some consulting question-answering to support new use-cases in the future and not custom dev work
- per-tenant admins can manage their own fiefdom however they want, so there's no central bottleneck for user management

docs/decisions/004-rbac.md
```diff
@@ -0,0 +1,35 @@
...
+- Global roles
+- Tenant-scoped roles
+- New roles to be created with custom rules by super admins
+
+Collections:
+- roles
+- roleAssignments
+- globalRoleAssignments
+
+## Consequences
```
stevekuznetsov, line 46: Up-front complexity, and we're going to need to document this well so that non-technical folks can wrap their heads around the concepts and how to use them. Not sure if a custom UI on the CMS/admin side will be necessary or not.

docs/decisions/002-sqlite-turso.md
```diff
@@ -0,0 +1,21 @@
...
+
+Date: 2025-03-18
+
+Status: accepted
+
+## Context
+
+Argument for SQLite is very well explained [here](https://github.com/epicweb-dev/epic-stack/blob/main/docs/decisions/003-sqlite.md).
+
+We want local development to be easy. We want to be able to share data between developers easily. We want a database that can be used in a serverless environment since we're targeting Vercel. It would be great to support database branching.
```
stevekuznetsov, line 11: I need to look into this a bit more but the [embedded replicas](https://docs.turso.tech/features/embedded-replicas/introduction) feature would allow us to get some pretty sweet performance since our use-case is going to be 99.999999% reads as well.

stevekuznetsov: Ah, looks like it's not compatible with cloud functions. Using the Turso edge and co-lo the db with our render workers should be sufficient.

docs/decisions/003-observability-tools.md
yuliadub: unless we get some random error that every user hits? but I guess those cases are rare enough one offs since we are not dealing with things changing under us like we are in the app with NAC errors we have run into so maybe they dont matter

I do think it'd be helpful to really think about what we want logs for, i.e. specifically analytics. that will likely help drive some of these decisions. we brifly chatted about this but have not done any sort of deep dive

docs/decisions/004-rbac.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Context
+
+Payload provides a very flexible approach to [access control](https://payloadcms.com/docs/access-control/overview). But we still need to design and implement RBAC based on our needs ourselves.
+
+Their multi-tenant plugin helps with access related to multi-tenancy but we still need to implement our custom RBAC requirements.
+
+We want:
+- Super admin role: can see all content regardless of tenant
+- Flexible tenant-scoped roles: can perform any of create, read, update, delete, * on any of the collections
```
yuliadub, line 15: maybe obvious but just in case to clarify: collections within the specific tenant right?

docs/decisions/004-rbac.md
yuliadub: blanking here -- when would we need to have accounts across tenants? give me an example

docs/decisions/004-rbac.md
yuliadub: this is likely beyond the V1 scope, but what about such users that may only be say creating/adding events but are not necessarily entitled to have access to anything else?

docs/decisions/003-observability-tools.md
busbyk: Vercel will log all http traffic in addition to errors logged to the console. But for our log drain we can configure what [sources are sent](https://vercel.com/docs/log-drains/configure-log-drains#select-sources-to-collect-logs) and more importantly for pricing considerations we can [set a sampling rate](https://vercel.com/docs/log-drains/configure-log-drains#select-your-sampling-rate) to reduce logs volume sent to our log drain. 

Realistically though, log drain pricing on Vercel is reasonable. They charge in $10 increments. $10 per 20GB. I have a client that got about 600 page views per day last month and sent 193.42 MB worth of logs to our log drain. Do we have a sense for our daily traffic for nwac/sac/snfac? 

I agree we should define what we want for analytics. I typically just use logs for debugging production issues and setting up alerts for specific things (like I want to get a notification if an important endpoint ever returns a non-200 status). But we could do a lot with web analytics. Do we know what metrics some of the centers pay attention to now?

docs/decisions/004-rbac.md
busbyk: > I think you're right to call this out, and it might even be in for v1 - when we finalize the Payload contract we get SSO, so that should fulfill this need, right?

I'm curious what their SSO integration includes. We haven't gotten docs from them about their enterprise offerings right?

I'm thinking about two different types of users:
1. Users who get access to the admin panel. Single or multi tenant scoped. SSO would work for these users because they would be logging in with email domains for their center. 
2. Users who don't get access to the admin panel. We'd build a basic authenticated area for them in the app. I'm thinking that these are users who've donated or signed up for a membership or submitted an event and want to manage it. External users without center email address domains. Payload's SSO features might provide OAuth features to use for these users? Or we could build provider OAuth flows ourselves (like Google OAuth). 

> blanking here -- when would we need to have accounts across tenants? give me an example

I don't think there's necessarily a _need_ for accounts across tenants but because of the way multi tenancy is going to be set up with Payload it would be easy to build (at least that's what I'm thinking). There's a single user collection and each user has one or many tenants (or we could _just_ allow a single tenant per user if we want). Each user has one or many roles. 

OK the use case (thinking out loud so please question):

Let's assume we've built donation functionality something like this: users create an account to donate, they can then view their donations in an authenticated section of the app we've built. So when an external user account is created they are created in the users collection, with the tenant of the center their donating to, and a role like "external-user" or "donation-user" or something. 

If they then move a year later and now want to donate to a different center on our platform they go to sac's donation portal and try to create an account. It tells them there's already an account with their email associated with it since it's a single user collection behind the scenes.

It would be ideal to not throw that error, gracefully handle this scenario in the sign up flow where if a user already exists but they're trying to sign up for a new center we add that to their tenants array and let them log in. Then, in the donations portal (or external user portal) they can see that they donated to nwac but can now _also_ donate to sac (or sign up for a membership, manage membership, etc.). 

OK that was really long 😅 . But that's what I'm thinking about.

docs/decisions/004-rbac.md
busbyk: > In addition to what you've got the other two big wins are:
> 
> * roles and bindings are just CMS content, so AC staff / our users can reconfigure these at their discretion and on their own time, so it's at best some consulting question-answering to support new use-cases in the future and not custom dev work
> * per-tenant admins can manage their own fiefdom however they want, so there's no central bottleneck for user management

Excellent. Added those.

busbyk: > this is likely beyond the V1 scope, but what about such users that may only be say creating/adding events but are not necessarily entitled to have access to anything else?

They should see _only_ events in the admin panel when they log in and only be able to create events. There is also the external user portal option for users like that but if we're ok with them being aware of an admin panel controlling what they see via these role rules should work well.

docs/decisions/002-sqlite-turso.md
busbyk: > Ah, looks like it's not compatible with cloud functions. Using the Turso edge and co-lo the db with our render workers should be sufficient.

🙌 

I'm excited to use Turso. I've known about it for a while but haven't used it yet.

docs/decisions/004-rbac.md
busbyk: Yea the tenant-scoped roles will apply the tenant filter first and then check the rules for the collection and action. I'll clarify that!

busbyk: Added more detail: https://github.com/NWACus/web/pull/68/commits/015f38b04ba94b8d40c71bd7343627608cf14758

@stevekuznetsov can you verify?

docs/decisions/004-rbac.md
busbyk: Thanks! Added: https://github.com/NWACus/web/pull/68/commits/c0e1889fcf555e29acbb66580211092733644666

Yea a custom UI might be really helpful. Probably something we could include as a low priority and build if we have time towards the end of v1.

docs/decisions/004-rbac.md
stevekuznetsov: Great thoughts @busbyk - right now we're deferring all of the auth'ed user flows for the website. So, like you said, SSO for users on the admin panel is what we're after right now. That being said,

>  So when an external user account is created they are created in the users collection, with the tenant of the center their donating to

The reason that the role assignments are not in the User object was specifically to decouple them - so in the current scheme we would just assign them the donation role in multiple tenants and it "should work". That being said, there are so many open questions for how all this works so I am not confident we can speak to it yet - for instance, we might store all donation-related content in a CRM and simply link accounts somehow.

docs/decisions/004-rbac.md
stevekuznetsov: The only wrinkle with some third-party service provider adding events is that we would need to do some more thought on how to give andy@coolguides.biz access to events created by or owned by coolguides.biz only. Likely not access to things created only by Andy, but all guides from their company. Won't fit in the current RBAC scheme but we can do some thinking here.

docs/decisions/004-rbac.md
stevekuznetsov: Yep, that all looks good. Only change I would make is that it's not 'supersede' but that authz is deny-by-default and role assignments are only additive, so the moment we find any rule that allows something, we short-circuit since other rules can only add further approvals.

stevekuznetsov:
Feel free to merge in whenever you're ready! Thanks for doing this - all these look great.

docs/decisions/003-observability-tools.md
busbyk: Per: https://nwactechcommittee.slack.com/archives/C08JD81P6GZ/p1742421512640929

If we assume the logging in my client's Remix site is representative of what our logging will be (obv we need to verify this but for napkin math's sake), then 600 pageviews * 30 days = 18k/mo. 

193MB / 18k pageviews = 0.0107

100k pageviews * 0.0107MB = 1.07GB

There's probably a big variation in centers' site traffic but just 1.07GB * 21 centers = 22.47GB. 

That would be $20/mo because they charge $10/20GB in $10 increments. 

That seems fine. We can always adjust the sampling rate too.

stevekuznetsov: That sounds like the right order of magnitude :)

docs/decisions/004-rbac.md
busbyk: Updated, thanks!

PR #78, opened 2025-03-28
Set first user as global admin

### LINKED ISSUE #54: Assign First User Super-Admin Permissions
In the normal PayloadCMS bootstrapping flow, a "first user" is created at startup before seeding the database. Today, to do anything meaningful with the site after seeding, it's necessary to log out of the first user and into admin@avy.com. This is annoying - we should be able to query the Payload API for the first user and assign them the super admin role [just like we do](https://github.com/NWACus/web/blob/cf46ffd0ebdcf7e0e17018b60e18309fd99a4023/src/endpoints/seed/index.ts#L584) for admin@avy.com in the seed endpoint.

busbyk:
Resolves #54 

This PR does two things:
1. Assigns the first user created the Admin role as a global role assignment
2. Allows the seed file to be run multiple times

Previously, if you ran the seed file more than once errors would be thrown while creating palettes and themes because they weren't being deleted. Brands, themes, and palettes need to be deleted in that order because brands have a required themes relation, and themes have a required palettes relation. So those need to be deleted first. I think Payload was not deleting those collections if the brands/themes still existed and required those fields to not be null. 

**Testing instructions**
1. Create a fresh database
2. Visit `localhost:3000/admin` - should redirect to create first user form
3. Create a user - redirects to admin landing page, you should only see Redirects, Forms, Form Submissions, Search Results, Header, Footer, and Media collections
4. Click "Seed your database" - once it's done seeding the page should reload and you should see all collections


src/components/BeforeDashboard/SeedButton/index.tsx
```diff
@@ -47,6 +47,9 @@ export const SeedButton: React.FC = () => {
                   if (res.ok) {
                     resolve(true)
                     setSeeded(true)
+
+                    // Seed script may have adjusted currently logged in user's permissions
+                    window.location.reload()
```
busbyk, line 52: This seemed like the simplest (although not the most elegant) way of getting the user's permissions to update after the seed file modifies them.

Open to other suggestions.

stevekuznetsov: We're only going to use this in test deployments, so I'm ok with some hackery if we all are!

src/endpoints/seed/index.ts
```diff
@@ -636,6 +642,31 @@ export const innerSeed = async ({
     }
   }
 
+  payload.logger.info('- Creating global role assignment with first user and Admin role...')
+
+  const firstUserRes = await payload.find({
```
stevekuznetsov, line 650: Smart!

stevekuznetsov:
Do we need to [update the built-in seeded database](https://github.com/NWACus/web?tab=readme-ov-file#update-the-seed-database) with this change?

rchlfryn:
I am not sure if we should  force the redirect for `Visit localhost:3000/admin - should redirect to create first user form`.
I could see scenarios where we are working on the database, and every time we wipe it, making a new user could be annoying. Is this current functionality?

src/components/BeforeDashboard/SeedButton/index.tsx
busbyk: That was my thought 😇

busbyk:
> Do we need to [update the built-in seeded database](https://github.com/NWACus/web?tab=readme-ov-file#update-the-seed-database) with this change?

Good thought! I just tested with the built-in seed database ran into a unique constraint error which makes me think I have another this-needs-to-be-deleted-before-that problem. I gotta run out now but I'll take a look later to resolve!

stevekuznetsov:
>I could see scenarios where we are working on the database, and every time we wipe it, making a new user could be annoying. Is this current functionality?

Yeah, right now if you need to do a full wipe on the DB it's more than a little annoying to start from scratch, make the user, and seed it, however, I was hoping that this flow would be really rare.

We can now start the server using the "seeded" DB file in the repo, which short-circuits all of that, and if you make schema changes while everything is running, re-generating the Payload bits and using the Drizzle [push mode](https://orm.drizzle.team/docs/kit-overview#prototyping-with-db-push) should mean you can iterate [without wiping everything](https://payloadcms.com/docs/database/postgres#prototyping-in-development-mode).

Does that sound reasonable?

busbyk:
> I am not sure if we should force the redirect for `Visit localhost:3000/admin - should redirect to create first user form`. I could see scenarios where we are working on the database, and every time we wipe it, making a new user could be annoying. Is this current functionality?

The redirect to the create first user form is a payload cms behavior if there aren't any users in the database. But if there are already users in the database (i.e. you're using the built-in db or one you've been developing with locally) visiting /admin won't redirect to the create first user form. 

I think you could seed the db without manually creating a user by just running this file - the [seed file mentions this](https://github.com/NWACus/web/blob/f0ac0cf9283787b7afb9d1da35d6f59fd0f85012/src/endpoints/seed/index.ts#L51) in a comment from the website template.

busbyk:
> > Do we need to [update the built-in seeded database](https://github.com/NWACus/web?tab=readme-ov-file#update-the-seed-database) with this change?
> 
> Good thought! I just tested with the built-in seed database ran into a unique constraint error which makes me think I have another this-needs-to-be-deleted-before-that problem. I gotta run out now but I'll take a look later to resolve!

Sweet, just needed to add the navigation collections to the list of collections to clear before seeding to avoid unique constraints during seeding of those collections.

src/endpoints/seed/index.ts
busbyk: Thanks! And my thought was that this would work well for the seeded db file committed to git but there's a user with email t@t.t that was created before it. 

I suppose there's no harm in that user having the global admin role as well as the "Super Admin" user. Or I could delete that user and commit my changes as the dev.db.seeded file. That's probably cleanest. I assume you just had to create a first user when you first seeded the db?

I'll make that change just let me know if there was an intent for that user.

busbyk:
Two more thoughts while I'm at it on this PR:

1. Do you both get this change in the `src/app/(payload)/admin/importMap.js` file? I know @stevekuznetsov you said you've seen this. If we're all experiencing this, let's just commit this. If any of us aren't then it's worth digging into what formatter payload is using behind the scenes and try to wrangle it into using our editor config. 
image

2. I keep getting this alert from Chrome when I sign into admin@avy.com with the password "password" (which yanno makes sense). It's only minorly annoying but are any of you also experiencing this? If so, how about we change the seeded passwords to something that doesn't throw this up like "testpass" or "localpass" or whatever else one of us prefers?
Screenshot 2025-03-28 at 9 37 48 PM

stevekuznetsov:
1. Feel free to take whichever path you prefer, but I think it's likely just easier to run `pnpm prettify` before committing regardless of if the changes were from the generator or from your editor as opposed to trying to figure out how to get the generated output to match the `prettier` config a priori.
2. Sure thing - let's choose whatever is easiest.

src/endpoints/seed/index.ts
stevekuznetsov: The `t@t.t.` was the user I bootstrapped it with - as in, when starting Payload with nothing, that was the user created using the web flow it redirected me to. So wouldn't that be the one that we want to give this role to?

busbyk:
> 1. Feel free to take whichever path you prefer, but I think it's likely just easier to run `pnpm prettify` before committing regardless of if the changes were from the generator or from your editor as opposed to trying to figure out how to get the generated output to match the `prettier` config a priori.

Good call, this seems like the best approach. I added husky and lint-staged and setup a pre-commit hook that will format the staged files using pnpm prettify. This should ensure we're always committing formatted files. 

> 2. Sure thing - let's choose whatever is easiest.

Went with "localpass".

Also: Ope, new laptop, new unverified commits. Will rebase and sign those.

src/endpoints/seed/index.ts
busbyk: Yea I guess I was just thinking that we don't need to have that user in the seeded db in git right? Since we don't mention it in the README at all. 

I did delete it from the seeded db but can always revert that change if we want to keep that user in. 

I added a check to see if the first created user already has the global admin role so we don't get duplicate global role assignments. So now with the seeded db in git if you re-run the seed everything should be the same. It will have just gotten deleted and recreated and the user "Super Admin" will still just have one global role assignment (the admin role).

PR #80, opened 2025-04-03
Multi tenancy with Payload plugin + middleware and frontend routing updates

busbyk:
Took a hybrid approach here: using some things from the @payloadcms/plugin-multi-tenant but not using it's core functionality because:
1. It assumes we will control which tenants a user can access based on a tenants array field on the user collection but we're using tenant-scoped roles for this
2. I wanted more control over fields in the tenant field which they don't provide a way to override (this one is harder to argue but simple things like disableListColumn were defaulted to true and I want to be able to see tenants in list views). 

This PR:
- Uses TenantSelectionProvider from @payloadcms/plugin-multi-tenant
- Uses custom tenantField in each collection that needs tenancy
	- Uses TenantField client component from @payloadcms/plugin-multi-tenant
- Uses GlobalViewRedirect for "global collections" (i.e. collections where the tenant field is unique, one per tenant)
	- Redirects to the document view for these types of collections
- Uses a local plugin to add tenant field and base list filter to collections controlled by plugins (except for search)
- Updates search plugin to include tenant on search docs
- Updates frontend routing in middleware
	- Enforces use of either tenant subdomain or custom domain
	- Adds a main domain landing page that points to each tenant's subdomain landing page
	- Adds env var to determine if we prefer redirecting to subdomains or custom domains in production (always subdomains in dev)
- Removes unused center theme in src/app/(frontend)/[center]/layout.tsx


src/access/byGlobalRole.ts
```diff
@@ -13,7 +13,7 @@ export const byGlobalRole: (method: ruleMethod, collection: ruleCollection) => A
 
     payload.logger.debug(`evaluating access by ${user.id} for ${method} on ${collection}`)
 
-    const globalRoles = globalRolesForUser(payload.logger, user)
```
stevekuznetsov: Any specific reason the order of args changed?

src/access/byTenant.ts
```diff
@@ -37,7 +37,7 @@ export const byTenant: (method: ruleMethod, collection: ruleCollection) => Acces
       )
       .map((assignment) => assignment.tenant)
       .filter((tenant) => typeof tenant !== 'number') // captured in the getter
-      .map((tenant) => tenant.id)
+      .map((tenant) => tenant?.id)
```
stevekuznetsov: Can we improve the `filter` call so `matchingTenantIds` doesn't end up as `[](number | undefined)`?

src/app/(frontend)/[center]/[...segments]/page.tsx
```diff
@@ -35,7 +35,9 @@ export async function generateStaticParams() {
       payload.logger.error(`got number for page tenant: ${JSON.stringify(post.tenant)}`)
       continue
     }
-    params.push({ center: post.tenant.slug, segments: [post.slug] })
+    if (post.tenant) {
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/[...segments]/page.tsx:38]: I'm surprised that tenant is not a required field on a post - is that something wrong with the types coming in from the Payload tenancy plugin?

src/app/(frontend)/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { getServerSideURL } from '@/utilities/getURL'
+
+export const dynamic = 'force-static'
+export const revalidate = 600
+
+export const metadata: Metadata = {
+  title: 'AvyFx',
+}
+
+export default async function LandingPage() {
```
stevekuznetsov, line 16: Could we file a follow-up to add `generateStaticParams`?

src/collections/Media.ts
```diff
@@ -7,9 +7,10 @@ import {
 } from '@payloadcms/richtext-lexical'
 import path from 'path'
 import { fileURLToPath } from 'url'
-import { tenantField } from '@/fields/TenantField'
+// import { tenantField } from '@/fields/TenantField'
```
stevekuznetsov: nit: leftover comment

src/collections/Pages/index.ts
```diff
@@ -10,7 +10,6 @@ import { slugField } from '@/fields/slug'
 import { populatePublishedAt } from '@/hooks/populatePublishedAt'
 import { generatePreviewPath } from '@/utilities/generatePreviewPath'
 import { revalidateDelete, revalidatePage } from './hooks/revalidatePage'
-import { tenantField } from '@/fields/TenantField'
```
stevekuznetsov, line 13: As a follow-up, I can add https://www.npmjs.com/package/prettier-plugin-organize-imports so we don't get diffs on imports moving around

src/endpoints/seed/index.ts
```diff
@@ -889,21 +916,33 @@ export const innerSeed = async ({
...
-    data: JSON.parse(JSON.stringify(contactFormData)),
-  })
+  const contactForms: Record<string, any> = {}
+  for (const tenant of Object.values(tenants)) {
+    payload.logger.info(`Creating contact form for tenant ${tenant.name}...`)
+    const contactForm = await payload
+      .create({
+        collection: 'forms',
+        depth: 0,
+        data: JSON.parse(JSON.stringify({ ...contactFormData, tenant: tenant })),
```
stevekuznetsov: Why do we need `JSON.parse(JSON.stringify())` here as opposed to just passing in the object?

src/fields/TenantField/index.ts
```diff
@@ -1,33 +1,50 @@
...
+        }
+
+        return value
+      },
+    ],
   },
   index: true,
+  label: 'Avalanche Center',
   relationTo: 'tenants',
-  required: true,
```
stevekuznetsov, line 27: I think losing this line is what caused you to have to do so many `tenant?` calls.

src/utilities/rbac/globalRolesForUser.ts
```diff
@@ -1,8 +1,11 @@
 import { Role, User } from '@/payload-types'
 import { Logger } from 'pino'
 
-export const globalRolesForUser = (logger: Logger, user: User): Role[] => {
+export const globalRolesForUser = (user: User, payloadLogger?: Logger): Role[] => {
   const globalRoles: Role[] = []
+
+  const logger = payloadLogger ?? console
```
stevekuznetsov: I don't see a callsite for `gloablRolesForUser` that doesn't pass a logger - was this just used for testing?

busbyk:
Explanation for my last commit (https://github.com/NWACus/web/pull/80/commits/94a7149f067723ecf60a896d8624ed2b469fc3ff): https://www.loom.com/share/872f9648debf45a59e4732404dd2f35d?sid=864a43c9-4426-4ef9-9a71-a20f3c48a450

src/collections/Media.ts
busbyk: All for nitpicks - I hate commented out code. Thanks!

src/collections/Pages/index.ts
busbyk: All for it!

src/endpoints/seed/index.ts
busbyk: Hm, I don't think we do. Def a case of quickly copy and pasting what was there before. As you may or may not know `JSON.parse(JSON.stringify())` is a classic approach in JS to create a new version of an object so the original isn't mutated. But I don't think that matters here since I can't imagine the `payload.create()` is going to mutate the data passed to it. 

Looks like maybe this was a hack to avoid this typescript error? (assuming this is leftover from the website template): 
image

In the newer version of the website template they don't `JSON.parse(JSON.stringify())` and type the contactFormData better: https://github.com/payloadcms/payload/blob/e87521a3762040e7b8be0f53ecc0ce6c9554bc7d/templates/website/src/endpoints/seed/contact-form.ts#L3-L4

So I've updated to use the same approach: https://github.com/NWACus/web/pull/80/commits/beaa09efe8029801c24c1a67d197018431162c0b

src/fields/TenantField/index.ts
busbyk: ahh! you're so right. I'll add this back in and update accessing tenant accordingly.

src/access/byGlobalRole.ts
busbyk: Ah this was for the `@payloadcms/plugin-multi-tenant`'s `userHasAccessToAllTenants` (https://github.com/payloadcms/payload/blob/e87521a3762040e7b8be0f53ecc0ce6c9554bc7d/packages/plugin-multi-tenant/src/types.ts#L140) config option. I wanted to use globalRolesForUser in that function but apparently you can't access payload in a plugin because (I think) payload hasn't been fully initialized yet. So I reordered the params here so the logger could be optional and then added that nullish coalescer to fallback to console as the logger. 

But since we're not using that part of `@payloadcms/plugin-multi-tenant` anymore I can revert this!

src/endpoints/seed/index.ts
stevekuznetsov: Ah, gotcha. Look at the `git blame` for your link to the template ;)

src/utilities/rbac/globalRolesForUser.ts
busbyk: Reverted: https://github.com/NWACus/web/pull/80/commits/a028698385522ba5a695c782cf62734d28602b42

stevekuznetsov:
Wow, TIL loom.com. That's so cool!

src/endpoints/seed/index.ts
busbyk: Ha! Niiiiice. Past Steve comin' in clutch with the appropriate typing 🙌 

I still have it on my todo list to diff the website template changes and bring over as much as makes sense.

src/fields/titleFromTenantAndCollectionField/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    beforeChange: [
+      async ({ collection, data }) => {
+        if (!collection?.slug) return ''
+
+        let title = toWords(singular(collection.slug))
+
+        if (data?.tenant) {
+          if (typeof data.tenant === 'number') {
+            const payload = await getPayload({ config: configPromise })
+            const tenant = await payload.findByID({
```
stevekuznetsov: As you mentioned - this is fully derivative data, as in, we can compute it from other data in the database and would need to re-compute it in case the tenant's name changed, etc, so instead of saving it in the db we can have a client component for the admin panel compute it at runtime and just display it without saving back to a field on the collection. The approach you have here looks good as a first step though, no need to change now.

busbyk:
> Wow, TIL loom.com. That's so cool!

Oh it's a game changerrrrrr. So convenient.

src/fields/titleFromTenantAndCollectionField/index.ts
busbyk: Well I _think_ it does need to be saved to the database in order to be used as the title sadly 😞 . The useAsTitle field in admin options can't use virtual fields: https://payloadcms.com/docs/configuration/collections#admin-options:~:text=useAsTitle,be%20used%20as%20the%20title.

So unless we can completely replace whatever they're using for the title component in the document view? I was hoping I'd find something on that here but doesn't look like we can _just_ replace the title component?

src/app/(frontend)/page.tsx
busbyk: I don't think we need it here though since this path doesn't have any dynamic segments to it and this isn't scoped to a tenant so only a single version of this page should be statically generated. 

We're setting this to be statically generated using the `export const dynamic = 'force-static'`. Although I suppose we may want to include logic in a hook on the tenants collection to revalidate this path if any of tenant slug, name change. 

Am I thinking about this correctly?

src/fields/titleFromTenantAndCollectionField/index.ts
stevekuznetsov: Dang, major bummer. Here I was thinking we could customize it. Let's choose an expedient path for now, since I would like to put cycles in to the site for deployment and would appreciate having this PR merged. I'm not sure having derivative data in the DB and the edit UX here is ideal, but if that's what we do for now, that's fine. Or we can punt the feature for later, whichever you prefer.

stevekuznetsov: We can also send a PR to make it configurable upstream - looks like the path is here: https://github.com/payloadcms/payload/blob/e87521a3762040e7b8be0f53ecc0ce6c9554bc7d/packages/next/src/elements/DocumentHeader/index.tsx#L29

src/app/(frontend)/page.tsx
stevekuznetsov: Oh, DUH. Sorry - I misunderstood the Next-ism here, no props + that `export const dynamic` should do the trick! Sorry for the noise :)

.env.sample
```diff
@@ -1,14 +0,0 @@
-# This is the URI for the SQLite database to use behind PayloadCMS.
```
rchlfryn, line 1: These descriptions seem nicer than the ones in `.env.example`. Can we transfer them?

README.md
```diff
@@ -42,6 +42,16 @@ pnpm dev
...
 You will need to add two keys to your `.env` file. Reach out to someone on the dev team.
 
+### Set up subdomains on localhost
+
+In order to use tenant scoped subdomains on localhost you'll need to add the following to your `/etc/hosts` on macOS or `C:\Windows\System32\drivers\etc\hosts` on Windows:
+
+```
+127.0.0.1       nwac.localhost
+127.0.0.1       sac.localhost
+127.0.0.1       snfac.localhost
```
rchlfryn, line 52: 👏

src/Footer/RowLabel.tsx
```diff
@@ -1,9 +1,9 @@
 'use client'
-import { Header } from '@/payload-types'
+import { Footer } from '@/payload-types'
```
rchlfryn, line 2: Curious about this change here... Is removing the tenant dropdown?

src/app/(frontend)/[center]/layout.tsx
```diff
@@ -44,245 +37,18 @@ type PathArgs = {
   center: string
 }
 
-const zinc: Theme = {
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/layout.tsx:47]: 🎗️ Add a note that we are removing these to the description

src/fields/titleFromTenantAndCollectionField/index.ts
busbyk: Yea I was disappointed as well. I'm happy to remove this for now then. I like the approach of making it configurable in Payload via PR. How has your experience been with contributing? If you have any lessons learned I'd be all ears.

src/collections/Pages/index.ts
```diff
@@ -35,23 +36,43 @@ export const Pages: CollectionConfig<'pages'> = {
     defaultColumns: ['title', 'slug', 'updatedAt'],
     baseListFilter: filterByTenant,
     livePreview: {
-      url: ({ data, req }) => {
+      url: async ({ data, req, payload }) => {
+        let tenant = data.tenant
+
+        if (typeof tenant === 'number') {
```
rchlfryn, line 42: I've seen this check multiple times in this PR. I wonder if it would be beneficial to add a helper function to return the type and we just call that?

src/endpoints/seed/index.ts
```diff
@@ -889,21 +916,33 @@ export const innerSeed = async ({
...
-  payload.logger.info(`— Seeding contact form...`)
+  payload.logger.info(`— Seeding contact forms...`)
 
-  const contactForm = await payload.create({
-    collection: 'forms',
-    depth: 0,
-    data: JSON.parse(JSON.stringify(contactFormData)),
-  })
+  const contactForms: Record<string, any> = {}
+  for (const tenant of Object.values(tenants)) {
```
rchlfryn, line 923: Curious, what's the reason for overloading `contactForm`

src/fields/titleFromTenantAndCollectionField/index.ts
stevekuznetsov: They were pretty great to work with - very responsive on Discord to questions, etc. I only did fixes that were obviously useful, though, so I don't have visibility into proposing new functionality like making this part of the page configurable. When I needed to, I just bugged folks on the [dev channel](https://discord.com/channels/967097582721572934/1102950643259424828) to review things and they got around to it fairly quickly!

src/collections/Pages/index.ts
stevekuznetsov: Some background in case it's helpful - the Payload query types are not smart enough to honor the `depth` parameter you set, and if `depth` isn't deep enough the API returns related items by ID, instead of doing the full join. It makes operating with Payload kind of annoying and I was always on the fence - do we write code to crash and burn in cases where we know we should get an object, but have just gotten its ID, or should we write defensive code to go out and fetch the data if we get the ID, maybe emitting a log or metric or something that could be monitored to notice something isn't working as expected? I often chose to log + crash, but that's a bad user experience, but it should only ever happen if we use the wrong `depth` ... /shrug

Maybe some helpers like you mention would smooth over the DX.

Would love to hear your thoughts on it!

src/endpoints/seed/index.ts
```diff
@@ -889,21 +916,33 @@ export const innerSeed = async ({
...
 
-  payload.logger.info(`— Seeding contact form...`)
+  payload.logger.info(`— Seeding contact forms...`)
 
-  const contactForm = await payload.create({
-    collection: 'forms',
-    depth: 0,
-    data: JSON.parse(JSON.stringify(contactFormData)),
-  })
+  const contactForms: Record<string, any> = {}
```
stevekuznetsov: Can we type this `any`?

src/fields/titleFromTenantAndCollectionField/index.ts
busbyk: Sweet!

.env.sample
busbyk: Yea sure, updated ✅ 

I didn't bring over
```
# This is the Turso database authentication token. Only necessary for production or branched database access.
DATABASE_AUTH_TOKEN=xxx
```
though since we'll never need that locally but let me know if you think we should include that anyways.

src/Footer/RowLabel.tsx
busbyk: Nah, I just removed the Header collection because it wasn't being used for anything and I don't think we'll have a need for a global header since everything will be tenant scoped. But this RowLabel still needs the navItems type which also exists on the Footer so just swapping out where the type comes from.

src/app/(frontend)/[center]/layout.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/layout.tsx:47]: Updated description

src/collections/Pages/index.ts
busbyk: I think it would be nice to write a helper function for this. Something like `safelyAccessTenant` which fetches the tenant with the appropriate depth like what's done here. We could log when `typeof tenant === 'number'` for awareness of where this happens.

src/endpoints/seed/index.ts
busbyk: ??

We can totally type this correctly as Form if that's what you mean? `const contactForms: Record<string, Form> = {}`

src/endpoints/seed/index.ts
busbyk: Sorry, I'm not exactly sure what you mean? If you're asking why this is storing the contact forms in a dictionary key by the tenant name, each contact form is used below when creating a contact page and accessing by tenant.name is easy.

busbyk:
Wowww that was an annoying build issue. Turns out in git the directory was still named with an uppercase T but locally was lowercased:
Screenshot 2025-04-04 at 2 19 32 PM
Screenshot 2025-04-04 at 2 19 45 PM

src/endpoints/seed/index.ts
stevekuznetsov: Yeah that's what I meant!

stevekuznetsov (approved):

src/collections/Pages/index.ts
stevekuznetsov: A generic that knows how to go query every possible collection type would be awesome, since we hit this for everything, not just tenants. I think it should be possible, I can put some time into it.

stevekuznetsov:
Awesome, love to see it merge!

PR #84, opened 2025-04-06
Blocks: Add a list of images w/ text

### LINKED ISSUE #96: Blocks: ImageText w/ columns
Add a block for Payload admins that allows them to add content with an image, header, and blurb.

- Enable 3 ways to display (full bleed, columns w/ icons above, columns with icons below)

rchlfryn:
### Description
Adds a block to allow users to use the icon w/ content layout.

### Issue or ticket?
Resolves https://github.com/orgs/NWACus/projects/8?pane=issue&itemId=105961121&issue=NWACus%7Cweb%7C96
Still need to add 2 other layout versions in another PR

### Images
| Desktop | Tablet |
| --- | --- |
|Payload-Website-Template-04-06-2025_04_46_PM | Payload-Website-Template-04-06-2025_04_47_PM |


src/blocks/IconList/config.ts
```diff
@@ -0,0 +1,47 @@
...
+  FixedToolbarFeature,
+  InlineToolbarFeature,
+  lexicalEditor,
+} from '@payloadcms/richtext-lexical'
+
+const columnFields: Field[] = [
+  {
+    name: 'icon',
+    type: 'upload',
+    relationTo: 'media',
```
rchlfryn: Not sure if this is needed?

tailwind.config.mjs
```diff
@@ -10,10 +10,6 @@ export default {
   plugins: [require('tailwindcss-animate'), require('@tailwindcss/typography')],
   prefix: '',
   safelist: [
-    'lg:col-span-4',
```
stevekuznetsov, line 13: Did you have to remove these manually?

src/blocks/IconList/Component.tsx
```diff
@@ -0,0 +1,76 @@
...
+                  )}
+                >
+                  {(icon || staticImage) && (
+                    <Media
+                      imgClassName={cn('border border-border rounded-[0.8rem]', imgClassName)}
+                      resource={icon}
+                      src={staticImage}
+                    />
+                  )}
+                  {title && <div className={cn('mt-6')}>{title}</div>}
```
stevekuznetsov: For my understanding, it looks like the field config makes the icon and title required, but the component here is flexible enough to handle missing titles & icons. Is that just for future-proofing or is the "requirement" in the field config less strict than I was expecting?

stevekuznetsov:
Looks awesome!

src/blocks/IconList/config.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import {
+  FixedToolbarFeature,
+  InlineToolbarFeature,
+  lexicalEditor,
+} from '@payloadcms/richtext-lexical'
+
+const columnFields: Field[] = [
+  {
+    name: 'icon',
```
busbyk: We should probably allow for some way of constraining the size of the image:
image

Maybe providing width/height inputs would be sufficient. Maybe we want to provide some variants that correspond to aspect ratios or width percentages (i.e. 1/4, 1/2, 3/4). 

Might get a little complicated with validation here since we wouldn't want to let a user set an aspect ratio _and_ and width + height but doable.

src/blocks/IconList/config.ts
```diff
@@ -0,0 +1,47 @@
...
+    label: false,
+  },
+]
+
+export const IconList: Block = {
+  slug: 'iconList',
+  interfaceName: 'IconList',
+  fields: [
+    {
+      name: 'columns',
```
busbyk: We need to constrain the number of columns that can be added. Right now a user can add as many as they want and above 4 the lg:col-span-{} class isn't added since the columns <-> lg:col-span-{} class map only goes up to 4. 

image

I think you can add a maxRows option here: https://payloadcms.com/docs/fields/array#:~:text=maxRows,value%20is%20present.

src/blocks/IconList/config.ts
busbyk: It seems to be yea. I tried removing it and got a TS error and application error.

src/blocks/IconList/Component.tsx
busbyk: The admin UI does require it and enforces that when trying to publish. We could make icon and title optional though. Users might want that flexibility.

src/blocks/IconList/Component.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import { Media } from '@/components/Media'
+
+type Props = IconListProps & {
+  breakout?: boolean
+  captionClassName?: string
+  className?: string
+  enableGutter?: boolean
+  imgClassName?: string
+  staticImage?: StaticImageData
```
busbyk: Where do these come from? I don't see them passed to this component: 
image

tailwind.config.mjs
busbyk: I think these were only necessary because of this from the website template: https://github.com/NWACus/web/blob/49d2e1e7d05c52d7cf0f50cf7ec99a1524329b2e/src/blocks/Content/Component.tsx#L29

That style of classname constructions isn't detectable by Tailwind (https://v3.tailwindcss.com/docs/content-configuration#dynamic-class-names:~:text=the%20corresponding%20CSS%3A-,Don%E2%80%99t%20construct%20class%20names%20dynamically,-%3Cdiv%20class) so these had to be manually added. 

But Rachel's [approach](https://github.com/NWACus/web/blob/96366b7c36b388780e4cb439ee5a4c5428d42b9f/src/blocks/IconList/Component.tsx#L25) is because she's using full classnames. 

Might make sense to update https://github.com/NWACus/web/blob/49d2e1e7d05c52d7cf0f50cf7ec99a1524329b2e/src/blocks/Content/Component.tsx#L29 to use the same approach.

busbyk (changes requested):
In the Figma (https://www.figma.com/design/vAJkcFrHt5gddho3ILBOEx/NWAC-UX-UI-Redesign---Work?node-id=453-10694&t=nqQQyA7kcI2Pydr4-4) it looks like Chris had variations for row and column layouts for the image and title/richtext (top row in screenshot being horizontal and bottom row being column). Do we want to support those variations in this PR?

image

tailwind.config.mjs
stevekuznetsov: Gotcha, yeah - I roughly understood why we needed the change, but not if anything would tell me if I made other changes that impacted the `safelist`

rchlfryn: I have never used a `safelist`. I removed it because of what Kellen said. I think I am still confused about your last question. Did Kellen answer it or is there something more?

stevekuznetsov: Nope, nothing more - sorry for not being clear!

src/blocks/IconList/config.ts
rchlfryn: This works. Was going to look into this but thanks for doing the lift!

src/blocks/IconList/Component.tsx
rchlfryn: Copy pasta from `Media block`

src/blocks/IconList/Component.tsx
rchlfryn: Another copy/paste moment. I cleaned it up to make the text required. We can remove it in the future if need be.

src/blocks/IconList/config.ts
rchlfryn: @busbyk I was thinking we just put a `max-height` on the icon, and hopefully all of them will be a 1:1 ratio.  Maybe we can note this and revisit it?
Screenshot 2025-04-10 at 12 40 13

rchlfryn:
> In the Figma (https://www.figma.com/design/vAJkcFrHt5gddho3ILBOEx/NWAC-UX-UI-Redesign---Work?node-id=453-10694&t=nqQQyA7kcI2Pydr4-4) it looks like Chris had variations for row and column layouts for the image and title/richtext (top row in screenshot being horizontal and bottom row being column). Do we want to support those variations in this PR?
> 
> image

I will add this to another PR

src/app/(payload)/admin/importMap.js
```diff
@@ -1,98 +1,65 @@
+import { TenantField as TenantField_1d0591e3cf4f332c83a86da13a0de59a } from '@payloadcms/plugin-multi-tenant/client'
```
rchlfryn: 😞 Do we just always commit this?

stevekuznetsov: `pnpm prettify` will bring it down to the minimum set of changes - maybe you could actually just add a `&& pnpm prettify` to the end of `pnpm generate:importmap` or `pnpm generate:types` since it's always going to be needed?

src/blocks/ImageTextList/Component.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Media } from '@/components/Media'
+
+type Props = ImageTextListProps & {
+  className?: string
+  enableGutter?: boolean
+  imgClassName?: string
+  staticImage?: StaticImageData
+}
+
+export const ImageTextList: React.FC<Props> = (props) => {
```
stevekuznetsov: nit: while you had this PR open we decided to omit `React.FC` from the typing, so this would become

```ts
export const ImageTextList = (props: Props) => {
```

src/app/(payload)/admin/importMap.js
rchlfryn: We can do this in another PR. `prettier` removed the` importMap` file for now

src/blocks/ImageTextList/config.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+]
+
+export const ImageTextList: Block = {
+  slug: 'imageTextList',
+  interfaceName: 'ImageTextList',
+  fields: [
+    {
+      name: 'columns',
+      label: 'Icons with text',
```
busbyk: One last reference to icons

src/blocks/ImageTextList/Component.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import type { ImageTextList as ImageTextListProps } from '@/payload-types'
+
+import { Media } from '@/components/Media'
+
+type Props = ImageTextListProps & {
+  className?: string
+  enableGutter?: boolean
+  imgClassName?: string
+  staticImage?: StaticImageData
```
busbyk: I was really confused about when `staticImage` would be passed. I saw this in the MediaBlock component as well. But I think I was assuming this block would only be used in a blocks field in Payload. I'm guessing the staticImage prop is to allow use of this component on some custom page, not rendered by the RenderBlocks component, using a static import like `import sickPic from './sick-pic.png'`. 

Is that the intent here? I wonder if we want to support that for this component since it's not as likely we'd use this outside of RenderBlocks as a MediaBlock?

src/blocks/ImageTextList/Component.tsx
```diff
@@ -0,0 +1,71 @@
...
+              <div
+                className={cn(
+                  `col-span-4 md:col-span-2 ${colsSpanClass} ${lastOddElement && 'md:col-span-full'}`,
+                )}
+                key={index}
+              >
+                <div
+                  className={cn(
+                    {
+                      container: enableGutter,
```
busbyk: Throughout the rest of the app the container class seems to be used largely (but not exclusively) for responsive max width and horizontal padding for a section not for padding between elements in a section. 

For a section: 
image

But you're using it for padding between elements in this section:
Screenshot 2025-04-14 at 5 21 11 PM

which does work but it's resulting in extra padding here:
image

I wonder if it would make more sense to just use container on the top level div here (line 30) and use [gap](https://v3.tailwindcss.com/docs/gap) on the grid container. 

Right now there's no vertical gap between elements on mobile:
image

Which might be fine / what we want but worth calling out.

busbyk: There's no way to disable this gutter right now right? Do we want to add that? I would think we'd just want to disable the gutter for the top level div here on line 30 though.

src/blocks/ImageTextList/Component.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    <Media
+                      imgClassName={cn('h-[108px]', imgClassName)}
+                      resource={image}
+                      src={staticImage}
+                    />
+                  )}
+                  <div className={cn('mt-6')}>{title}</div>
+
+                  <div className={cn('mt-2')}>
+                    <RichText data={richText} enableGutter={false} />
```
busbyk, line 45: There's more space between the paragraphs rendered from the richtext component than between the title and the richtext content. That strikes me as odd - I usually expect a bigger gap between a title and content than between paragraphs. Maybe a question for Chris but figured I'd mention it:
image

busbyk (changes requested):

src/blocks/ImageTextList/Component.tsx
rchlfryn: After messing around with this a bit more, it looks like `MediaBlock` takes videos as well. I would assume the `staticImage` would be for a thumbnail of the video.

Either way, I want to remove it.

src/blocks/ImageTextList/Component.tsx
rchlfryn: I noticed this as well, but just left it to talk about it with Chris when I finish #101 . 
I will remove the `container` class between elements.

rchlfryn: This might be nice to be a broader discussion. I will make a note to chat with all of us.

src/blocks/ImageTextList/Component.tsx
rchlfryn: I will chat with Chris on this

busbyk (approved):
Looks good pending that "Icons with text" label change. 🎉

Issue #89, opened 2025-04-08
Display header and navigation on frontend


busbyk:
Displays header and navigation on the frontend based on the Navigation _global collection_

![Image](https://github.com/user-attachments/assets/6c07c2b5-72a6-4598-b3f4-ba55b8fbf2f0)

Figma: https://www.figma.com/design/vAJkcFrHt5gddho3ILBOEx/NWAC-UX-UI-Redesign---Work?node-id=474-30507&t=nqQQyA7kcI2Pydr4-4

Screenshot and link show Sierra's header but this will be dynamic for the center.

Issue #18, opened 2025-03-05
Unused Features Are Handled Gracefully


rustynwac:
Not all ACs will use all built-in functionality; if an AC has no blogs, or has not configured any "about" sections, the website should gracefully omit any navigation elements so the user is none the wiser. Eliminating the "blog" navigation header is better than showing an empty blog.

Blog and events top level nav item should be hidden if there are no blogs/events.

busbyk:
We still need to make a decision around whether or not we want a tenant to be able to explicitly disable blogs/events.

rustynwac:
Blog - nav hides if no published content
Events - always on. We dont need to solve for an AC that doesn't have events currently

PR #90, opened 2025-04-09
Navigation modeled in the admin panel

### LINKED ISSUE #88: Navigation configuration in Payload Admin Panel
- Create a single _global collection_ for Navigation
- Should have top level nav sections:
  - forecasts
  - weather
  - observations
  - education
  - accidents
  - blog
  - about
  - support
  - call to action (probably just a donate button or restrict to a single button that's customizable)
- Each section will have a config option to disable it

busbyk:
Resolves: https://github.com/NWACus/web/issues/88 

This is just the backend changes for managing navigation per tenant. 

- Removes existing navigation collections
- Removes existing navigation rendering from the `src/Header/Component.tsx` component temporarily before https://github.com/NWACus/web/issues/89 is completed based off this PR
- Adds a new navigation global collection (same slug as old navigations collection)
  - Uses tabs to represent the top level nav items
  - Configuration of a tab can be disabled so a user just sees the description of that tab for visual representation of the frontend nav
  - Configurable tabs have boolean options to:
    - Enable the tab (if disabled, won't appear in frontend nav)
    - Be clickable (if clickable, user can configure the link)
    - Have sub nav items (if false, no dropdown menu will render)
  - Has three levels of nav items
- Updates the seed file to account for the new structure


src/app/(frontend)/[center]/next/preview/route.ts
```diff
@@ -32,7 +32,7 @@ export async function GET(
       return new Response('No path provided', { status: 404 })
     }
 
-    if (!slug) {
+    if (slug === null) {
```
Comment (busbyk) [inline src/app/(frontend)/[center]/next/preview/route.ts:34]: Needed to modify this to support live preview on a center's landing page (i.e. "/")

src/app/(frontend)/[center]/next/preview/route.ts
```diff
@@ -60,28 +60,30 @@ export async function GET(
...
-        draft: true,
-        limit: 1,
-        // pagination: false reduces overhead if you don't need totalDocs
-        pagination: false,
-        depth: 0,
-        select: {},
-        where: {
-          slug: {
-            equals: slug,
+    if (slug !== '') {
```
Comment (busbyk) [inline src/app/(frontend)/[center]/next/preview/route.ts:62]: Also for having live preview on a center's landing page ("/")

src/collections/Navigations/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+      ],
+    },
+  ],
+  versions: {
+    drafts: {
+      autosave: {
+        interval: 100,
+      },
+    },
+    maxPerDoc: 10,
```
busbyk, line 92: I just set this to a low number. Pages uses 50. We could also just omit this to use Payload's default 100.

src/endpoints/seed/navigation.ts
busbyk, line 1: This isn't exactly matching the Navigation Content Analysis but I used it as a reference for some examples. Just tried to have at least one instance of each variation of the links, nav items, sub items, etc.

busbyk, line 1: This isn't exactly matching the Navigation Content Analysis but I used it as a reference for some examples. Just tried to have at least one instance of each variation of the links, nav items, sub items, etc.

src/utilities/generatePreviewPath.ts
```diff
@@ -23,7 +23,7 @@ export const generatePreviewPath = ({ collection, slug, tenant, req }: Props) =>
   // Only include tenant slug in path if not already on tenant's domain or subdomain
   const shouldIncludeTenantSlug = tenant?.slug && !isTenantDomain && !isTenantSubdomain
 
-  const path = `${shouldIncludeTenantSlug && tenant ? `/${tenant.slug}/` : ''}${collectionPrefixMap[collection]}/${slug}`
+  const path = `${shouldIncludeTenantSlug && tenant ? `/${tenant.slug}` : ''}${collectionPrefixMap[collection]}${slug ? `/${slug}` : ''}`
```
busbyk, line 26: This was including an extra `/`. I think it wasn't an issue when there was actually a slug but was throwing errors for a blank slug '' for live preview of the center's landing page.

src/collections/Navigations/fields/itemsField.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ArrayField } from 'payload'
+
+export const itemsField = ({
+  label,
+  description,
+  overrides = {},
+}: {
+  label: string
+  description?: string
+  overrides?: Record<string, unknown>
```
stevekuznetsov: Do we have more than one set of overrides, and we can we model them explicitly? Generic "deep merge anything to make any functionality you want" kind of APIs can make this hard to work with since it's impossible to reason about the outcome of the `deepMerge` - will the user overwrite the name, the type, etc?

src/collections/Navigations/fields/topLevelNavTab.ts
```diff
@@ -0,0 +1,77 @@
...
+          },
+        },
+        {
+          name: 'clickable',
+          type: 'checkbox' as const,
+          label: 'Is Clickable',
+          defaultValue: false,
+          admin: {
+            description:
+              "Controls whether this nav item itself is clickable or if it will just display it's sub nav on hover.",
```
stevekuznetsov: Do we have a use-case for this today? If not, I think I'd prefer holding off on adding it until we have something concrete in mind.

src/collections/Navigations/fields/topLevelNavTab.ts
```diff
@@ -0,0 +1,77 @@
...
+            description: 'Controls whether this shows up in the nav bar.',
+          },
+        },
+        {
+          name: 'hasNavItems',
+          type: 'checkbox' as const,
+          defaultValue: true,
+          admin: {
+            description:
+              'Controls whether this item will show a dropdown menu of sub nav items on hover.',
```
stevekuznetsov: Wouldn't `hasNavItems` be always equal to `topLevelNavTab.navItems.length > 0`? Do we want a user to turn this to false when they have items to show, or true, then they do not?

src/collections/Navigations/fields/topLevelNavTab.ts
```diff
@@ -0,0 +1,77 @@
...
+        hidden: !isConfigurable,
+      },
+      fields: [
+        {
+          name: 'enabled',
+          type: 'checkbox',
+          label: 'Is Enabled',
+          defaultValue: true,
+          admin: {
+            description: 'Controls whether this shows up in the nav bar.',
```
stevekuznetsov: Similar to my question on `hasNavItems` - when would a user turn this off if they are adding items?

src/utilities/generatePreviewPath.ts
stevekuznetsov: Might be a good candidate for unit tests since it's data in, string out, but /shrug

src/app/(frontend)/[center]/next/preview/route.ts
Comment (stevekuznetsov) [reply src/app/(frontend)/[center]/next/preview/route.ts:34]: It might be more appropriate to model this as something other than a query param for "slug" since it's more than a page slug

src/collections/Navigations/fields/itemsField.ts
```diff
@@ -0,0 +1,59 @@
...
+  deepMerge(
+    {
+      name: 'items',
+      type: 'array',
+      label,
+      admin: {
+        description,
+      },
+      fields: [
+        link({
```
stevekuznetsov: If it's not a lot of work, I think we likely want to only link out to pages, not posts, since posts are used for the blog? Or maybe we don't have this buttoned up well yet?

src/endpoints/seed/navigation.ts
```diff
@@ -0,0 +1,244 @@
...
+    },
+    events: {
+      options: {
+        enabled: true,
+        hasNavItems: false,
+        clickable: true,
+        link: pageLink({ url: '/events', label: 'Events' }),
+      },
+    },
+    donate: {
```
stevekuznetsov, line 155: Have we determined if we are going to put donate buttons (or other buttons) on the top nav and, if so, how that might impact our model here?

src/endpoints/seed/navigation.ts
```diff
@@ -0,0 +1,244 @@
...
+        clickable: true,
+        link: pageLink({ url: '/posts', label: 'Blog' }),
+      },
+    },
+    events: {
+      options: {
+        enabled: true,
+        hasNavItems: false,
+        clickable: true,
+        link: pageLink({ url: '/events', label: 'Events' }),
```
stevekuznetsov: Do we want to allow users to re-write the events landing page?

src/endpoints/seed/navigation.ts
```diff
@@ -0,0 +1,244 @@
...
+          },
+        },
+      ],
+    },
+    blog: {
+      options: {
+        enabled: true,
+        hasNavItems: false,
+        clickable: true,
+        link: pageLink({ url: '/posts', label: 'Blog' }),
```
stevekuznetsov: Do we want to allow users to re-write the blog landing page?

src/endpoints/seed/navigation.ts
```diff
@@ -0,0 +1,244 @@
...
+    },
+    weather: {
+      options: {
+        enabled: true,
+        hasNavItems: true,
+        clickable: false,
+      },
+      items: [
+        {
+          link: pageLink({ url: '/weather/stations/map', label: 'Weather Stations' }),
```
stevekuznetsov: What does `/weather/stations` do in this example? I somehow expected `/weather/stations/map` to require a parent page

src/collections/Navigations/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        topLevelNavTab({ name: 'accidents', label: 'Accidents' }),
+        topLevelNavTab({
+          name: 'blog',
+          label: 'Blog',
+        }),
+        topLevelNavTab({
+          name: 'events',
+          label: 'Events',
+        }),
+        topLevelNavTab({ name: 'about', label: 'About' }),
```
stevekuznetsov: How are we intending to model the relative order of the biography pages (which will not be pages but driven by some `Biography` collection likely) and the custom pages added to this by users?

src/collections/Navigations/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+        return path
+      },
+    },
+  },
+  fields: [
+    tenantField({ unique: true }),
+    {
+      type: 'tabs',
+      tabs: [
```
stevekuznetsov, line 47: How would we model a center adding a custom entry or two?

src/utilities/generatePreviewPath.ts
busbyk: I'd be down to put some time into getting unit tests set up if we feel like that's valuable?

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: Yea the blog tab is a good example of this https://github.com/NWACus/web/blob/66da370e767635af254f514f358e1595b2f6f214/src/endpoints/seed/navigation.ts#L219 where it's clickable (i.e. is a link not just a target to trigger the dropdown).

stevekuznetsov: Do we expect the user to make it *not* clickable or change the link underneath it though? It almost seems like the behavior of `/blog` is not driven by the content in the CMS.

src/utilities/generatePreviewPath.ts
stevekuznetsov: Let's carve it out as a follow-up, I'll get a GitHub Action set up to run Jest and then the lift to add a test here and there will be smaller.

src/collections/Navigations/fields/itemsField.ts
busbyk: By the user, you mean a dev implementing this itemsField field? 

In this PR I'm only passing admin.condition and admin.hidden through this functions params:
- https://github.com/NWACus/web/blob/66da370e767635af254f514f358e1595b2f6f214/src/collections/Navigations/fields/topLevelNavTab.ts#L61-L62
- https://github.com/NWACus/web/blob/66da370e767635af254f514f358e1595b2f6f214/src/collections/Navigations/fields/topLevelNavTab.ts#L72-L74

If you'd prefer I could just accept condition and hidden. 

I don't personally see the issue with allowing a dev to pass a partial of valid field config but I can be more specific if we think that's easier to understand for us. I did update the type here to be: `overrides?: Partial<ArrayField>`, fyi.

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: I'm operating under a few assumptions after our meeting last week and previous discussions:
- We want to build with full customizability in mind but disable certain functionality for most users. I.e. a center admin is probably not controlling this but we (as super admins) may want to. 
- Since we're building with full customizability in mind and there is some hesitation about the need for code changes for small things, I included a lot of options here. I'm assuming that we're going to generate these links (and a default set of pages, posts, and probably many other collections eventually) for each center when they are onboarded.

So it's unlikely that we want to change where the blog link points but if we already have to generate some nav items, why not have the blog and event links be configurable? That's my thinking anyways. 

Another use case for this is the about link. According to the Navigation Content Analysis, clicking the About top nav item should navigate to the About Us page: 
image

This is also used in the admin panel to show or hide the link to configure for the top nav item. If that link field is shown, it requires either the internal link or the custom url and the label. But if we want to leave these blank (i.e. not clickable, no link), we need to hide that field or make those fields optional. I felt like hiding the link field was a nicer UX. 
https://www.loom.com/share/dd3de96c31534ea3a79550a1179a395c?sid=3c4e9811-3aae-4468-9b71-c63486aa3a1f

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: Yea it would. This is just used in the admin panel to hide or show sub nav items: https://www.loom.com/share/bf1ef52cb9094e0099f46dd1c207e3c0?sid=614864f1-bdf3-49fc-87a4-23d716ab6df6

I suppose we could just always show sub nav items and not enforce that they have items. I can remove.

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: Perhaps when a user wants to disable a top nav item but not lose the sub nav items permanently. It's also easier to just uncheck a checkbox than deleting all of the sub nav items. And we could allow certain roles to control the enabled state without allowing them to be able to modify the sub nav items themselves. 

This more granular control for the enabled state is in the mindset of: https://github.com/NWACus/web/issues/18

But we could certainly accomplish that by just removing all of the sub nav items and checking to make sure the top nav item itself doesn't have a link.

src/collections/Navigations/fields/itemsField.ts
busbyk: I suppose I don't see the harm in allowing a nav item to point to a blog post. I wonder if we'll want to support nav items pointing to blog category pages, events, or another collection in the future too. Like Education > Classes > Avalanche Awareness Classes -- that would probably point to a route that uses an events collection to render content. 

My goal was to make this really flexible, operating under the assumptions that we are going to generate these nav items and we won't allow center admins to update these. But by building a really flexible system, we could make modifications or handle one-off scenarios for specific centers. So it accomplishes pigeon-holing centers into a common/suggested nav structure but still allows us as devs/super admins to have a flexible system "behind the scenes". That's why I was thinking about access control.

src/endpoints/seed/navigation.ts
busbyk: In the mockups there is a donate button. When discussing with Rusty and Chris they didn't want to entertain the idea of having multiple CTA buttons so my assumption here was that we would style the Donate top nav item as a button - just hardcode that. 

But if we need to add a top nav item, we'd need to make a code change. So that is a limitation here.

src/endpoints/seed/navigation.ts
busbyk: Again, operating under the "really flexible behind the scenes but predefined for center admins" approach.

src/endpoints/seed/navigation.ts
busbyk: Again, operating under the "really flexible behind the scenes but predefined for center admins" approach.

src/endpoints/seed/navigation.ts
busbyk: Navigates to {baseUrl}/weather/stations/map. I.e. `src/app/(frontend)/[center]/weather/stations/map/page.tsx`

image

No parent page required. The folder structure defines the path and a parent page or layout file isn't necessary. This was already existing from your POC though.

src/collections/Navigations/index.tsx
busbyk: I think they will be pages. Or could be. 

We talked about pre-generating pages rather than having a global collection for each custom/one-off page type. So we'd have a single page layout where you can use whatever blocks we've built out. So we might have a StaffProfiles block that would display a card for each staff member with their photo, bio, name, etc. 

But we could always add valid collections to be used as references in the link field to allow pointing to more than just pages and posts (see my other comment mentioning this).

src/collections/Navigations/index.tsx
busbyk: That's a limitation of using these tabs. Rusty and Chris seem to think we wouldn't ever allow adding a custom top level nav item. But we would need to make a code change and add another tab here to do that. So if we want to be able to support that we'll need to use a different approach.

src/endpoints/seed/navigation.ts
stevekuznetsov: I noted that we have an example of more than one CTA button already (Bridgeport) so we should be careful to make that limitation unless we have some idea that they'd be OK losing it - likely good thing for @rustynwac to chase. Your suggestion for supporting one hard-coded donate button (is it a field in this nav collection?) sounds like a good first step.

src/collections/Navigations/fields/itemsField.ts
stevekuznetsov: The way RBAC functions we're going to have to invent some really yucky abstractions around which fields in a collection someone can and cannot change if we want to model permissions that allow edits to some parts but not all parts of an item in a collection. Right now the abstraction is fairly simple - what is the name of the collection you're targetting and what are the verbs you want to support? To do something like "this field (or subfield) in this object can be edited" users writing roles will need to know the API structure of the collection to write the JSONPath (or whatever equivalent) to indicate their intent. Did you have something specific in mind to model permissions like that?

src/collections/Navigations/fields/topLevelNavTab.ts
stevekuznetsov: My comments here are based on a couple ideal that I think will serve us well:

1. code the thing we need now and add flexibility if/when we have a clear need for it
  a. I understand the super-flexible approach but in my experience it has worked out better to add the flexibility when we actually end up using it (this might be in a month, or two, so not forever) as opposed to trying to intuit what we might need now, since it's really hard to know a priori and I'd suspect things will need to change anyway from the current implementation as we play around with it and realize things need to look a little different than we originally anticipated.
2. "Normal forms" or "single sources of truth" in the database - storing a field that is logically derivative from other fields, or storing a field that is synced in from another DB, or storing a field (like the `"/blog"` path) that is not truly controlled by the DB, in the DB, can be fragile and lead to mistakes either on our end or the admin end looking at the edit panel a year from now.

---

>So it's unlikely that we want to change where the blog link points but if we already have to generate some nav items, why not have the blog and event links be configurable? That's my thinking anyways.

How would it work to make it configurable? I guess I'd claim it's not _unlikely_ but rather _impossible_! I imagine we will forever have a `src/app/(frontend)/[center]/blog/index.tsx` page that knows how to list the posts, show the landing page with cards, etc. Even giving the super-admin the illusion that they can change the path for this page in the CMS is not a win, since changing that will require that we move the actual file-based-routing page around, right?

---

> This is also used in the admin panel to show or hide the link to configure for the top nav item.

It seems like "is clickable" and "has nav items" are actually not two booleans but one enum - you could also model this as a "type: direct_link | link_container" toggle, since the two booleans approach would also need to have validation for XOR - it's not valid for the user to make it clickable *and* add nav items, I imagine.

The hard-coded "About Us" sub-item that requires the clickable behavior is interesting. I see an opportunity for simplification if the website doesn't have a strong opinion on what the title of that page is, right, since at that point we will just have an entry in the `navItems` array and we'd display the title however their page decided to have it? Maybe let's run this by Rusty, it will help if we can simplify.

src/collections/Navigations/fields/itemsField.ts
stevekuznetsov: Yeah, possibly just a personal style thing but in my experience a method that makes no opinions about the inputs or has no constraints on what a user does can be really hard to wrap your head around. Like, if you assume that there's some value in the default configuration that resides inside of `itemsField()` and some inter-dependencies between parts of the defaults, letting a user pass in _anything_ to the point of letting them re-configure the name of the field or the type of it abdicates your ability to reason about what an `itemsField()` really is or if the defaults you thought you were getting are even applied.

src/collections/Navigations/fields/topLevelNavTab.ts
stevekuznetsov: >Perhaps when a user wants to disable a top nav item but not lose the sub nav items permanently. 

It's an interesting use-case but I suspect it will not be used. I think if/when we see this as a need from the users we can add the option in fairly easily given the way you've factored this, so that would be my suggestion.

src/endpoints/seed/navigation.ts
stevekuznetsov: In the POC the path segments were built up from the navigation groups that the user put in, so you could only ever configure the level you were at:

```ts
    for (const item of nav.docs[0].weather_extra.value.items) {
      if (typeof item === 'object' && typeof item.value === 'object') {
        items.push({
          title: item.value.title,
          href: `/weather/${nav.docs[0].weather_extra.value.slug}/${item.value.slug}`,
          description: item.value.meta?.description || undefined,
        })
      }
    }
```

Letting the user configure the full path would allow in the example we're commenting on to add a page to the weather nav that does not nest under `/weather` or to have more than one page accidentally (via copy-paste etc) attempt to route to the same path, etc. 

What we've got now in this PR is a good place to start but we may want to consider a data model that does not allow for that kind of mistake from AC staff in the admin panel later on, or some validation to catch them.

src/collections/Navigations/index.tsx
stevekuznetsov: There are a few places in the site where we have the ability to write custom-tailored Collections, biographies are one. I feel pretty strongly that we'll want to do that instead of asking folks to click-and-edit a bunch of Blocks. For one, the onus should be on us to build a great biographies page and not on the AC staff to make sure that every block/card has the same formatting, etc. We also get the ability to link between author call-outs on blogs and forecasts with the individual author card in the staff page, and other wins that only happen if we have structured data. I think we can probably just put the staff page in some predetermined position in the "about us" nav items.

src/collections/Navigations/fields/itemsField.ts
busbyk: Yea it would be a bail out from our current system. 

I was imagining something where we restrict what the read and update verbs for tenant-scoped roles on this collection specifically would only give you permission to update the enabled state or something along those lines. 

But global roles would still apply like normal.

busbyk: https://github.com/NWACus/web/issues/97

src/endpoints/seed/navigation.ts
busbyk: Just to document in case we ever look back at this: Rusty and Chris seem adamant that we won't allow more than one CTA button so we're sticking with this single donate button for now.

src/fields/navLink/index.ts
```diff
@@ -0,0 +1,77 @@
...
+      name: 'label',
+      type: 'text',
+      admin: {
+        components: {
+          Description: '@/fields/navLink/components/Description#Description',
+        },
+      },
+      label: 'Label',
+      validate: validateLabel,
+    },
```
busbyk: @stevekuznetsov I ended up keeping label mostly for external urls but also for the one instance of the about top level nav pointing to a page called "About Us" but the label needing to be "About". Just added a dynamic description to explain that to the user.

busbyk:
This is ready for re-review.

src/fields/navLink/index.ts
stevekuznetsov: Sounds good! We can make a follow-up to check with Rusty & Chris to see if that part of the design is truly necessary or if we can pare it down for radical simplicity.

src/app/(frontend)/[center]/next/preview/route.ts
Comment (stevekuznetsov) [reply src/app/(frontend)/[center]/next/preview/route.ts:34]: @busbyk WDYT here? Maybe a follow-up or ignore if there is just one and only one non-slug thing we preview?

src/app/(frontend)/[center]/next/preview/route.ts
```diff
@@ -59,28 +59,30 @@ export async function GET(
...
+      // Verify the given slug exists
+      try {
+        const docs = await payload.find({
+          collection,
+          draft: true,
+          limit: 1,
+          // pagination: false reduces overhead if you don't need totalDocs
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/next/preview/route.ts:69]: TIL! You'd think `limit: 1` would imply `pagination: false`

src/collections/Navigations/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+        topLevelNavTab({
+          name: 'forecast',
+          label: 'Forecast',
```
stevekuznetsov: Why do we need a label and a name for the top-level tabs?

src/endpoints/seed/navigation.ts
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }
+
+    return {
+      type: 'internal',
+      reference: {
+        value: pages[tenant.name][slug].id,
+        relationTo: 'pages',
+      },
+      label: label || pages[tenant.name][slug].title,
```
stevekuznetsov, line 42: n.b. this implementation will mean that a re-title of a page will cause all nav links to be out of date. Leaving it empty and using that as a trigger to compute the nav label at runtime based on the title of the page would avoid this, but we can hold off until we have a response to #99

src/fields/navLink/components/Description.tsx
```diff
@@ -0,0 +1,17 @@
+'use client'
+
+import { useField } from '@payloadcms/ui'
+import { FieldDescriptionClientComponent } from 'payload'
+
+export const Description: FieldDescriptionClientComponent = ({ path }) => {
+  const linkPath = path.split('.').slice(0, -1).join('.')
+  const linkTypePath = [linkPath, 'type'].join('.')
```
stevekuznetsov: This is a bit of wizardry - could you add a comment for what the expected input (`path`) and the expected output (for `useField`) are to help describe what the transforms are doing?

```
// transform input paths like `` to their field identifier ``
```

src/utilities/isAbsoluteUrl.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Validates whether a string is an absolute URL with a protocol
+ */
+export default function isAbsoluteUrl(url?: string | null) {
+  if (url == null) return false
+
+  try {
```
stevekuznetsov, line 7: For my knowledge - I learned that `new URL` will throw on parsing errors, but if I hadn't, would there be a good way to know that a `try/catch` is needed here?

stevekuznetsov:
@busbyk looks great! Some tiny comments only

dev.db.seeded
rchlfryn, line 1: Heads up - I changed the seeded db in #84 as well.

src/fields/link.ts
```diff
@@ -15,14 +15,14 @@ export const appearanceOptions: Record<LinkAppearances, { label: string; value:
   },
 }
 
-type LinkType = (options?: {
+export type LinkType = (options?: {
```
rchlfryn: is the `export` needed here?

src/payload.config.ts
```diff
@@ -56,7 +54,7 @@ export default buildConfig({
         {
           path: '@payloadcms/plugin-multi-tenant/rsc#GlobalViewRedirect',
           serverProps: {
-            globalSlugs: ['settings', 'navs', 'navigations', 'brands'],
+            globalSlugs: ['settings', 'brands', 'navigations'],
```
rchlfryn, line 57: ⛏️  do we want alphabetical sorting here?

src/collections/Navigations/fields/itemsField.ts
```diff
@@ -0,0 +1,33 @@
+import { navLink } from '@/fields/navLink'
+import merge from 'lodash.merge'
+import { ArrayField } from 'payload'
+
+export const itemsField = ({
```
rchlfryn, line 5: ⛏️  I wonder if `items` is the right name here. I put it as a nit pick because I don't know a better name. SubNav? link? I guess it is a menu item... maybe make it more descriptive to `navItem`

rchlfryn (approved):
Huge lift - great stuff. I mainly commented nit picky things (⛏️ ). feel free to respond or ignore those.

src/app/(frontend)/[center]/next/preview/route.ts
Comment (busbyk) [reply src/app/(frontend)/[center]/next/preview/route.ts:34]: I think what I want to do is update the validation here so that:
- path is always required
- if one of collection or slug is passed, the other is required too

This would handle previewing pages based on a document in a collection and it would handle previewing pages that aren't based off of collections (like a forecast page or a weather page). 

I think using slug makes sense - it will never be anything different than a slug or null. My goal with the change in this PR was essentially to ignore the page lookup and just redirect to the path with draft mode turned on. So updating the validation in this way makes sense to me. 

I created a separate task to do that under and will just leave this as it is for this PR: https://github.com/NWACus/web/issues/113

src/app/(frontend)/[center]/next/preview/route.ts
Comment (busbyk) [reply src/app/(frontend)/[center]/next/preview/route.ts:69]: Yea you would! What's interesting is that `pagination: false` actually still results in all of the pagination-related fields being returned. I think it's just that pagination related queries don't happen? 

So if I pass `pagination: faslse` I get:
```
{
  "docs": [
    {
      "id": 39
    }
  ],
  "hasNextPage": false,
  "hasPrevPage": false,
  "limit": 1,
  "nextPage": null,
  "page": 1,
  "pagingCounter": 1,
  "prevPage": null,
  "totalDocs": 1,
  "totalPages": 1
}
```

but if I don't pass `pagination: false` (which defaults to `pagination: true`) I get:
```
{
  "docs": [
    {
      "id": 39
    }
  ],
  "hasNextPage": true,
  "hasPrevPage": false,
  "limit": 1,
  "nextPage": 2,
  "page": 1,
  "pagingCounter": 1,
  "prevPage": null,
  "totalDocs": 3,
  "totalPages": 3
}
```

I guess it's nice in some ways that the same attributes are always returned in the .find response. It's essentially a findMandy query so regardless of limiting to a single result it's still returning a findMany result.

src/endpoints/seed/navigation.ts
busbyk: Well this is just a helper function for seeding which needs to support passing a label for external links. But if you're point is that it's confusing I could break this out into two different helper functions: one for creating internal links and one for external links?

Is that what you're thinking?

src/collections/Navigations/index.tsx
busbyk: Oh good point. We can just use name and the payload `toWords` util fn. I also can use virtual fields for the non-configurable tabs so we don't store those in the db. Those updates: https://github.com/NWACus/web/pull/90/commits/7b1dd6bf12297b0429deae954efcb69bf1968cb6

src/fields/navLink/components/Description.tsx
busbyk: Good call, this was confusing. 

I updated this to be simpler and more explicitly for use only by a nav link's label field. 

https://github.com/NWACus/web/pull/90/commits/e2a0c9ce62a5431a49eeac5bb5e96c0357496a5f

src/endpoints/seed/navigation.ts
stevekuznetsov: I wasn't talking about the seed, I was saying that having the label precomputed as a separate field in the database would lead to desync issues. We can hold off changing what you have in this PR until #99 is figured out

src/utilities/isAbsoluteUrl.ts
busbyk: You're not going to like the answer 😅 . There's no way to know that without checking the documentation or having prior knowledge. TS won't tell you and intellisense probably won't tell you (unless it was called out specifically in the jsdoc comment). 🤷

dev.db.seeded
busbyk: This brings up a worth discussion: we can't really diff the seeded db so what's our practice for merging those changes? 

My assumption was that the seeded db should always match the output of running the seed process on the existing seeded db. Does that hold true for your changes in #84 ?

If I merged your changes in and then created a new db from the seeded db could I just run the seed process and still have all the changes you made in that db?

If that's true, do we even need to be committing a seeded db to git? 🤔

src/payload.config.ts
busbyk: If that's a norm we want to commit to as a team I'm down for that. I usually don't worry too much about sorting. Do you use a VS Code plugin to alphabetically sort array constants for you?

busbyk: Is the idea that it's easier to determine what changed at a glance in diffs if it's sorted alphabetically?

src/fields/link.ts
busbyk: Good call! I'd been using it in another file before I broke out nav links into a separate component. Removed.

src/collections/Navigations/fields/itemsField.ts
busbyk: It could be navItemsField if we want to make it more explicitly for the navigation. I was thinking that the nesting of it under `src/collections/Navigations/fields` would be enough to make it clear that it's just for the navigations collection but it does come up in the search files prompt as a generic items field which could be confusing:
image

I used plural "items" because it's an array field. We could go with "navItemsField" and use name "navItems" to be more descriptive?

Here's an example API response with the current naming for reference: 
[example-navigations-response.json](https://github.com/user-attachments/files/19743633/example-navigations-response.json)

src/utilities/isAbsoluteUrl.ts
stevekuznetsov: Good thing you have the experience then :D

dev.db.seeded
stevekuznetsov: I think we require PRs to be rebased on latest `main` to merge, so that should mean that if we add in CI to detect when the current seeded content does not match what would be committed if a seed were ran from `HEAD`, we should be OK. 

We were committing the seeded database to git in order to:

1. continue to make incremental dev work as easy and fast as possible (but Rachel let me know that my expectation for dev work taking microseconds was unrealistic for Next, and the existing seed endpoint was "fast enough")
2. make `next build` not require anything for our CI

I am open to other ideas! If this is more of a hassle than it's worth, no need to keep it.

stevekuznetsov (approved):

stevekuznetsov:
LGTM! Leaving in case @rchlfryn wants to do another round of review, otherwise let's merge it!

dev.db.seeded
busbyk: Yea I haven't been using the seed db very much at all. I usually just: `rm dev.db` `pnpm dev` and run the seed process. 

I'd be down to remove it.

We might want it for the build if we start running tests though? Although I suppose we could just run the seed script in CI. That would make CI slower.

busbyk:
> Huge lift - great stuff. I mainly commented nit picky things (⛏️ ). feel free to respond or ignore those.

I'm all for nitpicks btw so don't be afraid to keep making comments like these 🙏

dev.db.seeded
stevekuznetsov: I'll follow up on this

rchlfryn: My plan was to pull from `main` and reseed, but that might get missed if there are two PRs that update it.

I am confused how we can remove it and have the tests still pass. My mental model of how this file works is failing. I'll bring this up at stand.

PR #92, opened 2025-04-10
prettier: organize imports

### LINKED ISSUE #81: As a follow-up, I can add https://www.npmjs.com/package/prettier-plugin-organize-imports so we don't get diffs on imports moving around
As a follow-up, I can add https://www.npmjs.com/package/prettier-plugin-organize-imports so we don't get diffs on imports moving around

_Originally posted by @stevekuznetsov in https://github.com/NWACus/web/pull/80#discussion_r2026983490_

stevekuznetsov:
Looks like we already had the plugin in package.json, just did not turn it on from the config.

Fixes https://github.com/NWACus/web/issues/81


stevekuznetsov:
Keep forgetting we need to make PRs from the repo ...

PR #95, opened 2025-04-10
tsconfig.json: enable strict mode for tsc

stevekuznetsov:
tsconfig.json: enable strict mode for tsc

---

src/utilities: replace deepMerge with lodash

Our implementation runs afoul of the type checker in many ways, it's not
clear that we need to have our own implementation over what `_.merge()`
provides.

---

update gitignore for ts build info

---

*: fix simple mistakes in types

Either missing types, or incorrect or imprecise types.

---

*: remove plural/singular copy-pasta

This doesn't seem like something we should even have in the code since
these are configurable at the collection config level.

---

*: ignore tsc errors for complex Payload-isms

We likely want to fix these in the future, but it's a complex and
unintuitive set of code right now. Payload authors don't like
switch/case statements, so they do this dynamic assignment business
often and have no way to prove that the types work out correctly.

---

collections/Pages: fix a hook

Type checker caught that this hook was bound to the wrong place.

---

*: run prettier

---


src/fields/link.ts
```diff
@@ -135,5 +135,5 @@ export const link: LinkType = ({ appearances, disableLabel = false, overrides =
     })
   }
 
-  return deepMerge(linkResult, overrides)
```
busbyk, line 138: If this is all that lodash is being used for we could just install this one function: https://www.npmjs.com/package/lodash.merge

I don't believe Lodash is tree-shakeable so the entire package would be included in our bundles (we're statically generated pages yes but we still send javascript to hydrate the page). Apparently there is an alternative `lodash-es` package but importing the single function packages is a nice approach imo.

src/components/PageRange/index.tsx
busbyk, line 1: You could simplify this to this now that it's not using defaults:

```
import React from 'react'

export const PageRange: React.FC<{
  className?: string
  collectionLabels: {
    plural?: string
    singular?: string
  }
  currentPage?: number
  limit?: number
  totalDocs?: number
}> = ({ className, collectionLabels, currentPage, limit, totalDocs }) => {
  let indexStart = (currentPage ? currentPage - 1 : 1) * (limit || 1) + 1
  if (totalDocs && indexStart > totalDocs) indexStart = 0

  let indexEnd = (currentPage || 1) * (limit || 1)
  if (totalDocs && indexEnd > totalDocs) indexEnd = totalDocs

  const { plural, singular } = collectionLabels

  return (
    <div className={[className, 'font-semibold'].filter(Boolean).join(' ')}>
      {(typeof totalDocs === 'undefined' || totalDocs === 0) && 'Search produced no results.'}
      {typeof totalDocs !== 'undefined' &&
        totalDocs > 0 &&
        `Showing ${indexStart}${indexStart > 0 ? ` - ${indexEnd}` : ''} of ${totalDocs} ${
          totalDocs > 1 ? plural : singular
        }`}
    </div>
  )
}
```

src/components/TenantSelector/index.tsx
```diff
@@ -8,7 +8,13 @@ import React from 'react'
...
 import './index.scss'
 
-const TenantSelector: React.FC = ({ label, viewType }: { label: string; viewType?: ViewTypes }) => {
+const TenantSelector: React.FC<{ label: string; viewType?: ViewTypes }> = ({
+  label,
+  viewType,
+}: {
+  label: string
+  viewType?: ViewTypes
+}) => {
```
busbyk: There seems to be quite the discussion about the use of React.FC<{...prop types}> vs. just typing the function arguments ({ ...props}: { ...prop types}) => {}. 

In this case we only need one of these approaches here.

So:
```
const TenantSelector = ({
  label,
  viewType,
}: {
  label: string
  viewType?: ViewTypes
}) => {
```
or 
```
const TenantSelector: React.FC<{ label: string; viewType?: ViewTypes }> = ({
  label,
  viewType,
}) => {
```

I've always preferred just typing the props. I think it's cleaner. React.FC<> implicitly adds the children prop but I'd rather just add that myself _if_ the component actually accepts/uses children. 

There is _a lot_ of React.FC usage from the website template though - the Payload team seems to prefer it. The React Typescript docs use function props without React.FC: https://react.dev/learn/typescript#. 

It def seems like the majority of the community have moved away from using React.FC. It's not clear to me why the Payload team prefers it.

src/endpoints/seed/index.ts
```diff
@@ -702,7 +702,7 @@ export const innerSeed = async ({
         },
       })
     } catch (err) {
-      payload.logger.error(`Failed assign global admin role to first user, error: ${err.message}`)
+      payload.logger.error(`Failed assign global admin role to first user, error: ${err}`)
```
busbyk: err gets typed as unknown in catch blocks but it will often still be an error instance in which case we would want to be logging the message attribute. 

You can use: `payload.logger.error(`Failed assign global admin role to first user, error: ${err instanceof Error ? err.message : err}`)`

src/components/AdminBar/index.tsx
```diff
@@ -62,8 +51,8 @@ export const AdminBar: React.FC<{
           cmsURL={getClientSideURL()}
           collection={collection}
           collectionLabels={{
-            plural: collectionLabels[collection]?.plural || 'Pages',
-            singular: collectionLabels[collection]?.singular || 'Page',
+            plural: collection + 's',
```
busbyk: We have a package installed called `pluralize` we could use for this `pluralize(collection)` to be a little safer. I think I may have installed it an ended up not using it 😅

src/fields/link.ts
stevekuznetsov: Weirdly I think we already have it (transitively, somehow) - here's current `main`: https://github.com/NWACus/web/blob/c5c6eaa2fa737cd91c0adfbcb4d9fab56e6ed177/pnpm-lock.yaml#L5841

Maybe I need to import it correctly?

src/components/TenantSelector/index.tsx
stevekuznetsov: Thank you for the context! The error that I was getting with only having the props typed was:

```
src/components/TenantSelector/index.tsx:11:7 - error TS2322: Type '({ label, viewType }: { label: string; viewType?: ViewTypes; }) => React.JSX.Element | null' is not assignable to type 'FC<{}>'.
  Types of parameters '__0' and 'props' are incompatible.
    Property 'label' is missing in type '{}' but required in type '{ label: string; viewType?: ViewTypes | undefined; }'.

11 const TenantSelector: React.FC = ({ label, viewType }: { label: string; viewType?: ViewTypes }) => {
         ~~~~~~~~~~~~~~

  src/components/TenantSelector/index.tsx:11:58
    11 const TenantSelector: React.FC = ({ label, viewType }: { label: string; viewType?: ViewTypes }) => {
                                                                ~~~~~
    'label' is declared here.
```

So I really like your suggestions for simplification, but I think we can't only type the props and keep React.FC. Would you rather I remove the explicit type from the props, or remove React.FC?

stevekuznetsov: (Seems like we should stay away from `React.FC` but just want to make sure)

src/fields/link.ts
busbyk: Yea it's a peer dependency of eslint. If we're going to use it I think it'd be nice to include it as a dependency in our package.json explicitly.

src/components/TenantSelector/index.tsx
busbyk: Yea we'd completely remove React.FC. It's used in quite a few places throughout the codebase though so might be a good idea to update all of those and make a point that we all just type the props going forward instead of using React.FC.

busbyk (approved):
Sweet!

PR #111, opened 2025-04-13
Blocks: clean up unused blocks

rchlfryn:
### Description
When making #7 I noticed we had a few blocks we will never touch from initial set up. This PR remove `ArchiveBlock` and `Code` block in nice neat commits in case we want to revert at a later time.


busbyk:
I love unused code getting deleted!

PR #112, opened 2025-04-14
package.json: run prettier after code generation

stevekuznetsov:
The code generators do not output code perfectly to the Prettier specification, so we can run Prettier after generation to minimize the diff. Having to remember to do this every time is annoying, and we know we need to do it every time, so we might as well just inline the task.


stevekuznetsov:
@rchlfryn broke this out from the other PR like you asked

busbyk:
We may still want this but I did add a husky pre-commit hook using lint-staged to run pnpm prettify on commit which should help with diffs. But this would be nice for during development. 

I think this will only help for when we have to manually run these commands though. 

I assume that payload doesn't use pnpm to run their import generation when import generation happens automatically. The only times I've had to run the `payload generate:importmap` is when I include a custom component and payload can't detect that I added an import because it's just a string.

stevekuznetsov:
@busbyk I'm happy with whatever is less annoying for you two! @rchlfryn does the pre-commit hook work for ya?

busbyk:
> @busbyk I'm happy with whatever is less annoying for you two! @rchlfryn does the pre-commit hook work for ya?

I think this can't hurt. Might as well just format it when we run these manually. The pre-commit hook should handle the scenario where payload generated these.

busbyk:
@stevekuznetsov the build is complaining about a missing secret key. I can't see the GH actions secrets for this repo. Can you take a look when you have a chance?

stevekuznetsov:
@busbyk that happens when you make  PR from your fork of the repo instead of a branch *on* the repo - I have some local git automation to push PRs from forks so it gets messy when I forget that doesn't work on this repo. I can fix it up if we want to keep these changes but if the pre-commit hook works, we can just close this out

stevekuznetsov:
Ah sorry, didn't see your comment - superseded by https://github.com/NWACus/web/pull/112

PR #117, opened 2025-04-15
.github: use an empty initial DB, initialize and seed at runtime

stevekuznetsov:


seed-build.sh
```diff
@@ -0,0 +1,41 @@
...
+  if ! curl -s http://localhost:3000; then
+    echo "Dev server not yet started, waiting..."
+    sleep 1
+  else
+    break
+  fi
+done
+
+echo "[INFO] Logging in with the bootstrap user and seeding the database..."
+curl -s -H 'Content-Type: application/json' -H 'Accept: application/json' -X POST http://localhost:3000/api/users/login --data '{"email":"bootstrap@avy.com","password":"test"}' > login.json
```
stevekuznetsov: What I did to generate `dev.db.initialized` was just start the server and create the user with these credentials. Unfortunately, if we can't automate *that*, we have the same problem because changes to the schemas without updated to the `dev.db.initialized` will cause hangups:

```
[✓] Pulling schema from database...
? Warnings detected during schema push: 

· You're about to add not-null fake column without default value, which contains 1 items

DATA LOSS WARNING: Possible data loss detected if schema is pushed.

Accept warnings and push schema to database? › (y/N)
```

stevekuznetsov: Let me see if I can automate the first user as well ...

busbyk: This should do it: 

```
curl -X POST \
  http://localhost:3000/api/users/first-register \
  -H "Content-Type: multipart/form-data" \
  -F "_payload={\"name\":\"Avy\",\"email\":\"bootstrap@avy.com\",\"password\":\"somepassword\",\"confirm-password\":\"somepassword\"}"
```

stevekuznetsov: Yeah, already done, sorry

busbyk (approved):

PR #119, opened 2025-04-16
.github: add workflows to run deployments

stevekuznetsov:


.github/workflows/development.yaml
busbyk, line 1: Wouldn't this be the production deploy? Should we name this `production.yaml`?

.github/workflows/cleanup.yaml
```diff
@@ -0,0 +1,21 @@
+name: cleanup
+on:
+  delete:
+    branches-ignore:
+      - 'main'
+      - 'preview'
+      - 'release'
+jobs:
+  Deploy-Production:
```
busbyk: Should we name this `cleanup`?

.github/workflows/development.yaml
```diff
@@ -0,0 +1,59 @@
...
+        run: npm install --global vercel@latest
+      - name: Pull Vercel Environment Information
+        run: vercel pull --yes --environment=preview --token=${{ secrets.VERCEL_TOKEN }}
+      - name: Update database URI in Vercel
+        run: vercel env add DATABASE_URI preview main --sensitive <<<"${{ steps.connection.outputs.uri }}"
+      - name: Update bootstrap password in Vercel
+        run: vercel env add PAYLOAD_SEED_PASSWORD preview main --sensitive <<<"${{ secrets.PAYLOAD_SEED_PASSWORD }}"
+      - name: Build Project Artifacts
+        run: vercel build --yes --target=preview --token=${{ secrets.VERCEL_TOKEN }}
+      - name: Deploy Project Artifacts to Vercel
```
busbyk, line 78: Nicee

.github/workflows/preview.yaml
```diff
@@ -0,0 +1,65 @@
...
+on:
+  push:
+    branches:
+      - '*'
+      - '**'
+      - '*/*'
+    branches-ignore:
+      - 'main'
+      - 'preview'
+      - 'release'
```
busbyk: Do we still care to ignore these?

.github/workflows/preview.yaml
```diff
@@ -0,0 +1,65 @@
+name: preview
+env:
+  VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+  VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+on:
+  push:
```
busbyk: Do we want to run this on pushes or only on pull requests?

.github/workflows/development.yaml
stevekuznetsov: `main` branch is our non-prod persistent deployment

.github/workflows/preview.yaml
stevekuznetsov: If you have a PR open and update the commits in it, I think it would be good to update the preview

busbyk: I've used this before to get that behavior:
```
on:
  pull_request:
    types: [opened, synchronize]
```

I'm just thinking that we don't want to be creating Turso dbs when pushing to feature branches that are in progress 🤷

stevekuznetsov: (& it was not obvious if the `pull_request` [type](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) supported update actions)

.github/workflows/development.yaml
busbyk: Ahh the intent is to have `release` as production still?

.github/workflows/preview.yaml
stevekuznetsov: Ah, `synchronize` - will update this in the morning, thanks.

stevekuznetsov:
@busbyk hope it's ok - going to iterate this morning while nobody's around & pull this in

stevekuznetsov:
Well, looks like making it run on pull request events makes it run on _this_ pull request, which is nice!

stevekuznetsov:
```
> cross-env NODE_OPTIONS=--no-deprecation next build
   ▲ Next.js 15.2.4
   Creating an optimized production build ...
 ✓ Compiled successfully
./src/Header/Component.tsx
25:9  Warning: 'nav' is assigned a value but never used. Allowed unused vars must match /^_/u.  @typescript-eslint/no-unused-vars
info  - Need to disable some ESLint rules? Learn more here: https://nextjs.org/docs/app/api-reference/config/eslint#disabling-rules
```

Wild, we run `eslint` in CI already - why didn't this get caught?

stevekuznetsov:
Deployment worked, but:

1. /admin gets a 500
2. links on the home page are for localhost
3. we're still under `avy-*-nwac.vercel.app`, likely want something like `unique-id.preview.avy-fx.org`

stevekuznetsov:
I'm going to merge this in to see what the actual deployment looks like on avy-fx.org

PR #125, opened 2025-04-16
*: DRY out server URL logic

stevekuznetsov:
We inherited a bit of a mess from the website template. This commit DRYs it out (while leaving the client-side URL for now). We bridge the gap between our pipeline-set `$SERVER_DOMAIN` and all of the `$VERCEL*URL` variables, which only hold the domain, and the
`$NEXT_PUBLIC_SERVER_URL`, which holds the full URL.


next-sitemap.config.cjs
```diff
@@ -1,7 +1,7 @@
-const SITE_URL =
-  process.env.NEXT_PUBLIC_SERVER_URL ||
-  process.env.VERCEL_PROJECT_PRODUCTION_URL ||
-  'https://example.com'
+const domain = process.env.SERVER_DOMAIN || process.env.VERCEL_URL
```
stevekuznetsov, line 1: can't import anything in this context, so we can copy paste...

next.config.js
```diff
@@ -2,8 +2,9 @@ import { withPayload } from '@payloadcms/next/withPayload'
 
 import redirects from './redirects.js'
 
-const NEXT_PUBLIC_SERVER_URL = process.env.VERCEL_PROJECT_PRODUCTION_URL
-  ? `https://${process.env.VERCEL_PROJECT_PRODUCTION_URL}`
+const domain = process.env.SERVER_DOMAIN || process.env.VERCEL_URL
```
stevekuznetsov, line 5: can't import anything in this context, so we can copy paste...

src/utilities/isPayloadAdminPanel.ts
```diff
@@ -1,10 +0,0 @@
-import type { PayloadRequest } from 'payload'
```
stevekuznetsov, line 1: This file had server URL logic but was unused, so I removed it

PR #130, opened 2025-04-17
Forcing light mode

busbyk:
This keeps the code that would handle toggling between light and dark mode but forces the site to use light mode. 

Note: admin panel still uses Payload's auto/light/dark settings and is unaffected by this change.


busbyk:
A few questions for review:
1. Should I just delete all of this code and we can revert/cherry pick from this PR when we need it again? Are my changes here confusing because they ignore the determination of the mode based on the user's preference?
2. Do we even need this concept of a header theme? I'm not clear on it's intent. Many page client components use it throughout the codebase: https://github.com/NWACus/web/blob/main/src/providers/HeaderTheme/index.tsx
image

src/Header/Component.client.tsx
```diff
@@ -1,25 +1,6 @@
 'use client'
```
stevekuznetsov, line 1: I think for most of these, no client component is necessary whatsoever if we don't have the toggle, so perhaps in the spirit of this not being hard to un-do, could we just keep these files as-is but remove the user-visible toggle in the header/footer?

rchlfryn:
Responses to your qs:
1. I think this makes sense to remove the code. I do think we should label it as `nwac-light`  or `nwac` or something showing NWAC is the them, since those are the colors we are using. I also think it is good to delete because at this point we are not maintaining dark mode. Maybe you can level a comment or a TODO with a link to this PR since we know we are going to add this back in? But it should be easy enough to find.
2. I am not sure what this is either.

rchlfryn (changes requested):
Responses to your qs:

1. I think this makes sense to remove the code. I do think we should label it as nwac-light or nwac or something showing NWAC is the them, since those are the colors we are using. I also think it is good to delete because at this point we are not maintaining dark mode. Maybe you can level a comment or a TODO with a link to this PR since we know we are going to add this back in? But it should be easy enough to find.
2. I am not sure what this is either.

busbyk:
> Responses to your qs:
> 
> 1. I think this makes sense to remove the code. I do think we should label it as nwac-light or nwac or something showing NWAC is the them, since those are the colors we are using. I also think it is good to delete because at this point we are not maintaining dark mode. Maybe you can level a comment or a TODO with a link to this PR since we know we are going to add this back in? But it should be easy enough to find.
> 2. I am not sure what this is either.

Cool, I'm with you. I'm thinking I'll remove the theme concept altogether for now then and reference this PR in a new issue for setting up theming based on settings from the CMS. 

But I'll remove the entire header theme concept and I don't think we'll add that back since none of us know what the intent is there 😄

src/Header/Component.client.tsx
busbyk: Yea I agree. It only needed to be a client component due to the use hooks. 

I'm thinking I'm going to remove the header theme concept https://github.com/NWACus/web/pull/130#issuecomment-2814256829 so I can just delete this particular component.

stevekuznetsov (approved):

rchlfryn (approved):

Issue #52, opened 2025-03-15
Ensure Next.js cache revalidation is done as an afterChange hook on every collection


stevekuznetsov:
We should ensure that we have an `afterChange` hook configured for all collections that influence website content and make sure that we call `revalidatePath` for the path(s) affected by the change, as well as the sitemap(s) that these path(s) are shown in.

busbyk:
Whoever works on this should consider https://github.com/NWACus/web/pull/132#discussion_r2050844832 which calls out the complexity involved in revalidating pages with [relation blocks](https://github.com/NWACus/web/blob/busbyk/docs/docs/decisions/005-collection-relation-blocks.md). When a document is updated in a collection that's referenced by a block, we need to revalidate the page that block is included in.

PR #132, opened 2025-04-18
decision doc for collection relation blocks approach

busbyk:


docs/decisions/005-collection-relation-blocks
```diff
@@ -0,0 +1,38 @@
...
+We could create a few blocks:
+- Staff Card (a single staff member's data displayed in a card)
+- Staff List (multiple Staff Cards composed together), order is controllable
+- Staff Group (multiple Staff Cards composed together), filtered by staff relation to staff groups collection which would be passed a prop to this block via the admin panel
+
+We could enforce that block order is always controlled or include an option in the admin panel to auto populate from the collection using a default sort order. This approach is flexible.
+
+## Consequences
+
+This introduces complexity into page revalidation because updating a collection updates the content on many pages.
```
busbyk: This probably warrants further discussion. This only occurred to me as I was writing this up. I think we could effectively lookup which pages contain blocks with relations to this collection on update but that logic could get complex, quick. Maybe worth a POC.

stevekuznetsov: I had a PoC of this in January - it's a long tree walk but not impossible. I'll try to tackle this sometime soon once we have relationships set up and real data. If we're worried that logic doesn't land before we go live, we can do something simplistic and revalidate everything.

docs/decisions/005-collection-relation-blocks
```diff
@@ -0,0 +1,38 @@
...
+Staff as an example:
+
+We could create two collections:
+- staff
+- staff groups
+
+We could create a few blocks:
+- Staff Card (a single staff member's data displayed in a card)
+- Staff List (multiple Staff Cards composed together), order is controllable
+- Staff Group (multiple Staff Cards composed together), filtered by staff relation to staff groups collection which would be passed a prop to this block via the admin panel
```
stevekuznetsov: I don't think I grok what this means 😅

docs/decisions/005-collection-relation-blocks
busbyk: Yea I think it'll be necessary. That work could be done under this task? https://github.com/NWACus/web/issues/52

Or might be worth creating a separate one.

docs/decisions/005-collection-relation-blocks
busbyk: Haha sorry let me try to elaborate. 

I'm calling out a few variations we could have with "relation blocks" and highlight that they are composable. 

If we assume Biographies and Teams collections where Teams have a one-to-many relationship with Biographies. 

You could have a few different [Blocks](https://payloadcms.com/docs/fields/blocks) and associated component rendered by the RenderBlocks component (some of which you've created on #137 ):

1. Team (ordered automatically): Displays all Biographies related to the selected team. Displayed in the order returned by payload. 
2. Biographies (ordered manually): Displays Biographies manually added by an editor in a specific order (using an array of Biography blocks). Could include a custom title/header and maybe a description for this too. 
3. Biography: Displays a single Biography. Maybe this is used for the [highlight biography](https://www.figma.com/design/vAJkcFrHt5gddho3ILBOEx/NWAC-UX-UI-Redesign---Work?node-id=607-32819&t=wtJQPMoFAOzSW0Iy-4).

You already have 1 and 3 in #137 but I could see a situation where an editor would want to highlight only some biographies, not necessarily grouped by team and control the order. 

Does that make more sense? Just highlighting some variation options.

stevekuznetsov: Oh yes 1000%. That sounds like a really awesome pattern that this approach unlocks for us!

docs/decisions/005-collection-relation-blocks
stevekuznetsov: Wow look at past me writing things down! #52 is what I intended to track this.

rchlfryn (approved):

docs/decisions/005-collection-relation-blocks
busbyk: Sweet, will get this section updated and then we can merge 🏋️

docs/decisions/005-collection-relation-blocks
busbyk: Perfect!

PR #136, opened 2025-04-19
*: add biographies, teams

stevekuznetsov:
Two new collections are added:

 - biographies: Information about someone associated with an avalanche center. These are disjoint from Users (but may refer to them) since:
   - not all people we want to highlight have logins, some are emeritus
   - some users are highlighted on more then one center's page
 - teams: A list of biographies with a title.

These collections are threaded through to the front-end using custom Blocks, and a "who we are" page is built using these blocks.


stevekuznetsov:
Blah, from my fork again ...

PR #137, opened 2025-04-19
*: add biographies, teams

### LINKED ISSUE #140: Blocks: Staff, Groups
We want AC staff to be able to enter data about folks who work for or are associated with the AC, what groups they fall into, etc. We need these blocks to display the data and be embeddable into pages.

stevekuznetsov:
Two new collections are added:

 - biographies: Information about someone associated with an avalanche center. These are disjoint from Users (but may refer to them) since:
   - not all people we want to highlight have logins, some are emeritus
   - some users are highlighted on more then one center's page
 - teams: A list of biographies with a title.

These collections are threaded through to the front-end using custom Blocks, and a "who we are" page is built using these blocks.


stevekuznetsov:
See [the preview](https://avy-a2dob7m7d-nwac.vercel.app/nwac/who-we-are) for the front-end, but I did not put a huge amount of effort into it!

stevekuznetsov:
Fixes #140

dev.db.2
stevekuznetsov, line 1: Oops, I'll delete after review.

src/components/ui/avatar.tsx
```diff
@@ -0,0 +1,47 @@
+'use client'
+
+import * as AvatarPrimitive from '@radix-ui/react-avatar'
+import * as React from 'react'
+
+import { cn } from 'src/utilities/cn'
```
stevekuznetsov: @busbyk not sure if you ran into this, but the CLI adds `from '@utilities/cn'` which doesn't work - is there something we can reconfigure to fix that?

src/collections/Biographies.ts
```diff
@@ -0,0 +1,71 @@
...
+    } else {
+      return true
+    }
+  } else {
+    if (!name) {
+      return 'A name must be provided if no user is linked to this biography.'
+    } else {
+      return true
+    }
+  }
```
rchlfryn: This logic is kind of confusing. Maybe just an explicit `if/else` here to call out the two scenarios we care about
```suggestion
  if (siblingData.user && name) {
    return 'A name cannot be provided when a user is linked to this biography.'
  }
   else if(!siblingData.user && !name) {
     return 'A name must be provided if no user is linked to this biography.'
  } else {
    return true
  }
```

seed-build.sh
```diff
@@ -10,11 +10,13 @@ function cleanup() {
     kill -SIGTERM "${job}"
     wait "${job}"
   done
+  rm -f login.json
```
rchlfryn, line 13: What are the two changes in this file doing? 

⛏️ If it has nothing to do with biographies I would suggest making a separate PR for it or calling it out in the description, otherwise it looks like a commit mistake

src/blocks/Biography/Biography.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Card, CardContent } from '@/components/ui/card'
+import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover'
+import { format, parseISO } from 'date-fns'
+import type { BiographyBlock as BiographyBlockProps } from 'src/payload-types'
+
+type Props = BiographyBlockProps
+
+export const BiographyBlock = ({ biography }: Props) => {
+  if (typeof biography !== 'object' || typeof biography.photo !== 'object') {
+    // TODO: figure out how to handle this - do we have access to the Payload client?
```
rchlfryn: We should have access to the payload client with a custom component. When will biography not be an object?

src/endpoints/seed/biographies.ts
rchlfryn, line 1: 👏

src/endpoints/seed/index.ts
```diff
@@ -1136,23 +1143,3 @@ export const innerSeed = async ({
 
   payload.logger.info('Seeded database successfully!')
 }
-
-async function fetchFileByURL(url: string): Promise<File> {
```
rchlfryn, line 1140: This is used by other non biography photos starting on like `726` so maybe we should move this under a utility file instead of nesting it under `./biographies`

src/collections/Pages/index.ts
```diff
@@ -93,7 +95,15 @@ export const Pages: CollectionConfig<'pages'> = {
...
               type: 'blocks',
-              blocks: [CallToAction, Content, FormBlock, ImageTextList, MediaBlock],
+              blocks: [
+                CallToAction,
+                Content,
+                FormBlock,
+                ImageTextList,
+                MediaBlock,
+                BiographyBlock,
+                TeamBlock,
```
rchlfryn: ⛏️  Alphabetical order here

src/blocks/Team/Team.tsx
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  const teamMembers: Biography[] = team.members.filter((member) => typeof member === 'object')
+
+  return (
+    <div className="container mx-auto py-12 px-4">
+      <h2 className="text-3xl font-bold text-center mb-12">{team.name}</h2>
+      <div className="grid grid-cols-1 sm:grid-cols-2 lg:grid-cols-3 gap-6">
+        {teamMembers.map((member, index) => (
+          <BiographyBlock key={index} biography={member} blockType={'biography'} />
```
rchlfryn: What was the reason for changing the member prop here? I think later on `biography.biography` is confusing. Maybe consider leaving it as `member`?

src/endpoints/seed/biographies.ts
```diff
@@ -0,0 +1,1332 @@
...
+      start_date: '2017-12-01',
+      biography:
+        'As a passionate backcountry skier, father of two young children (who love to ski!), and an advocate for responsible recreating in the backcountry, I’m honored to have served on the NWAC Board since 2017. In addition to skiing, I love taking advantage of all the outdoor activities the PNW has to offer: ultrarunning, mountaineering. biking, hunting, and fishing. Originally from California, I served as a U.S. Army infantry captain and am a decorated combat veteran (Operation Iraqi Freedom). My professional background includes management consulting with McKinsey & Company and operations executive roles at Expedia, A Place For Mom, and Porch Group, where I currently lead our insurance agency. I hold a mechanical engineering degree, with honors, from the United States Military Academy at West Point and an MBA from the Stanford Graduate School of Business.',
+    },
+    {
+      tenant: tenants['snfac'],
+      photo: 1,
+      name: 'Louise Stumph',
+      title: 'Director',
+      start_date: null,
```
rchlfryn, line 326: ⛏️  I noticed all of the `snfac` are `null`. Maybe make sure the have at least one in each tenant with and without the optional fields. Same goes for `title`. I think all `nwac` bios have a title. Maybe have one without a title

src/components/ui/avatar.tsx
rchlfryn: was this file nested under `ui` when you added `cn`? or did you move it after?
I haven't had an issue with this.

src/endpoints/seed/biographies.ts
rchlfryn: The seeding process takes quite a bit now. Do we want to keep adding data this way? Or maybe add a seperate seed file that looks more like prod? Probably a better conversation off github.

rchlfryn (changes requested):
Huge lift! took an initial review. 

I think we need to fix how the the image are rendering. right now they are distorted, but I can't find where this is being done. Can you point this out to me? It looks like they are being rendered square at 96x96. is this a shadcn thing?

src/components/ui/avatar.tsx
stevekuznetsov: I'm just adding components with the Shad CN CLI like `pnpm dlx shadcn@latest add popover` - absolutely no idea. I think the `cn` utility came in from the website template.

src/blocks/Biography/Biography.tsx
stevekuznetsov: > We should have access to the payload client with a custom component.

I couldn't figure out how to get access to the `payload` client in a custom block, specifically - it looks like we need to pass it from our `page.tsx` into `<RenderBlocks/>` and clean up our demultiplexing in `<RenderBlocks/>`. I can take a stab at that.

> When will biography not be an object?

If the output of `payload.find()` doesn't have `depth` set sufficiently in the top-level query, these references will just be IDs.

src/endpoints/seed/biographies.ts
stevekuznetsov: I realized that we can download the database that was seeded for the last merge to `main` branch and start from there. Having real data in the database IMO is really important, or these pages will not really look like anything. I would rather look into flows to make local data easy rather than optimize for quick seeds if possible. Checking in the seeded DB was originally supposed to do that.

src/blocks/Team/Team.tsx
stevekuznetsov: The prop types for `BiographyBlock` are auto-generated by Payload from the `Block` configuration. Changing to `member` there would mean that the field is called `member` to the users in the admin panel - that feels a bit awkward as well? Not sure what you'd prefer here.

src/endpoints/seed/biographies.ts
stevekuznetsov: This is simply a reflection of how the different centers are using the features today on their websites - so it was moreso pulling in their real data rather than trying to make sure some fields are or are not set.

seed-build.sh
stevekuznetsov: Sure, I can break it out.

stevekuznetsov:
> I think we need to fix how the the image are rendering. right now they are distorted, but I can't find where this is being done. Can you point this out to me? It looks like they are being rendered square at 96x96. is this a shadcn thing?

I think this is something coming from Shad CN's `Avatar` - if it's alright with you, since I am such a noobie with front-end, the TODO to figure out how to get that to play nice with Next.js image component could be something one of you two picks up? I am way out of my depth.

src/blocks/Biography/Biography.tsx
stevekuznetsov: Please see the latest commit. This is pretty hairy.

src/endpoints/seed/biographies.ts
rchlfryn: I am totally for building a DB, but I wonder if there is a way to migrate it rather than wipe it and start from scratch every time. When reviewing PRs, I am finding I am spending a lot of time blowing away the DB and waiting for it to build and I know that will only get worse.

src/blocks/Team/Team.tsx
rchlfryn: I don't have a better nomenclature for this, so let's leave it.

rchlfryn:
> > I think we need to fix how the the image are rendering. right now they are distorted, but I can't find where this is being done. Can you point this out to me? It looks like they are being rendered square at 96x96. is this a shadcn thing?
> 
> I think this is something coming from Shad CN's `Avatar` - if it's alright with you, since I am such a noobie with front-end, the TODO to figure out how to get that to play nice with Next.js image component could be something one of you two picks up? I am way out of my depth.

Don't worry about this. We can get it in a follow up PR. I was more so curious how/why it was rendering stretched. Happy to help to get you back in your depth if you want to pair/chat things out.

src/components/ui/avatar.tsx
rchlfryn: Oh interesting... I tried to dig around it seems like they have a few mapping issues - see the last comment (https://github.com/shadcn-ui/ui/issues/6409) 
for a solution and we may just have to reimport it when we add it... seems like the lowest lift since I dont know how we can add it as an `alias` in the `components.json` file

rchlfryn: when I imported a component it added `import { cn } from '@/utilities` which is correct from the `components.json` alias. we just need to change it to `import { cn } from '@/utilities/cn`

busbyk: Yea I ran into the same thing and just updated the import after running `pnpm dlx shadcn@latest add`. 

I just glossed over this tbh since it was a quick import fix and moved on but I think we could change our aliases.utils to `@/utilities/cn` and the shadcn cli will pick this up correctly. I just tried it and the import is correct. 

```
  "aliases": {
    "components": "@/components",
    "utils": "@/utilities/cn" // just add /cn
  }
```

busbyk: Oh looks like you already added this!

src/blocks/Form/Component.tsx
busbyk, line 1: These are just typescript fixes?

src/components/ui/avatar.tsx
stevekuznetsov: Yep sorry, I updated `components.json` in the latest commit in this PR.

src/blocks/RenderBlocks.tsx
```diff
@@ -25,25 +19,35 @@ export const RenderBlocks = (props: { blocks: Page['layout'][0][] }) => {
...
+    case 'formBlock':
+      return <FormBlock key={key} {...block} />
+    case 'imageTextList':
+      return <ImageTextList key={key} {...block} />
+    case 'mediaBlock':
+      return <MediaBlock key={key} {...block} />
+    case 'team':
+      return <TeamBlock key={key} {...block} />
+  }
+}
```
busbyk: You can't access the key prop like this - it's not exposed by React. 

This is resulting in this error logged to the console: 
```
 Server  RenderBlock: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://react.dev/link/special-props)
```

I believe React is falling back to using index - I think that's the default behavior. 

And actually we only really need the key prop at the first element rendered in a .map():

```tsx
import { Fragment } from 'react'

import type { Page } from '@/payload-types'

import { BiographyBlock } from '@/blocks/Biography/Biography'
import { CallToActionBlock } from '@/blocks/CallToAction/Component'
import { ContentBlock } from '@/blocks/Content/Component'
import { FormBlock } from '@/blocks/Form/Component'
import { ImageTextList } from '@/blocks/ImageTextList/Component'
import { MediaBlock } from '@/blocks/MediaBlock/Component'
import { TeamBlock } from '@/blocks/Team/Team'

export const RenderBlocks = (props: { blocks: Page['layout'][0][] }) => {
  const { blocks } = props

  const hasBlocks = blocks && Array.isArray(blocks) && blocks.length > 0

  if (hasBlocks) {
    return (
      <Fragment>
        {blocks.map((block) => {
          return (
            <div className="my-16" key={`${block.id}__${block.blockType}`}>
              <RenderBlock block={block} />
            </div>
          )
        })}
      </Fragment>
    )
  }

  return null
}

export const RenderBlock = ({ block }: { block: Page['layout'][0] }) => {
  const { blockType } = block
  switch (blockType) {
    case 'biography':
      return <BiographyBlock {...block} />
    case 'cta':
      return <CallToActionBlock {...block} />
    case 'content':
      return <ContentBlock {...block} />
    case 'formBlock':
      return <FormBlock {...block} />
    case 'imageTextList':
      return <ImageTextList {...block} />
    case 'mediaBlock':
      return <MediaBlock {...block} />
    case 'team':
      return <TeamBlock {...block} />
  }
}
```

In this case I changed the key to be based on the block's id and blockType. Not a huge deal but I generally try to use data from the database rather than something like an index whenever possible (see: https://react.dev/learn/rendering-lists#where-to-get-your-key). So block id + block type should be unique.

src/collections/Biographies.ts
```diff
@@ -0,0 +1,65 @@
+import { accessByTenantWithPermissiveRead } from '@/access/byTenant'
+import { filterByTenant } from '@/access/filterByTenant'
+import { tenantField } from '@/fields/tenantField'
+import { Biography } from '@/payload-types'
+import { CollectionConfig, TextField, Validate } from 'payload'
+
+// the types like TextFieldValidation are not parameterized by the collection, so they have `unknown` for siblingData
+const validateName: Validate<string, unknown, Biography, TextField> = (name, { siblingData }) => {
```
busbyk, line 8: Nice

src/collections/Biographies.ts
```diff
@@ -0,0 +1,65 @@
...
+      name: 'start_date',
+      type: 'date',
+      label: 'Start Date',
+      required: false,
+    },
+    {
+      name: 'biography',
+      type: 'text',
+      required: false,
+    },
```
busbyk: Should this be a RichText field? 🤔

busbyk: Could limit to bold, italic, underline, things like that. 

This <input> doesn't look very good for longer biographies: 
image

Or actually a https://payloadcms.com/docs/fields/textarea field might be sufficient.

src/blocks/Biography/Biography.tsx
busbyk: I think we decided on _not_ doing this to respect access control but FYI something like this is possible if you ever need to access the local API in a server component: 
```
import configPromise from '@payload-config'

type Props = BiographyBlockProps

export const BiographyBlock = async ({ biography: maybeBiography }: Props) => {

  let biography = maybeBiography;

  if (typeof biography === 'number') {
    try {
      biography = await payload.findByID({
        collection: 'biographies',
        id: biography
      })
    } catch (err) {
      console.error(`Failed to populate biography with id ${biography}. Err: `, err)
      return <></>
    }
  }
...
```

src/blocks/Biography/Biography.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const name: string =
+    biography.name || (typeof biography.user === 'object' && biography.user?.name) || 'Unknown'
+  const initials = name
+    .split(' ')
+    .map((part) => part.substring(0, 1))
+    .join(' ')
+
+  // TODO: do we need to do something special to get Next Image support in there?
+  return (
+    <Card className="w-full max-w-sm overflow-hidden">
```
busbyk: `max-w-sm` is causing extra spacing on larger screen sizes: 
Screenshot 2025-04-24 at 1 09 11 PM

Removed: 
Screenshot 2025-04-24 at 1 09 28 PM

I also don't think we need overflow-hidden here unless you were seeing something that needs that?

src/blocks/Biography/Biography.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const name: string =
+    biography.name || (typeof biography.user === 'object' && biography.user?.name) || 'Unknown'
+  const initials = name
+    .split(' ')
+    .map((part) => part.substring(0, 1))
+    .join(' ')
+
+  // TODO: do we need to do something special to get Next Image support in there?
```
busbyk: I like this approach [here](https://github.com/shadcn-ui/ui/issues/368#issuecomment-2678650485) using `getImageProps` from next/image. 

But we'd need to do some work on the image we get from payload first similar to what is happening in ImageMedia. 

Like generating the right sizes: https://github.com/NWACus/web/blob/e8c040517dc9cb133613ea135f510d10ea84bddf/src/components/Media/ImageMedia/index.tsx#L49

And potentially everything else in this if block: https://github.com/NWACus/web/blob/e8c040517dc9cb133613ea135f510d10ea84bddf/src/components/Media/ImageMedia/index.tsx#L36

FYI this is where that placeholder data uri is defined: https://github.com/NWACus/web/blob/e8c040517dc9cb133613ea135f510d10ea84bddf/src/components/Media/ImageMedia/index.tsx#L16

At the very least we should remove that haha. 

I actually think we should write our own image wrapper or modify this one to suit our needs a little better.

src/blocks/Biography/Biography.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <Card className="w-full max-w-sm overflow-hidden">
+      <CardContent className="p-6">
+        <div className="flex flex-col items-center space-y-4">
+          {biography.biography ? (
+            <Popover>
```
busbyk, line 31: I see why you wanted to use a popover - we don't want to expand all of the other cards in the grid by using an accordion or collapsible. But I wonder if it wouldn't be better to just include the entire biography if it exists below the their title and just restrict the biography length. I'll be interested to hear what Chris thinks. 

Maybe a Dialog.

src/components/ui/avatar.tsx
```diff
@@ -0,0 +1,47 @@
...
+))
+Avatar.displayName = AvatarPrimitive.Root.displayName
+
+const AvatarImage = React.forwardRef<
+  React.ElementRef<typeof AvatarPrimitive.Image>,
+  React.ComponentPropsWithoutRef<typeof AvatarPrimitive.Image>
+>(({ className, ...props }, ref) => (
+  <AvatarPrimitive.Image
+    ref={ref}
+    className={cn('aspect-square h-full w-full', className)}
```
busbyk: This `className={cn('aspect-square h-full w-full object-cover', className)}` solves the distorted images problem: [object-cover](https://v3.tailwindcss.com/docs/object-fit).

I think it makes sense to add this here as it's a reasonable default rather than adding it every time we use Avatar.

src/blocks/Form/Component.tsx
stevekuznetsov: Yeah, the commit message has context

src/blocks/Biography/Biography.tsx
stevekuznetsov: Our friends at v0.dev did this for me, and I am too stupid to know better :)

src/blocks/Biography/Biography.tsx
stevekuznetsov: Do you think that needs to happen in this PR, or should we break it out?

src/blocks/Biography/Biography.tsx
stevekuznetsov: They're all pretty long, so I don't know that chopping them off to a length makes a meaningful difference if we need to invent a UX to let users see the rest anyway - perhaps a dialog is better?

src/blocks/Biography/Biography.tsx
busbyk: I think that's worth it's own PR! I'll create a task for it.

busbyk: https://github.com/NWACus/web/issues/144

src/blocks/Biography/Biography.tsx
busbyk: That is the approach used on the nwac site currently - that could work well. I think including the photo in the dialog as well as the name, title, and full biography would be nice.

src/blocks/Biography/Biography.tsx
busbyk: Silly robots!

busbyk (approved):
Approving assuming we'll do a little UI tweaking under #140 still. Great work @stevekuznetsov !

src/blocks/Biography/Biography.tsx
stevekuznetsov: Splitting out the polish into https://github.com/NWACus/web/issues/146

rchlfryn (approved):
Reminder to rename `cn.ts` to `ui.ts`

stevekuznetsov:
@rchlfryn dang, good call - forgot that one. I don't have access to my OSS development environment (signing keys etc) until I'm back next Wednesday. Happy to do that in a follow-up when I'm back or we can keep this open until then.

rchlfryn:
> @rchlfryn dang, good call - forgot that one. I don't have access to my OSS development environment (signing keys etc) until I'm back next Wednesday. Happy to do that in a follow-up when I'm back or we can keep this open until then.

:shipit: ship it - I can make a PR follow up

PR #141, opened 2025-04-21
Blocks: Link preview

### LINKED ISSUE #107: Blocks: LinkPreview
Add block to allow AC staff to show previews of external cross links

Image
Image
Image

rchlfryn:
### Description
Add block that allows user to add cards with an image, title, text and button link

### Issue
Resolves #107 

### Screenshots
linkPreview2


src/blocks/LinkPreview/config.ts
```diff
@@ -0,0 +1,47 @@
...
+    type: 'text',
+    required: true,
+  },
+
+  {
+    name: 'text',
+    type: 'textarea',
+    required: true,
+  },
+  link({
```
stevekuznetsov: Can we maybe factor out a more narrow `link` that 

a) doesn't have a label, since it would not be clear to me what the difference between this card field's title and the link's label is
b) lets you compose in things like the button label, rather than having a wide-open "override anything you want" API that would make refactors to the core logic hard to do since you never know what users are doing with your `link()`

src/blocks/LinkPreview/Component.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const colsClasses: { [key: number]: string } = {
+    2: 'lg:col-span-6',
+    3: 'lg:col-span-4',
+  }
+  const colsSpanClass = colsClasses[numOfCols]
+
+  return (
+    <div className="bg-slate-500 py-8">
+      <div className="container grid grid-cols-4 lg:grid-cols-12 gap-x-2">
+        {cards &&
```
stevekuznetsov, line 29: Why not use the ShadCN `<Card>`?

rchlfryn: Completely forgot about ShadCN.. updated in [4ad2bed](https://github.com/NWACus/web/pull/141/commits/4ad2bed00c0e114fcc0a2eba1a32af81412b29b3)

stevekuznetsov:
Very cool - how did you generate the thumbnail? Photoshop on your machine?

src/blocks/LinkPreview/config.ts
rchlfryn: @stevekuznetsov I am a bit confused about this.
a) I think it is pretty clear the difference between the button label and title of the card. Was that what you meant? Take a look at the interface and let me know


b) I was assuming we just use `link` for buttons, but that doesn't seem like the case?? how should I tighten down the hatches? I think it is useful to be able to link to pages within the website as well as link to other websites. Thoughts??

rchlfryn:
> Very cool - how did you generate the thumbnail? Photoshop on your machine?

I just mocked it up in Figma real quick. You can see them in the file.

src/blocks/LinkPreview/config.ts
stevekuznetsov: ~In the admin UI it seems clear, but I guess what I am missing is what this ends up looking like once it renders - if it's a button, I understand what happens to be button label text, but where does the `Title` field of this link end up? Is that a mouseover?~ edit: re-looked at the example page and it now makes sense, sorry!

For b), I think it's just a stylistic thing in my opinion, and I seem to be in the minority on this in JavaScript world, so feel free to just ignore. It seems difficult in my opinion to reason about `link()` as it stands - if you were refactoring something inside of it, adding a new named field for instance, you have no real idea of how callers of `link()` could be affected since those users could override literally any part of it - what would you expect to happen if someone passed in an override for `{fields: ...}` or changed the `{type: ...}`, or...? Even though the contract of the function is "override anything you could possibly want" the reality is you need to carefully read the implementation to figure out which parts are and are not safe to override. Exposing the parts that actually need to vary explicitly would be how I'd approach it, but happy to quiet up on this if it's just not how things are done in JS.

rchlfryn: @stevekuznetsov This is a very interesting perspective, so thank you for sharing. I would assume, that anyone who would want to change anything about `link()` would need to look where it is called/used and not create any breaking changes. Or, it they do want to completely diverge, created a secondary `link()` file with a new name. 

That being said, I think I understand what you are saying now. We should refactor `link()` a bit more so specific and not a catch all? Totally doable. For this instance, make it button specific..

stevekuznetsov: We could do that refactor, yeah - not something you need to do ASAP nor should it block the PR here though.

rchlfryn: Refactored `link` into `button`... it should probably be `buttonLink` since it is a link disguised as a button. I would assume button to have a submit action. I might change that.

stevekuznetsov (approved):

PR #147, opened 2025-04-26
Update cn import to match payload

rchlfryn:
#### Description
In #137 we noticed the import for `cn` when adding a shadcn component was wrong.  (See [discussion](https://github.com/NWACus/web/pull/137#discussion_r2051351658) ). Payload calls this file `ui.ts`, so we will copy that.


stevekuznetsov:
Thanks for picking this up for me!

PR #150, opened 2025-04-28
Blocks: image link grid

### LINKED ISSUE #105: Blocks: ImageGridLinks
Allow AC staff to add ImageGrid which contains

- Image grid that is flexible between 1,2,4 columns
- Make images linkable

Image
Image

rchlfryn:
#### Description
Adds custom block for ImageLinkGrid

#### Issue
Resolves #105 

#### Screenshot
linkGrids


src/payload-types.ts
```diff
@@ -1778,12 +1817,13 @@ export interface PagesSelect<T extends boolean = true> {
...
     | T
     | {
+        biography?: T | BiographyBlockSelect<T>;
         cta?: T | CallToActionBlockSelect<T>;
         content?: T | ContentBlockSelect<T>;
         formBlock?: T | FormBlockSelect<T>;
+        imageLinkGrid?: T | ImageLinkGridSelect<T>;
         imageTextList?: T | ImageTextListSelect<T>;
         mediaBlock?: T | MediaBlockSelect<T>;
-        biography?: T | BiographyBlockSelect<T>;
```
stevekuznetsov: Sorry I had these out-of-order!

stevekuznetsov (approved):
Looks good! Leaving for Kellen since I have no expertise on the front-end.

src/blocks/ImageLinkGrid/Component.tsx
```diff
@@ -0,0 +1,49 @@
...
+    1: 'sm:col-span-12',
+    2: 'sm:col-span-6',
+    3: 'sm:col-span-4',
+    4: 'sm:col-span-3',
+  }
+  const colsSpanClass = colsClasses[numOfCols]
+
+  return (
+    <div className="container">
+      <div className="grid sm:grid-cols-12 gap-y-2 gap-x-2">
```
busbyk: The gap in Figma is 12px which would be `gap-{x/y}-3`. You could just use `gap-3` if you're using the same vertical and horizontal gap but I'm fine with using gap-y- and gap-x- separately if you prefer.

src/blocks/ImageLinkGrid/Component.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {columns &&
+          columns?.length > 0 &&
+          columns.map((col, index) => {
+            const { caption, image, link } = col
+            const lastOddElement = numOfCols % 2 && index === numOfCols - 1
```
busbyk: `lastOddElement` is never used. Do you want to remove?

src/blocks/ImageLinkGrid/Component.tsx
```diff
@@ -0,0 +1,49 @@
...
+            return (
+              <div className={cn(`relative ${colsSpanClass} ${className}}`)} key={index}>
```
busbyk: I think passing in `className` like this will technically still merge tailwind class names correctly but if it's undefined we'll see that string in the classes (not really a big deal). But there is an extra `}` here. 

rendered classes

```suggestion
              <div className={cn('relative', colsSpanClass, className)} key={index}>
```

src/blocks/ImageLinkGrid/Component.tsx
```diff
@@ -0,0 +1,49 @@
...
+                <a href={`${link?.url}`}>
+                  <ImageMedia
+                    imgClassName={cn('object-cover w-full h-[280px]', imgClassName)}
+                    resource={image}
+                  />
+                  <div className="absolute bottom-0 w-full bg-accent p-2 text-small text-center">
+                    {caption}
+                  </div>
```
busbyk: This is an overlay on top of the image. Do we want that? Or should this element be inline?

src/blocks/ImageLinkGrid/Component.tsx
```diff
@@ -0,0 +1,49 @@
...
+                </a>
+              </div>
+            )
```
busbyk: We could add a hover state. Something like:?

hover state example


```suggestion
            return (
              <div
                className={cn('relative group overflow-hidden', colsSpanClass, className)}
                key={index}
              >
                <a href={`${link?.url}`}>
                  <ImageMedia
                    imgClassName={cn(
                      'object-cover w-full h-[280px] group-hover:scale-105 transition-transform duration-300 ease-in-out',
                      imgClassName,
                    )}
                    resource={image}
                  />
                  <div className="absolute bottom-0 w-full bg-accent p-2 text-small text-center group-hover:bg-accent/90">
                    {caption}
                  </div>
                </a>
              </div>
            )
```

src/blocks/ImageLinkGrid/config.ts
```diff
@@ -0,0 +1,105 @@
+import type { Block } from 'payload'
+
+import type { GroupField } from 'payload'
+
+const linkField: GroupField = {
```
busbyk, line 5: I wonder if we should just modify the [link field that came with the Payload website template](https://github.com/NWACus/web/blob/2ea77ea5b06b9571d30216026c3620e55d3b2023/src/fields/link.ts#L24) since this is so close to that and we'll probably use this in many places. Maybe a separate PR?

busbyk (changes requested):

src/blocks/ImageLinkGrid/config.ts
busbyk: There's also the [navLink I created](https://github.com/NWACus/web/blob/main/src/fields/navLink/index.ts) in #90

src/blocks/ImageLinkGrid/Component.tsx
stevekuznetsov: n.b. eslint should catch this given our config - should figure out why it didn't flag it in CI

rchlfryn: @stevekuznetsov Looks like we need to add `"noUnusedLocals": true` to our `tsconfig`. When I added it, there were a lot of other unused variables so I will make a new PR.

src/blocks/ImageLinkGrid/Component.tsx
rchlfryn: I dont like the overlay either, but I didn't know a better way to keep the image square and then have the caption. If the caption is inline, then we have to do some crazy math magic, or have the block not be 1:1 square. Thoughts? I can look into using [AspectRatio](https://ui.shadcn.com/docs/components/aspect-ratio) but an overlay seemed easier

src/blocks/ImageLinkGrid/Component.tsx
rchlfryn: Yeah! Updated in [24572fc](https://github.com/NWACus/web/pull/150/commits/24572fcda724d0ed5c2bae8c2f06d5d406814692) to avoid conflicts

src/blocks/ImageLinkGrid/config.ts
rchlfryn: @stevekuznetsov and I were talking about this in [another PR](https://github.com/NWACus/web/pull/141#discussion_r2052547943) and would like to not make `link` a catch all. I made an [issue](https://github.com/NWACus/web/issues/174) for us to discuss.

src/blocks/ImageLinkGrid/Component.tsx
busbyk: Well the images aren't square right now though right? The four column and three column layouts are close to square based on the fixed height you selected. But if you wanted them to be truly square you could use `aspect-square` and not specify a height. 

But for the two column and one column layouts you don't want the images to be square anyways, their heights would be way too big. 

So could adjust the fixed height down a little bit? 

But if you take off the absolute positioning for the caption, you'll need a wrapper for the image for the overflow: hidden so it doesn't overflow over the caption:
```
        return (
              <div className={cn('relative group', colsSpanClass, className)} key={index}>
                <a href={`${link?.url}`}>
                  <div className="w-full h-[280px] overflow-hidden">
                    <ImageMedia
                      imgClassName={cn(
                        'object-cover w-full h-full group-hover:scale-105 transition-transform duration-300 ease-in-out overflow-hidden',
                        imgClassName,
                      )}
                      resource={image}
                    />
                  </div>
                  <div className="w-full bg-accent p-2 text-small text-center group-hover:bg-accent/90">
                    {caption}
                  </div>
                </a>
              </div>
            )
```

stevekuznetsov (approved):
:shipit:

busbyk (approved):
I'm happy with this once [this](https://github.com/NWACus/web/pull/150#discussion_r2080400819) is resolved.

src/blocks/ImageLinkGrid/Component.tsx
rchlfryn: Personally I like position absolute over overflow hidden, but I'll concede this one since inline is better.

Issue #32, opened 2025-03-06
AC staff can style their site as desired


rustynwac:
Branding
Colors
Logo

stevekuznetsov:
Good to specify if we think this does or does not apply to widget content.

PR #153, opened 2025-04-29
Blocks: content

### LINKED ISSUE #34: Blocks: Content - headers and text
AC staff has a view that displays blocks with headers and text

Image

### LINKED ISSUE #103: Blocks: Content w/ dark styling
AC staff has a view that displays blocks with headers and text w/ columns with dark styling

Image

rchlfryn:
### Description
A lot is going on with this PR. Overall it alters the **main** content block.
* Changes content block admin interface
    * Include horizontal rule and moves optional link
    * Removes size field
    * Adds a background color picker (TODO - In another PR - make this dynamic for all avalanche centers by importing theme colors)
    * Adds ability to not enable columns
* Adds `color` [package](https://www.npmjs.com/package/color) 

### Issue
Resolves #34 #103 

### Discussion topics
1. Do we want to use the color package? It is used to determine text color for background based on the user's selection. This might be overkill and we can do this manually but thought it was interesting.
2. Do we need `enableColumns` checkbox field? I am not convinced it is a better UI experience. I just think it is confusing to only add one column. Perhaps the solution is to just change the column name to something else.
3. Tailwind safelist - We can use `style`, but I am curious about using safelist. I think this will be a problem once we start importing theme colors. 
4. Are we okay altering this content block or should be make a new one?

### Screenshots
Columns disabled
Screenshot 2025-04-28 at 18 40 30

Columns enabled
Screenshot 2025-04-28 at 18 41 15


src/fields/color.ts
```diff
@@ -0,0 +1,15 @@
...
+  name: 'color',
+  type: 'text',
+  label: 'Background color',
+  required: true,
+  admin: {
+    components: {
+      Field: '@/components/ColorPicker',
+    },
+  },
+}
```
busbyk: If we want this to be a reusable field we'd probably want to make this a function and accept label and required as props with defaults. Could do that now or wait until we use this somewhere else.

src/endpoints/seed/pages/page.ts
```diff
@@ -113,6 +113,7 @@ export const page: (
           },
         ],
         blockType: 'content',
+        enableColumns: true,
```
busbyk: Since there's only one column here I think it makes sense to set this to false.

src/blocks/Content/Component.tsx
```diff
@@ -1,41 +1,49 @@
...
+            columns?.map((col, index) => {
+              const { richText } = col
 
-                {enableLink && <CMSLink {...link} />}
-              </div>
-            )
-          })}
+              return (
+                <div
+                  className={cn(`col-span-4 md:col-span-2 ${colsSpanClass} ${textColor}`)}
```
busbyk: Def a bit of a preference thing in this scenario since colsSpanClass and textColor will never be undefined but I find this a little easier to write too. Up to you!

```suggestion
                  className={cn('col-span-4 md:col-span-2', colsSpanClass, textColor)}
```

Could update on line 42 as well if you like this style better.

src/blocks/Content/Component.tsx
```diff
@@ -1,41 +1,49 @@
...
+    1: 'lg:col-span-12',
+    2: 'lg:col-span-6',
+    3: 'lg:col-span-4',
+    4: 'lg:col-span-3',
   }
+  const colsSpanClass = colsClasses[numOfCols]
 
+  // TODO - import color list from theme
+  const bgColor = Color(`${color}`)
+  const textColor = bgColor.isLight() ? 'text-black' : 'text-white'
```
busbyk, line 23: Ooh this is very cool. Could be really helpful for https://github.com/NWACus/web/issues/32?issue=NWACus%7Cweb%7C72

With the #157 idk if we'll need this here because each theme background class should have an associated -foreground class to use. But if we allow the use of the default tailwind colors we might need this.

busbyk (changes requested):
2. Do we ever want to allow more than two columns? Maybe a layout field with radio options of one column and two column with the default being one column is a better UX. In the designs we mostly see single columns.
3. I think safelist is fine for this PR but when we start using theme colors we shouldn't need safelist at all. And we also should need to use hex colors - we can just reference the theme class names for the color picker. We'll need to set the avalanche center's center id (i.e. same as the theme name) as a class on the color picker container to have the correct colors show up for the theme. 
4. I'm good with altering this content block. Let's modify the payload website template as much as we need.

busbyk:
@rchlfryn Do you have an issue created for the follow up work of using theme colors? I can leave some comments on there with some context from chatting with Chris.

src/fields/color.ts
rchlfryn: Easy enough fix. done in [47e345c](https://github.com/NWACus/web/pull/153/commits/47e345c51d4c880cbe3d92dee6511c8bdba7cfe0)

src/blocks/Content/Component.tsx
rchlfryn: Makes sense to me to separate them by a comma.

src/blocks/Content/Component.tsx
rchlfryn: I will keep it for now and we can remove and maybe reuse it later.

rchlfryn:
> @rchlfryn Do you have an issue created for the follow up work of using theme colors? I can leave some comments on there with some context from chatting with Chris.

@busbyk I do not have an open issue. I think you have a bit more context on colors at the moment so if you could make one and leave a comment here once you do.

> 2. Do we ever want to allow more than two columns? Maybe a layout field with radio options of one column and two column with the default being one column is a better UX. In the designs we mostly see single columns.

We can get up to 4 columns. I wish there was an easy way to pre-populate the columns based on the radio selection. Maybe we could do this in an `action`?

src/endpoints/seed/pages/page.ts
```diff
@@ -76,9 +76,9 @@ export const page: (
     },
     layout: [
       {
+        color: 'white',
         columns: [
           {
-            size: 'full',
```
stevekuznetsov, line 81: @rchlfryn this looks like Content > Layout 1 > Content > RichText - maybe this is the error?

stevekuznetsov:
```
[18:39:14] INFO: Creating pages['nwac']['donate-membership'] with hash 39402c69...
[18:39:14] ERROR: The following field is invalid: Content > Layout 1 > Content > RichText
    err: {
      "type": "ValidationError",
      "message": "The following field is invalid: Content > Layout 1 > Content > RichText",
      "stack":
          ValidationError: The following field is invalid: Content > Layout 1 > Content > RichText
```

stevekuznetsov: Wait no, rich text is after this diff and has no changes ...

busbyk:
> > @rchlfryn Do you have an issue created for the follow up work of using theme colors? I can leave some comments on there with some context from chatting with Chris.
> 
> @busbyk I do not have an open issue. I think you have a bit more context on colors at the moment so if you could make one and leave a comment here once you do.
> 

Cool, created #179. This is based on conversations I had with Chris. We may not want to accept _all_ of the theme classes and obv open to discussion on this since you haven't had a chance to give input 🤘

busbyk:
> > 2. Do we ever want to allow more than two columns? Maybe a layout field with radio options of one column and two column with the default being one column is a better UX. In the designs we mostly see single columns.
> 
> We can get up to 4 columns. I wish there was an easy way to pre-populate the columns based on the radio selection. Maybe we could do this in an `action`?

Looks like we _could_ write a custom component to manage the number of columns based on the radio selection using [`useForm`'s {add/remove/replace}Field functions](https://payloadcms.com/docs/admin/react-hooks#useform:~:text=addFieldRow,or%20block%20field). 

That could end up being a bit of a rabbit hole at this stage so I think I'm fine with it how it is. What do you think?

stevekuznetsov:
> Looks like we could write a custom component

Sounds like awesome V2 issue :sunglasses:

rchlfryn:
> Looks like we _could_ write a custom component to manage the number of columns based on the radio selection using [`useForm`'s {add/remove/replace}Field functions](https://payloadcms.com/docs/admin/react-hooks#useform:~:text=addFieldRow,or%20block%20field).
> 
> That could end up being a bit of a rabbit hole at this stage so I think I'm fine with it how it is. What do you think?

~@busbyk Made [an issue](https://github.com/orgs/NWACus/projects/8?pane=issue&itemId=109868404&issue=NWACus%7Cweb%7C181) for it for v2.~

Ended up removing the `enableColumn` toggle

src/blocks/Content/config.ts
```diff
@@ -9,54 +9,46 @@ import {
...
-    // Show ability to add columns if checkbox is enabled
     {
       name: 'columns',
-      type: 'array',
-      admin: {
-        condition: (_, siblingData) => siblingData.enableColumns,
+      label: false,
+      labels: {
+        plural: 'Columns',
+        singular: 'Columns',
```
busbyk: Do you want this to be plural instead of singular? (is this intentional?)

busbyk (approved):
Just left one comment but looks good!

src/blocks/Content/config.ts
rchlfryn: thank you!

PR #154, opened 2025-04-29
Seed pages

busbyk:
- Removes home static page (we have a landing page at [center]/)
- Renames who we are page to staff page based on [Navigation Content Analysis](https://docs.google.com/spreadsheets/d/1I-FM6WudxPnbGfEHZA7glMPv-T3QFPVBf7HkPMuCdxo/edit?gid=0#gid=0)
- Moves page helpers into their own directory
- Removes pages from POC and replaces them with pages from the Navigation Content Analysis
- Reorders collections for deletion. Collections with required fields that are relations must be deleted before those related collections or we get NOT NULL errors when seeding a previously seeded db.


stevekuznetsov (approved):

stevekuznetsov:
Weird error on preview ... https://github.com/NWACus/web/actions/runs/14746833233/job/41395674510#step:15:42

src/endpoints/seed/index.ts
```diff
@@ -970,106 +969,201 @@ export const innerSeed = async ({
...
+          'Volunteer',
+          'Interested in volunteering your time for the center? We are always looking for help at events and with various projects.',
+          'volunteer',
         ),
         page(
           tenant,
-          images[tenant.name]['home'],
+          images[tenant.name]['hero'],
           images[tenant.name]['image2'],
           'About Us',
```
stevekuznetsov, line 1019: @busbyk sorry, missed this - you left in this page, which overrides the other one.

busbyk:
> Weird error on preview ... https://github.com/NWACus/web/actions/runs/14746833233/job/41395674510#step:15:42

Yea that's super strange - couldn't reproduce locally and looks like it didn't happen again on build once merged 🤔

src/endpoints/seed/index.ts
busbyk: I renamed the page you included the biographies and teams blocks on to "staff" so this is the only "about-us" page (unless I'm totally missing something). But there is a duplicate "staff" page here now which I just removed: https://github.com/NWACus/web/pull/156

PR #156, opened 2025-04-30
Removing duplicate staff page

busbyk:
`src/endpoints/seed/pages/staff-page.ts` already creates the staff page (previously who-we-are but updated based on Navigation Content Analysis). 

Screenshot 2025-04-30 at 8 04 51 AM

Update: We actually still want the "staff" page to be called Who We Are. So reverting that back to that naming.


stevekuznetsov:
Lgtm

stevekuznetsov (approved):

PR #157, opened 2025-04-30
Semantic class names for themes + NWAC theme

busbyk:
See the decision doc for full explanation. The list of semantic names is a collaboration between Chris and I but all feedback is welcome. 

Leaving this as a draft for now because Chris may change the NWAC colors a bit.


docs/decisions/006-semantic-class-names-for-themes.md
```diff
@@ -0,0 +1,176 @@
...
+      primary: {
+        DEFAULT: 'hsl(var(--primary))',
+        foreground: 'hsl(var(--primary-foreground))',
+      }
+    }
+  }
+}
+```
+
+This setup let's you use classnames like `bg-primary` and `text-primary-foreground` that can result in different colors depending on the value of the CSS Variable. This is how Shadcn and Tailwind allows for theming while being able to use the same class names in your code.
```
stevekuznetsov:
```suggestion
This setup lets you use classnames like `bg-primary` and `text-primary-foreground` that can result in different colors depending on the value of the CSS Variable. This is how Shadcn and Tailwind allow for theming while being able to use the same class names in your code.
```

docs/decisions/006-semantic-class-names-for-themes.md
```diff
@@ -0,0 +1,176 @@
...
+| destructive           | destructive-foreground   |                                       |
+| border                | -                        |                                       |
+| input                 | -                        |                                       |
+| ring                  | -                        |                                       |
+| radius                | -                        | Not a color, but included for completeness |
+| header                | header-foreground        | **Custom**                            |
+| footer                | footer-foreground        | **Custom**                            |
+| callout               | callout-foreground       | **Custom**                            |
+| brand-50 ... brand-950| brand-foreground-50 ... brand-foreground-950 | **Custom** Brand color scale         |
+| success               | -                        | **Custom** Status color               |
```
stevekuznetsov, line 94: Strikes me that most dashboard users of Shadcn will have a need to communicate info/success/warning/error in toasts - what does e.g. PayloadCMS use? Do they map those into the semantic names that Shadcn comes with, or do they also need to add on?

docs/decisions/006-semantic-class-names-for-themes.md
```diff
@@ -0,0 +1,176 @@
...
+This approach gives us a consistent set of semantic class names for theming, while allowing flexibility for our own design needs.
+
+### Toggling themes
+
+We will use this CSS Variables pattern to toggle themes between avalanche centers. Each avalanche center could have a light mode and a dark mode.
+
+We would have a theme defined in our globals.css for each avalanche center like:
+
+```
+// globals.css
```
stevekuznetsov, line 111: Are we not tackling colors chosen by admins in Payload out of principle, or to make our progress today expedient?

stevekuznetsov: Oops sorry, didn't read all the way to the bottom, disregard.

tailwind.config.mjs
```diff
@@ -49,26 +49,12 @@ export default {
...
-          foreground: 'hsl(var(--accent-foreground))',
-        },
         background: 'hsl(var(--background))',
-        border: 'hsla(var(--border))',
+        foreground: 'hsl(var(--foreground))',
         card: {
           DEFAULT: 'hsl(var(--card))',
           foreground: 'hsl(var(--card-foreground))',
         },
-        destructive: {
```
stevekuznetsov, line 62: Just to confirm - you sorted these, or changed the values, or both?

stevekuznetsov (approved):
Looks great - feel free to merge in early, it's super easy to change the exact colors up if Chris updates them.

docs/decisions/006-semantic-class-names-for-themes.md
```diff
@@ -0,0 +1,176 @@
+# Using Semantic Class Names for Themes
+
+Date: 2025-04-30
+
+Status: accepted
+
+## Context
+
+[Shadcn uses CSS Variables](https://ui.shadcn.com/docs/theming#css-variables) to enable theming based on [Tailwind's suggested approach](https://v3.tailwindcss.com/docs/customizing-colors#using-css-variables) to defining colors using CSS Variables.
```
rchlfryn: This was confusing to me. I dont know if what i wrote is any better... 😂 

```suggestion
Enabling [Shadcn use of CSS Variables](https://ui.shadcn.com/docs/theming#css-variables) for theming, follows [Tailwind’s recommended pattern](https://v3.tailwindcss.com/docs/customizing-colors#using-css-variables) of defining and consuming colors as CSS variables.
```

rchlfryn (approved):

tailwind.config.mjs
busbyk: Sorted and added.

docs/decisions/006-semantic-class-names-for-themes.md
busbyk: How about:

```suggestion
[Shadcn uses CSS Variables](https://ui.shadcn.com/docs/theming#css-variables) for theming, which follows [Tailwind’s recommended pattern](https://v3.tailwindcss.com/docs/customizing-colors#using-css-variables) of defining and consuming colors as CSS variables.
```

I guess we're enabling in a sense but I'm just stating a fact about Shadcn and Tailwind for background. I like the rest of the suggestion!

docs/decisions/006-semantic-class-names-for-themes.md
busbyk: Yea tbh I don't know where we'll use these ones but they make sense to keep. They were included in the payload website template and the only use I can find is in the https://github.com/NWACus/web/blob/main/src/blocks/Banner/Component.tsx 

I don't think these are used by Shadcn at all. I couldn't find any usages here: https://github.com/shadcn-ui/ui/tree/c41c6ece86b5de472668729ed336ee4aac96ee16/apps/www/registry/default/ui

So they're really just for that one banner component at the moment.

docs/decisions/006-semantic-class-names-for-themes.md
busbyk: @rchlfryn if you like this feel free to commit it and merge 🚀

rchlfryn: Done :shipit:

stevekuznetsov:
@busbyk rebased these commits on top of latest `main` to remove conflicts

Issue #50, opened 2025-03-15
Publish Updated Sitemaps


stevekuznetsov:
Once we have determined the appropriate navigation experience, we need to ensure our sitemap generators are spitting out the full set of pages for a site.

xref https://github.com/NWACus/web/issues/2

busbyk:
Note for future self: Should take care of this TODO under this issue: https://github.com/NWACus/web/blob/website-template-cleanup/src/app/(frontend)/%5Bcenter%5D/forecasts/avalanche/%5Bzone%5D/page.tsx#L25

Issue #98, opened 2025-04-10
Payload website template cleanup and updates


busbyk:
There are still a few things left over from the Payload website template [Payload website template](https://github.com/payloadcms/payload/tree/main/templates/website) that we're not using like:

- [x] A concept of heros
- [x] Their link field (might just need slight modification)
- [x] Some built-in blocks
- [x] Logo component
- [x] The "home" page concept
- [x] Remove search plugin

And there are defaults left over that we should update like:
- [x] OpenGraph data
- [x] Default meta data
- [x] Image loading placeholders

Other
- [x] Hide seeding button in production

There have also been updates to the template since we used it as a starting point. 

It would be great to diff the current version with the version we started with and pull over any changes that we think are worth updating in our version. 

This task might also be a good opportunity to try to fix some of the `ts-ignore` comments we've had to use.

busbyk:
Might as well address: #192 and do a little work on the app skeleton (like making sure our <body> is at least full height of the viewport so the footer always appears at the bottom).

PR #158, opened 2025-05-01
Frontend Navigation

### LINKED ISSUE #89: Display header and navigation on frontend

busbyk:
Resolves #89 

Background on navigation-menu viewport animation changes: shadcn's navigation-menu viewport doesn't position under the active/hovered menu item. This is a [long-running issue](https://github.com/shadcn-ui/ui/issues/94) which is also [present in Radix](https://github.com/radix-ui/primitives/issues/1462#issuecomment-2275683692). I ended up going with a similar workaround to [this](https://github.com/radix-ui/primitives/issues/1462#issuecomment-2275683692). 

This PR:
- Moves existing header into components folder
- Fixes forecast typo
- Uses pages from the [Navigation Content Analysis](https://docs.google.com/spreadsheets/d/1I-FM6WudxPnbGfEHZA7glMPv-T3QFPVBf7HkPMuCdxo/edit?gid=1536438174#gid=1536438174) for navigation seeding
- Renames staffPage to whoWeAre page to match it's slug and title
- Installs and modifies the navigation-menu components from shadcn assuming that this will only be used for the main navigation
- Installs accordion from shadcn
- Adds a callout variant to button for use in the header (donate button)
- Builds out mobile and desktop navigation with:
    - Nesting of urls under their parent nav items for internal links
    - Active link highlighting for internal links
    - External link icon indicator for external links
- Changes the tailwind config file reference in `components.json`. I needed this to install shadcn components correctly. 

This PR does not update the `src/app/(frontend)/[center]/[...segments]/page.tsx` route handler or the sitemap to determine the nesting of a page based on its position in the navigation. I'd like to do that under #50. 

Links to pages currently work even though the urls are nested because `src/app/(frontend)/[center]/[...segments]/page.tsx` currently queries based on the last path part.


src/components/Header/DesktopNav.client.tsx
```diff
@@ -0,0 +1,140 @@
+'use client'
+import { cn } from '@/utilities/ui'
+import { ChevronDown } from 'lucide-react'
+import { useEffect, useRef, useState } from 'react'
+import { Button } from '../ui/button'
```
stevekuznetsov, line 5: Silly question, but what's the difference between our `@/path/whatever` imports and these relative path ones?

src/components/Header/DesktopNav.client.tsx
```diff
@@ -0,0 +1,140 @@
...
+export const DesktopNav = ({ topLevelNavItems }: { topLevelNavItems: TopLevelNavItem[] }) => {
+  const [activeMenuItem, setActiveMenuItem] = useState<string>()
+  const containerReference = useRef<HTMLElement>(null)
+
+  useEffect(() => {
+    const container = containerReference.current
+
+    if (!container) return
+
+    const updatePosition = (item: HTMLElement) => {
```
stevekuznetsov: Is this the fix for the Radix rendering at (0,0) bug you were talking about? Is there any chance we could propose the change upstream or somehow cross-link this code to the open issues they have for it?

src/components/Header/DesktopNav.client.tsx
```diff
@@ -0,0 +1,140 @@
...
+                      )
+                    })}
+                  </div>
+                </NavigationMenuContent>
+              </NavigationMenuItem>
+            )
+          })}
+          <NavigationMenuItem>
+            <NavigationMenuLink className={cn(navigationMenuTriggerStyle(), 'py-0')} asChild>
+              <RenderNavLink link={topLevelNavItems[topLevelNavItems.length - 1].link}>
```
stevekuznetsov: What gives us the guarantee that the last item in this array is the donate link? Can we break that out in the schema so we don't need to make such an assumption? e.g. array of links separate from explicit donate link

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { ImageMedia } from '../Media/ImageMedia'
+import { DesktopNav } from './DesktopNav.client'
+import { MobileNav } from './MobileNav.client'
+import { getTopLevelNavItems } from './utils'
+
+export async function Header({ center }: { center?: string }) {
+  const { isEnabled: draft } = await draftMode()
+  const payload = await getPayload({ config: configPromise })
+
+  const media = await payload.find({
```
stevekuznetsov:
```suggestion
  const brand = await payload.find({
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,78 @@
...
+  })
+
+  if (media.docs.length < 1) {
+    payload.logger.error(`Brand for tenant ${center} missing`)
+  }
+
+  const banner = media.docs[0]?.banner
+
+  if (media.docs.length > 0 && typeof banner !== 'object') {
+    payload.logger.error(`Banner for tenant ${center} missing`)
```
stevekuznetsov: I wonder - can we also `throw` something here (and generally in all cases where we're doing this defensive log), so that `pnpm build` errors out and we catch this in CI?

src/components/Header/MobileNav.client.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { TopLevelNavItem } from './utils'
+
+export const MobileNav = ({
+  topLevelNavItems,
+  banner,
+}: {
+  topLevelNavItems: TopLevelNavItem[]
+  banner?: Media
+}) => {
+  const [mobileNavOpen, setMobileNavOpen] = useState(false)
```
stevekuznetsov, line 31: Does tapping outside of the menu close it?

src/components/Header/RenderNavLink.tsx
```diff
@@ -0,0 +1,53 @@
...
+      >
+        {children || link.label}
+      </Link>
+    )
+  }
+
+  if (link.type === 'internal-reference' && link.reference) {
+    const url =
+      link.reference.relationTo === 'pages'
+        ? `/${typeof link.reference.value === 'object' ? link.reference.value.slug : link.reference.value}`
```
stevekuznetsov: When `value` is not an object, why is it the slug itself?

src/components/Header/RenderNavLink.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  if (link.type === 'internal-relative') {
+    return (
+      <Link href={link.url} className={className} onClick={onClick}>
+        {children || link.label}
+      </Link>
+    )
+  }
+
+  // Fallback
```
stevekuznetsov: If this should never happen and the `link.type` is strongly-typed you could use a `switch` statement with a `never` at the bottom.

src/components/Header/RenderNavLink.tsx
stevekuznetsov: Oh, I see - sorry. `link` is a type you defined.

src/components/Header/utils.ts
```diff
@@ -0,0 +1,265 @@
+import { Navigation, Page, Post } from '@/payload-types'
+
+export type NavLink =
+  | {
+      type: 'internal-reference'
+      label: string
+      reference:
+        | { relationTo: 'pages'; value: number | Page }
+        | { relationTo: 'posts'; value: number | Post }
```
stevekuznetsov: Yeah, looks like `value` is a reference if it's not a filled-in object, so using the number as the slug won't work in the code above - likely better to use the log/throw bail-out.

src/components/Header/utils.ts
```diff
@@ -0,0 +1,265 @@
...
+  // Handle internal links with references
+  if (
+    link.type === 'internal' &&
+    link.reference &&
+    (link.reference.relationTo === 'pages' || link.reference.relationTo === 'posts')
+  ) {
+    const reference = link.reference
+
+    // If reference.value is a number, don't render this nav item
+    if (typeof reference.value === 'number') {
```
stevekuznetsov: Hm, but the logic here already does this - so it looks like you may be able to do better type narrowing on the output of this function?

src/components/Header/utils.ts
```diff
@@ -0,0 +1,265 @@
...
+      url: link.url,
+      newTab: !!link.newTab,
+    }
+  }
+
+  // Handle internal links with references
+  if (
+    link.type === 'internal' &&
+    link.reference &&
+    (link.reference.relationTo === 'pages' || link.reference.relationTo === 'posts')
```
stevekuznetsov, line 152: n.b. you could also formulate the output type as simply having a relative path and using it as an opaque value, in this function you know if it's a relation to pages or posts and can format the full relative path using the slug, then simplify the output type significantly to not need a `reference` sub-object, just a path.

src/components/Header/utils.ts
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    // Try to get title from reference value if it exists
+    if (
+      !linkLabel &&
+      typeof reference.value === 'object' &&
+      reference.value &&
+      'title' in reference.value
+    ) {
+      linkLabel = (reference.value as any).title
```
stevekuznetsov: Why do we need `as any` here?

src/components/Header/utils.ts
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    // If we still don't have a label, don't render this nav item
```
stevekuznetsov: IIRC `title` is required on pages and posts, so we should `throw` here and hard-crash during CI or whatever, it would mean we somehow invalidated fundamental assumptions

src/components/Header/utils.ts
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  // Process dropdown items if they exist
+  if (tab.items && tab.items.length > 0) {
+    result.items = tab.items
+      .map((item) => {
+        if (!item) return null
+
+        const navItem: NavItem = {
+          id: item.id || generateRandomId(),
```
stevekuznetsov: Similarly here - in what case would we not have `item.id`? A random ID here is of little use since it will change on every render, I would prefer if we can to tighten up the requirements on `item.id` or throw here.

stevekuznetsov (commented):
Looking awesome!

src/components/Header/DesktopNav.client.tsx
busbyk: @ is just an alias which always resolves to /src in our case. Configured in our tsconfig: https://github.com/NWACus/web/blob/9ee93564436afe109f9f2f235f8e56b4906ccf31/tsconfig.json#L36

Alias imports are nice because if I moved this file around I wouldn't have to update the relative imports. I import using `Cmd + .` - I bet I could configure that to prefer using aliases over relative imports. That would be nice.

src/components/Header/DesktopNav.client.tsx
busbyk: Yea this moves the navigation viewport (i.e. dropdown) to be positioned under the active navigation item. 

I can leave a comment with a link to the Radix issue. 

Here are my notes I was planning to include in my description when this is ready for review:

shadcn's navigation-menu viewport doesn't position under the active/hovered menu item. This is a [long-running issue](https://github.com/shadcn-ui/ui/issues/94) which is also [present in Radix](https://github.com/radix-ui/primitives/issues/1462#issuecomment-2275683692). I ended up going with a similar workaround to [this](https://github.com/radix-ui/primitives/issues/1462#issuecomment-2275683692).

src/components/Header/DesktopNav.client.tsx
busbyk: It's explicitly positioned last here: https://github.com/NWACus/web/blob/cb4a347a6f03f9ccceb1be428ea18a65cc5b0d2d/src/components/Header/utils.ts#L264

I also do not like the slice() and this last item reference just because it's not super clear when reading the code. I'll make it more explicit for someone reading this by setting separate variables here.

src/components/Header/Header.tsx
busbyk: Oh yea good call. Just copy pasted from how it was before. But will update.

src/components/Header/DesktopNav.client.tsx
stevekuznetsov: There may be an eslint config that enforces the alias ones - IIRC we do it in the mobile app repo, then `--fix` also re-formats everything correctly.

stevekuznetsov: IIRC it's these plugins - https://github.com/NWACus/avy/blob/5ad60e5af5c17acba207180af1088c32cca94d65/.eslintrc.json#L10 

We may appreciate the `sort-keys` one as well if we can get it to sort .e.g the arrays in payload config

src/endpoints/seed/navigation.ts
```diff
@@ -47,117 +47,123 @@ export const navigationSeed = (
...
       items: [
         {
           link: pageLink({ slug: 'about-us' }),
         },
         {
-          link: pageLink({ slug: 'about-the-forecasts' }),
+          link: pageLink({ slug: 'agency-partners' }),
+        },
+        {
+          link: pageLink({ slug: 'staff' }),
```
stevekuznetsov: this renders as 'missing page'

stevekuznetsov: Actually noticed there are a lot of missing page warnings other than this - I imagine this specific case just needs to be updated to `who-we-are` but wanted to confirm that the rest are expected - I think you were just laying out pages that need to exist in the future but do not yet?

```
[15:32:00] WARN: Page radar-and-satellite-index not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page avalanche-awareness-classes not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page courses-by-local-providers not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page snowpack-scholarship not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page about-us not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page about-us not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page become-a-member not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page workplace-giving not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page corporate-sponsorships not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Northwest Avalanche Center
[15:32:00] WARN: Page radar-and-satellite-index not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page avalanche-awareness-classes not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page courses-by-local-providers not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page snowpack-scholarship not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page about-us not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page about-us not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page become-a-member not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page workplace-giving not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page corporate-sponsorships not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Sierra Avalanche Center
[15:32:00] WARN: Page radar-and-satellite-index not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page avalanche-awareness-classes not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page courses-by-local-providers not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page snowpack-scholarship not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page about-us not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page about-us not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page become-a-member not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page workplace-giving not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page corporate-sponsorships not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Sawtooth Avalanche Center
[15:32:00] WARN: Page about-the-forecasts not found for tenant Sawtooth Avalanche Center
```

src/components/Header/DesktopNav.client.tsx
stevekuznetsov: If it's reasonable to split it out into `{topLevelNavItems: NavItem[], donateButton: Button}` or whatever at the data store level, it would make it less likely to mess this up in the future.

src/components/Header/utils.ts
busbyk: Was just making typescript happy since Payload's types indicate that id could be undefined. But I can just use a non-null assertion instead.

src/components/Header/utils.ts
busbyk: Oops, we don't!

src/components/Header/MobileNav.client.tsx
busbyk: Yep! There's a fixed overlay for that: https://github.com/NWACus/web/blob/693e1abbd65bac20c7ff0cd0c1434e4e62012ee0/src/components/Header/MobileNav.client.tsx

src/endpoints/seed/navigation.ts
busbyk: I'm only getting a missing page on staff page. Pushed a fix for that. Once this is ready for review, let me know if you're still seeing these errors.

stevekuznetsov: (I'm getting those errors on `main` :D)

src/components/Header/utils.ts
```diff
@@ -54,9 +55,10 @@ function topLevelNavItem(
       .map((item) => {
         if (!item) return null
 
+        invariant(item.id, `Tab ${label} has an item without an id.`)
```
stevekuznetsov, line 57: This is very cool

busbyk: Yea it's a nice little package!

src/app/(frontend)/globals.css
```diff
@@ -77,6 +77,8 @@
     --success: 196 52% 74%;
     --warning: 34 89% 85%;
     --error: 10 100% 86%;
+
+    --radix-navigation-menu-item-active-left: 0px;
```
stevekuznetsov: Why does this have to live in the global file?

src/components/Header/utils.ts
```diff
@@ -0,0 +1,255 @@
+import { Navigation, Page, Post } from '@/payload-types'
+import invariant from 'tiny-invariant'
+
+export type NavLink =
+  | {
+      type: 'internal'
+      label: string
+      url: string
+      newTab?: undefined
```
stevekuznetsov: Why do we need this field?

src/components/Header/utils.ts
```diff
@@ -0,0 +1,255 @@
...
+      type: 'external'
+      label: string
+      url: string
+      newTab: boolean
+    }
+
+export type NavItem = {
+  id: string
+  link?: NavLink
+  items?: NavItem[]
```
stevekuznetsov, line 20: Is there a semantic difference between a zero-length array of items and `undefined`? Do we need to allow both?

src/components/Header/utils.ts
```diff
@@ -0,0 +1,255 @@
...
+            navItem.link = convertedLink
+          }
+        }
+
+        if (item.items && item.items.length > 0) {
+          navItem.items = item.items
+            .map((nestedItem) => {
+              if (!nestedItem) return null
+
+              invariant(nestedItem.id, `Tab ${label} > item ${item.id} has an item without an id.`)
```
stevekuznetsov:
```suggestion
              invariant(nestedItem.id, `Tab ${label}[${item.label}][${index}] has no id.`)
```

Should be sufficient info to quickly track down exactly the misbehaving data

src/components/Header/utils.ts
```diff
@@ -0,0 +1,255 @@
...
+                  ...(navItem.link ? [navItem.link.label.toLowerCase()] : []),
+                ])
+                if (convertedNestedLink) {
+                  nestedNavItem.link = convertedNestedLink
+                }
+              }
+
+              return nestedNavItem.link ? nestedNavItem : null
+            })
+            .filter(Boolean) as NavItem[]
```
stevekuznetsov, line 93: If we did the Typescript correctly we should not need the `as` - what's the complaint from `tsc` without it?

src/components/Header/utils.ts
```diff
@@ -0,0 +1,255 @@
...
+        `Could not determine label for link with reference ${JSON.stringify(reference)}`,
+      )
+    }
+
+    let url =
+      link.reference.relationTo === 'pages'
+        ? `/${reference.value.slug}`
+        : `/posts/${reference.value.slug}`
+
+    if (parentItems?.length && parentItems.length > 0) {
```
stevekuznetsov, line 180: I'd need to check my math with jsfiddle but I think the `.join()` should be safe to run for any array (even zero-length)

src/app/(frontend)/globals.css
busbyk: Good call, it does not!

I'm actually going to remove this completely. The intent was to have a default value for this but after some testing it's defaulting to 0 anyways.

src/components/Header/utils.ts
busbyk: Yea I could default parentItems to an empty array and always run the `.join()` but I'd still need to check if it has items to make sure I have slashes in the right place. We always want to have a leading slash.

src/components/Header/utils.ts
busbyk: Typescript doesn't infer the correct type when using `.filter(Boolean)` - it still thinks that the array could contain nulls. Pretty common to use as NavItem in this case (I think it's easier to understand) but I updated to use a type predicate.

src/components/Header/utils.ts
busbyk: I suppose we don't!

src/components/Header/utils.ts
busbyk: Not a huge difference - just different in how you check for it. I prefer checking to see if items exists in the desktop and mobile nav rather than checking for it's length. There's some UI I don't want to render if there are no items.

.github/workflows/preview.yaml
```diff
@@ -18,8 +18,11 @@ jobs:
       - name: Create a new database for the preview
         id: database
         run: |
-          ref="${{ github.head_ref }}"
-          name="payloadcms-preview-${ref//[^a-z0-9\-]/x}"
+          head_ref="${{ github.head_ref }}"
```
rchlfryn: ⛏️ Great add - maybe put it in a new PR or call it out in the description so it looks intentional if someone has to troubleshoot it.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import { draftMode } from 'next/headers'
+import Link from 'next/link'
+import { ImageMedia } from '../Media/ImageMedia'
+import { DesktopNav } from './DesktopNav.client'
+import { MobileNav } from './MobileNav.client'
+import { convertToNavLink, getTopLevelNavItems, TopLevelNavItem } from './utils'
+
+export async function Header({ center }: { center?: string }) {
+  const { isEnabled: draft } = await draftMode()
```
rchlfryn, line 13: cool deconstruction. I haven't seen this before. maybe keeping `is` to easily show this is a boolean.. `isDraft`?

src/components/Header/DesktopNav.client.tsx
```diff
@@ -0,0 +1,207 @@
...
+                  asChild
+                  active={navItem.link?.type === 'internal' && navItem.link?.url === pathname}
+                >
+                  <RenderNavLink
+                    link={navItem.link}
+                    className={cn(
+                      navigationMenuTriggerStyle(),
+                      navItem.link?.type === 'internal' &&
+                        navItem.link?.url === pathname &&
+                        'text-white hover:text-white/90',
```
rchlfryn: I noticed a weird bug on the main nav. If you click a dropdown menu item, it will remain white even if you are not on that page. Is it coming from `navigationMenuTriggerStyle()`?

Otherwise it could lead to a confusing user experience.
Northwest-Avalanche-Center---Avalanche-Observations

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const payload = await getPayload({ config: configPromise })
+
+  const brands = await payload.find({
+    collection: 'brands',
+    depth: 10,
```
rchlfryn: is there a reason you picked 10?

src/components/Header/MobileNav.client.tsx
```diff
@@ -0,0 +1,131 @@
...
+        />
+        <DialogContent className="md:hidden max-h-[calc(100vh-64px)] overflow-y-auto fixed z-40 bg-header text-header-foreground pb-2 shadow-lg transition ease-in-out data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:duration-300 data-[state=open]:duration-500 inset-x-0 top-[64px] border-b border-b-white data-[state=closed]:slide-out-to-top data-[state=open]:slide-in-from-top">
+          <DialogTitle className="sr-only">menu</DialogTitle>
+          <DialogDescription className="sr-only">navigation menu</DialogDescription>
+          <Accordion type="single" collapsible asChild>
+            <nav className="divide-y divide-header-foreground/20 px-2">
+              {topLevelNavItems.map((navItem) => {
+                const label = navItem.label ?? navItem.link?.label
+
+                if (!label) return null
```
rchlfryn: I saw this in the Desktop file as well. When will this be true?

src/components/Header/MobileNavItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+          className,
+        )}
+        onClick={() => setMobileNavOpen(false)}
+      />
+    ) : (
+      <span className={className}>{label}</span>
+    )
+  }
+
+  if (!navItem.items || navItem.items.length === 0) return <span>{label}</span>
```
rchlfryn: I dont think we will ever get to this line. It seems like this is returned in line 48 and will never get here.

.github/workflows/preview.yaml
```diff
@@ -50,12 +53,27 @@ jobs:
         shell: bash
       - name: Install Vercel CLI
         run: npm install --global vercel@latest
+      - name: Generate preview domain names
```
stevekuznetsov: n.b. if you `git rebase` instead of making merge commits, this stuff doesn't show up and the contents of the PR stay minimal to what is net new on top of `main`

src/components/Header/utils.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  if (
+    link.type === 'internal' &&
+    link.reference &&
+    (link.reference.relationTo === 'pages' || link.reference.relationTo === 'posts')
+  ) {
+    const reference = link.reference
+
+    if (typeof reference.value === 'number') {
+      throw new Error(
```
stevekuznetsov: How do you decide between this and `tiny-invariant`?

rchlfryn (changes requested):
Looks great!

.github/workflows/preview.yaml
busbyk: I might not be understanding correctly but this change is from #170 which was targeted at this branch. So when that got merged those changes landed here and these changes aren't in `main` so it makes sense that they're showing up right?

#170 probably should have targeted `main`. I wanted to have the navigation changes in that branch to test client side navigation was working correctly, though. 

I suppose I could what cherry pick these changes out of this PR and target `main`? I'm not very good with cherry pick, could use a hand if that's the right approach.

.github/workflows/preview.yaml
busbyk: Yea this is from #170 - see context: https://github.com/NWACus/web/pull/158#discussion_r2080355882

.github/workflows/preview.yaml
stevekuznetsov: Aaaah I didn't realize that targeted this and not `main` - sorry! I'm more used to a linear flow where stuff ends up in `main` from one merge. No big deal.

src/components/Header/Header.tsx
busbyk: This is from the Payload website template. We have it in several places:

CleanShot 2025-05-08 at 12 36 55@2x

I think the intent is to be able to pass it directly to a payload.find call:
```
await payload.find({
  ...
  draft,
  ...
})
```

rather than:
```
await payload.find({
  ...
  draft: isEnabled,
  ...
})
```

We could update the naming in #98 if you have a preference?

src/components/Header/DesktopNav.client.tsx
busbyk: Ah yea it's because it's in a focus state after you click it. I can modify the focus state in navigationMenuTriggerStyle. We still need to support the focus state for accessibility reasons though so maybe just a more opaque white or I could add a ring/outline for the focus state. I typically don't love how ring/outline looks though.

src/components/Header/Header.tsx
busbyk: Oo good question. This is just what it was set to before: https://github.com/NWACus/web/blob/6501249b739a14d3b22b7a4d68f5e8d1fefa5053/src/Header/Component.tsx#L13

I'll update this to be the smallest depth value while still populating all sub items' links.

src/components/Header/Header.tsx
stevekuznetsov: (only semi-related: non-JavaScript Steve dislikes implicit object field notation like `{ draft }` since go-to-definition goes to the definition of your variable, not the definition of `draft` field in the input props to `payload.find()` - is there a workflow that you use to get around that?)

src/components/Header/Header.tsx
stevekuznetsov: That risks making data unset if the depth changes, why not 999999? If we set to 10, and it needs 11, we will see a build fail (hopefully from the `tiny-invariant` bits) and update to 11, which seems like busywork.

src/components/Header/MobileNavItem.tsx
busbyk: You're so right!

src/components/Header/MobileNav.client.tsx
busbyk: Hmm maybe I could write the type for TopLevelNavItem better but I think TS isn't recognizing that a TopLevelNavItem must have either a label or a link. Weirdly it's indicating that the link label could be either string or undefined when it's typed as being string. 

I think I'm fine with just changing this to use invariant to warn us if a label is ever missing but I don't think that should happen.

src/components/Header/Header.tsx
busbyk: Ooh interesting. No I don't have a workflow to get around that. go-to-definition on `.find` and then go-to-definition on `.find`'s options I suppose.

src/components/Header/Header.tsx
busbyk: Potentially just a performance hit. If pages end up having large relations that get all get included in the response.

busbyk: Oh sorry yanno what it doesn't really matter for brands since there aren't that many nested relationships. Was thinking about the navigation.

busbyk: Although what's really weird is that no matter how high I set the depth I can't get all of the data on a page reference to populate for example. And there's no maxDepth set on the link field so I don't really understand why setting depth to some high number doesn't populate all of the data on a page reference. Maybe the payload local API has a default max depth or something? 

Very weird. So I've just set these both to 99.

src/components/Header/DesktopNav.client.tsx
busbyk: Set to use focus-visible instead of focus 👍

src/components/Header/utils.ts
busbyk: I just didn't update these when I pulled that in for the specific use case. I've updated these to use `tiny-invariant` instead. I think it looks nicer / is easier to read. 

I think the only time I might not use it is when a one-liner would be hard to read or not effectively check for something.

stevekuznetsov (approved):
:shipit:

stevekuznetsov:
Filed https://github.com/NWACus/web/issues/187 as a follow-up

busbyk:
@stevekuznetsov @rchlfryn got the 👍 from Chris so this is ready pending your review @rchlfryn

src/components/Header/Header.tsx
rchlfryn: I dont think it's worth changing

rchlfryn (approved):
LGTM

PR #168, opened 2025-05-02
seed: add an incremental seed using hashes

stevekuznetsov:
seed: add an incremental seed using hashes

We can store a hidden field on our collections that contains the hashed
inputs used to create the object, so a re-seed only needs to update the
(few) items that really changed. The current iteration does not remove
collection items not specified in the newest seed, but that can be
easily added if we find it to be necessary. On my machine, this brings
incremental seed times down to <5s.

---


README.md
busbyk, line 1: Can you add something here for reseed.sh too?

I also had the thought that it might be nice to add package.json scripts for running boostrap.sh and reseed.sh simply for convenience and potentially for discoverability. I find it nice to have all my scripts defined in package.json.

src/collections/Pages/hooks/revalidatePage.ts
busbyk, line 1: Are these changes for the incremental reseed?

src/endpoints/seed/navigation.ts
```diff
@@ -46,7 +47,7 @@ export const navigationSeed = (
 
   return {
     _status: 'published',
-    tenant,
+    tenant: tenant.id,
```
busbyk, line 50: I see a lot of this change (full doc -> just the doc id) in this PR. What is this for? You can pass either the id of a doc or a partial of the doc in the payload local API so I don't think there's a functional change here right? Just reducing the size of data passed around?

busbyk (commented):
Cool solution!

Def adds some complexity but seems fine. 

Can you change the target branch to skuznetz/fix-wal since this has a lot of changes from that PR?

src/collections/Pages/hooks/revalidatePage.ts
stevekuznetsov: No, good catch, it's separate. Can break out a commit or PR as you prefer. The logic was just incorrect as-is (and will need to be updated for nav as well)

src/endpoints/seed/navigation.ts
stevekuznetsov: Sorry, I thought I had documented that. We have a couple of collections like User with fields that populate post-creation, like `saveToJwt: true` on fields, or references, etc. What we want to be able to do with the hash is capture the _inputs_ that we're using to create the object. When we were putting in the full `tenant`, we might put in some minimal fields the first time seeding, but on re-seed when we look up the tenant we will get more fields, computed fields, etc. Without adding logic to prune out every type of computed field, using `tenant.id` instead of `tenant` was the simplest way I could figure to keep the set of input parameters constant so the hash would not change. This has the downside of being fragile as deleting all the tenants and starting over in some other order would change which ID maps to which tenant, but this script will be very useful to us even with that caveat. If you're doing big changes like that you can always blow away the DB and seed from scratch.

busbyk: Ah sweet, makes sense! Thanks for the explanation.

.github/workflows/ci.yaml
```diff
@@ -48,8 +48,16 @@ jobs:
       - name: 📦 Opt out of image optimization
         run: "sed -i 's/unoptimized: false/unoptimized: true/' next.config.js"
         shell: bash
+      - name: 🧹 Run seed
```
busbyk, line 51: Nice

busbyk (approved):
Looks good!

PR #170, opened 2025-05-07
Preview environment aliases

busbyk:
This creates aliases with a wildcard domain to enable our middleware to work as expected with tenant subdomains (i.e. `{tenant slug}.{branch name}.preview.avy-fx.org`). It also sets the SERVER_DOMAIN and NEXT_PUBLIC_SERVER_URL env vars with the correct alias domain and alias url for our getURL functions. 

Leaves a comment on the PR with the alias domain. 

Without the wildcard alias, using subdomains doesn't work and it would be a pain to set this up manually for each preview deployment.


.github/workflows/preview.yaml
```diff
@@ -50,12 +50,30 @@ jobs:
         shell: bash
       - name: Install Vercel CLI
         run: npm install --global vercel@latest
+      - name: Generate preview domain names
+        id: domains
+        run: |
+          REF="${{ github.head_ref }}"
+          CLEAN_REF="${REF//[^a-z0-9\-]/x}"
```
stevekuznetsov: Shove this into the step output for the `database` step so we have one source of truth please

.github/workflows/preview.yaml
```diff
@@ -90,4 +112,31 @@ jobs:
...
+          echo "Deployment URL: ${DEPLOYMENT_URL}"
+      - name: Create aliases for the deployment
+        id: aliases
+        run: |
+          echo "Setting preview alias: ${{ steps.domains.outputs.preview_alias }}"
+          vercel alias --token=${{ secrets.VERCEL_TOKEN }} set "${{ steps.deployment.outputs.url }}" "${{ steps.domains.outputs.preview_alias }}"
+
+          echo "Setting wildcard preview alias: ${{ steps.domains.outputs.wildcard_preview_alias }}"
+          vercel alias --token=${{ secrets.VERCEL_TOKEN }} set "${{ steps.deployment.outputs.url }}" "${{ steps.domains.outputs.wildcard_preview_alias }}"
+      - name: Comment on PR with deployment URL
```
stevekuznetsov: Will this clean up previous comments?

busbyk: It will now! Updated

src/utilities/getURL.ts
```diff
@@ -16,9 +16,5 @@ export const getClientSideURL = () => {
     return `${protocol}//${domain}${port ? `:${port}` : ''}`
   }
 
-  if (process.env.VERCEL_PROJECT_PRODUCTION_URL) {
-    return `https://${process.env.VERCEL_PROJECT_PRODUCTION_URL}`
-  }
-
```
busbyk, line 22: This always points to the production url (in our case avy-fx.org) which would cause issues in non-prod environments.

.github/workflows/preview.yaml
```diff
@@ -53,11 +56,8 @@ jobs:
       - name: Generate preview domain names
         id: domains
         run: |
-          REF="${{ github.head_ref }}"
-          CLEAN_REF="${REF//[^a-z0-9\-]/x}"
-
-          PREVIEW_ALIAS="${CLEAN_REF}.preview.avy-fx.org"
-          WILDCARD_PREVIEW_ALIAS="*.${CLEAN_REF}.preview.avy-fx.org"
+          PREVIEW_ALIAS="${{ steps.database.outputs.name }}.preview.avy-fx.org"
```
stevekuznetsov:
```suggestion
          PREVIEW_ALIAS="${{ steps.database.outputs.ref }}.preview.avy-fx.org"
```

Since the `name` is the `payloadcms-preview-${ref}`

stevekuznetsov:
Amazing :sunglasses:

PR #182, opened 2025-05-09
Set sans font to lato

### LINKED ISSUE #131: Change frontend font to Lato
This should set the frontend font to Lato instead of Geist Mono/Geist Sans that was configured by the Payload website template. 

Slack conversation for context: https://nwactechcommittee.slack.com/archives/C08JD81P6GZ/p1744922083021709?thread_ts=1744829204.444219&cid=C08JD81P6GZ

busbyk:
Resolves #131

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "clsx": "^2.1.1",
     "cross-env": "^7.0.3",
     "date-fns": "^4.1.0",
-    "geist": "^1.3.1",
     "graphql": "^16.10.0",
     "json-stable-stringify": "^1.3.0",
     "libsql": "^0.5.6",
--- a/src/app/(frontend)/layout.tsx
+++ b/src/app/(frontend)/layout.tsx
@@ -1,8 +1,6 @@
 import type { Metadata } from 'next'
 
 import { cn } from '@/utilities/ui'
-import { GeistMono } from 'geist/font/mono'
-import { GeistSans } from 'geist/font/sans'
 import React from 'react'
 import './globals.css'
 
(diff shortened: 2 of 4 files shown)


src/app/(frontend)/layout.tsx
```diff
@@ -13,6 +11,14 @@ import { mergeOpenGraph } from '@/utilities/mergeOpenGraph'
 import { getServerSideURL } from '@/utilities/getURL'
 import { draftMode } from 'next/headers'
 
+import { Lato } from 'next/font/google'
+
+const lato = Lato({
+  subsets: ['latin'],
+  weight: ['100', '300', '400', '700', '900'],
```
busbyk, line 18: Lato is not a variable font so these are all of the available weights. See: https://fonts.google.com/specimen/Lato

src/app/(frontend)/layout.tsx
```diff
@@ -21,7 +27,7 @@ export default async function RootLayout({ children }: Args) {
   const { isEnabled } = await draftMode()
 
   return (
-    <html className={cn(GeistSans.variable, GeistMono.variable)} lang="en" suppressHydrationWarning>
+    <html className={cn(lato.variable)} lang="en" suppressHydrationWarning>
```
rchlfryn, line 30: Maybe add `font-sans` for good measure.

package.json
```diff
@@ -57,7 +57,6 @@
     "clsx": "^2.1.1",
     "cross-env": "^7.0.3",
     "date-fns": "^4.1.0",
-    "geist": "^1.3.1",
     "graphql": "^16.10.0",
```
rchlfryn, line 60: Well this is a first. I have never seen a font as a package

rchlfryn (approved):
LG - pending one fix

package.json
busbyk: I know - I was surprised too. And it's funny they (payload) installed it as a separate package because it's available in next/font.

busbyk: Pretty nice though.

src/app/(frontend)/layout.tsx
busbyk: Tailwind's [preflight adds that to the html tag](https://v3.tailwindcss.com/docs/font-family#customizing-the-default-font) so it would be specified twice. Not that that would cause any problems I can foresee but I think I'd rather let the variable do it's thing.

stevekuznetsov:
:shipit:

PR #183, opened 2025-05-09
SAC theme

### LINKED ISSUE #171: Add SAC theme
Update globals.css to include a theme for SAC based on colors from Chris: https://www.figma.com/design/FchXa0tSmVCREq1ZMuTJK0/Sierra-Brand-Colors?node-id=8-10&t=aV4ns5jrmvFKDH5n-1

Should we hide the theme-preview page in production? I think yes.

busbyk:
Resolves #171


src/app/(frontend)/globals.css
```diff
@@ -80,14 +80,9 @@
   }
 
   .nwac {
-    --background: 0 0% 100%;
-    --foreground: 240 10% 3.9%;
```
rchlfryn, line 84: These colors are used in come of the components. Just a reminder to  update the components since I dont think you want to leave these in

rchlfryn (changes requested):

src/app/(frontend)/globals.css
busbyk: I removed them because they are the same as the `:root`. So these will resolve to the same values as they were before!

I'm only including classes here that are different for this theme.

stevekuznetsov (approved):

rchlfryn (approved):

PR #185, opened 2025-05-09
Blocks: membership

### LINKED ISSUE #108: Blocks: Membership
Create block for AC staff to edit membership information
NOTE: show already member blurb

Image
Image

rchlfryn:
### Description
Adds a dynamic membership block with the ability to add an optional callout

TODO: Decide button colors - in figma it is a brand color but we will need to change this

### Issue
Resolves #108 

### Screenshots
| with callout | without callout |
|---|---|
| Screenshot 2025-05-09 at 14 43 57 | Screenshot 2025-05-09 at 14 45 22 |


src/blocks/Button/config.ts
```diff
@@ -0,0 +1,21 @@
+import { button } from '@/fields/button'
+import type { Block } from 'payload'
+
+export const ButtonBlock: Block = {
```
busbyk: Should this be plural since its an array and could be multiple buttons?

src/blocks/Button/config.ts
```diff
@@ -0,0 +1,21 @@
...
+    {
+      name: 'buttons',
+      type: 'array',
+      label: false,
+      labels: {
+        singular: 'Button',
+        plural: 'Buttons',
+      },
+      fields: [button(['default', 'secondary'])],
+      maxRows: 2,
```
busbyk: What about removing maxRows and adding flex-wrap to the ButtonBlock Component's container div? I know there's only two buttons in the membership design but if this is going to be a generic component editors may want to add more than two buttons. 

Unless we're intentionally trying to restrict users based on a design mentality that more than two buttons is too much?

src/blocks/Membership/Component.tsx
```diff
@@ -0,0 +1,22 @@
...
+import RichText from '@/components/RichText'
+
+export const MembershipBlock = ({ callout, enableCallout, richText }: MembershipBlockProps) => {
+  return (
+    <div className="container">
+      {enableCallout ? (
+        <div className="flex flex-col md:flex-row items-center gap-4">
+          {richText && <RichText className="mb-0" data={richText} enableGutter={false} />}
+          {callout && (
+            <div className="p-6 bg-slate-200 rounded-sm">
```
busbyk: What about bg-brand-200 instead of bg-slate-200? I think that would fit with our intention for those classes in our themes. The NWAC theme specifically uses slate for the brand shades so it will be an exact match but I think it would work in other themes as well.

src/components/RichText/index.tsx
```diff
@@ -8,21 +8,21 @@ import { SerializedEditorState } from '@payloadcms/richtext-lexical/lexical'
 import {
   JSXConvertersFunction,
   LinkJSXConverter,
-  RichText as RichTextWithoutBlocks,
+  RichText as RichTextWithConverters,
```
busbyk: I think this rename is just to avoid conflicting with the exported component name in this file. What about `RichTextPayload` or `RichTextLexical`?

Since we add the converters to this import on line 65 I think this name is a little confusing.

tailwind.config.mjs
```diff
@@ -153,6 +153,11 @@ export default {
                 fontWeight: 'normal',
                 marginBottom: '0.25em',
               },
+              p: {
+                marginTop: '1em',
+                marginBottom: '1em',
+                lineHeight: 1.2,
+              },
```
busbyk, line 160: 🙌

busbyk (commented):
The mapping of button field variant -> shadcn buttom component variant -> semantic classname is currently correct. Although I wonder if renaming "default" to "primary" would be more clear. Not sure on that though. 

Based on the conversation in #108: should this be an even more generic component? If there's nothing membership-specific to this, we could call this like "ContentWithCallout" or something along those lines. If we did that maybe it would make sense to always have the callout too instead of enabling it with the checkbox since if a content editor didn't want the callout they could just use the Content block with a single column. 

If we did this maybe we'd want to:
1. Enable all variants of the `buttonBlock` in ContentWithCallout
2. and/or enable the `buttonBlock` in the Content component too -- thinking that content editors may want to use buttons for links in other content.

src/blocks/Button/config.ts
rchlfryn: I think we can remove the `maxRows`. There is a flex already on the components so I was more worries about restricting users to not add 5 buttons, but I don't think that will happen.

src/blocks/Membership/Component.tsx
rchlfryn: I wrote and stashed the code before the themeing PR. I will update to `bg-brand-200`

rchlfryn:
> The mapping of button field variant -> shadcn buttom component variant -> semantic classname is currently correct. Although I wonder if renaming "default" to "primary" would be more clear. Not sure on that though.

I am not sure on this either. Maybe just change the label and leave a comment? What is the drive to use primary?

> Based on the conversation in #108: should this be an even more generic component? If there's nothing membership-specific to this, we could call this like "ContentWithCallout" or something along those lines. If we did that maybe it would make sense to always have the callout too instead of enabling it with the checkbox since if a content editor didn't want the callout they could just use the Content block with a single column.
> 
> If we did this maybe we'd want to:
> 
> 1. Enable all variants of the `buttonBlock` in ContentWithCallout
> 2. and/or enable the `buttonBlock` in the Content component too -- thinking that content editors may want to use buttons for links in other content.

1. I love the name change. And thats makes sense to me to add all variants
2. I like keeping the two components separate, so I will add the `buttonsBlock` to Content and also alter `ContentWithCallout` to not have a check box.
I made `richText` into its own field component so we wont have to update in multiple places

src/blocks/Button/config.ts
rchlfryn: Updated it to be plural.

src/components/RichText/index.tsx
rchlfryn: Woah - I don't know what I was thinking here. Updated to `RichTextLexical`

src/fields/richText.ts
busbyk, line 1: We have richText type fields in a few places.
image

If we're going to create a re-usable field I think we should make it a function that accepts some parameters like name and a features list. We could use this as a way to have a default list of features but then be able to override or add to the features list. 

We could also have re-usable richText fields for more specific uses like this one that has the Buttons block that maybe we wouldn't include in the default richText field?

Thoughts?

src/blocks/ContentWithCallout/config.ts
```diff
@@ -0,0 +1,35 @@
...
+            ...rootFeatures,
+            FixedToolbarFeature(),
+            BlocksFeature({
+              blocks: [ButtonsBlock],
+            }),
+            HeadingFeature({ enabledHeadingSizes: ['h2', 'h3', 'h4'] }),
+          ]
+        },
+      }),
+    },
```
busbyk, line 48: Based on my other comment about re-usable richText fields config - when to use a re-usable richText field function instead of just writing this out? 

I guess in this case this has fewer features than `@/fields/richText.ts` so it makes sense to just write it out? 🤷

src/blocks/RenderBlocks.tsx
```diff
@@ -42,8 +43,8 @@ export const RenderBlock = ({ block, payload }: { block: Page['layout'][0]; payl
   switch (blockType) {
     case 'biography':
       return <BiographyBlock {...block} payload={payload} />
-    case 'cta':
-      return <CallToActionBlock {...block} />
+    case 'buttonsBlock':
+      return <ButtonsBlock {...block} />
```
busbyk: The rest of these blocks are "section-level" blocks but I don't think `src/blocks/Buttons/Component.tsx` would render well as a section-level block right? It doesn't use the `container` class or have any padding. 

If we want to allow this as a section-level block maybe add some props that will add the same classes we're using for other section-level block containers like `container py-16`?

src/collections/Pages/index.ts
```diff
@@ -97,8 +100,9 @@ export const Pages: CollectionConfig<'pages'> = {
               type: 'blocks',
               blocks: [
                 BiographyBlock,
-                CallToAction,
+                ButtonsBlock,
```
busbyk: Same thought as my comment in `src/blocks/RenderBlocks.tsx`

busbyk:
> > The mapping of button field variant -> shadcn buttom component variant -> semantic classname is currently correct. Although I wonder if renaming "default" to "primary" would be more clear. Not sure on that though.
> 
> I am not sure on this either. Maybe just change the label and leave a comment? What is the drive to use primary?

I think just to stay consistent with the theme classnames. But let's just leave it as "default".

src/fields/richText.ts
stevekuznetsov: The list of features is the only thing that this field currently has an opinion on, without that it's ~4 lines of code:

```ts
export const richText: Field = {
  name: 'richText',
  type: 'richText',
  editor: lexicalEditor({  }),
  label: false,
}
```

In cases like this, I prefer to follow the paradigm of "a little copying is better than a lot of dependency" - I would personally prefer to just not have the helper, since it does very little for us if we're not standardizing around a common set of features - we wouldn't be DRY-ing out anything of significant value. 

Conversely, if we _are_ standardizing around a specific set of features, and not making it configurable, I would expect this to have a semantically meaningful name (what's special about _this_ rich text field? why is it different than another? when would I choose to use it?)

src/collections/Pages/index.ts
rchlfryn: Great catch!

src/blocks/RenderBlocks.tsx
rchlfryn: great catch! I will take it out of section blocks for now

src/fields/richText.ts
rchlfryn: I was just trying to make a `content richText` field but I will revert this change.
I think @stevekuznetsov makes a great point and copy and pasting a few lines is fine for now. If we want to make it a reusable component, that should be a separate PR.

busbyk: All sounds good to me! Agreed.

busbyk (approved):
🚀

Issue #197, opened 2025-05-15
Can we update the widget text styles to match the new multi-tenant website?


chrisharrisonjones:

busbyk:
@chrisharrisonjones What do you want customized? 

We can hook into the css variables they use for some things and I can target specific classnames for others. 

#198 at least addresses setting the font to Lato (and changing the background color of the widgets but that's for #192 )

chrisharrisonjones:
I think these are my main goals. Matching header styles and digging into formatting would be great, but I worry that we'll get into too many variables for placements we have no control over?

busbyk:
Potentially. The css selectors are pretty good so let me know if/how you'd like to modify certain styles and I can let you know if they're easily targeted. I'd be happy leaving it at just using Lato and our background color though since the more custom css we write the more potential we have for those styles to "break" if the widget css selectors change. 

But I also think that we can inform Chris of our dependency on certain things.

chrisharrisonjones:
I think background color and font are a good first step. Let's see what it looks like integrated and if there's little tweaks we can do to improve. Thanks!

busbyk:
Resolved by #198

PR #198, opened 2025-05-16
NAC Widgets Enhancements

### LINKED ISSUE #161: NAC widget enhancements/customizations
This is a parent issue for all things related to NAC widgets.

Some things we can address solely on our end but this can also house tasks for updating things in the widgets' codebase.

### LINKED ISSUE #162: Widgets do not re-render on client side navigation
Demo: https://www.loom.com/share/71e5a2d4ce7247589799287fef9b19ca?sid=0a7e8703-c6aa-452e-9072-442f7979b9fd

This happens because during client side routing Next.js does not re-render the entire page, only sections of the page. The NAC loader script is not re-initializing the widget on client side navigation because the script was already run and it is not re-added to the page. Potentially due to using the next/script package but more investigation is needed. 

Generally, you will use a useEffect that adds the script tag on mount and removes the script tag on unmount.

### LINKED ISSUE #163: NAC widgets cause layout shift
Demo: https://www.loom.com/share/4151781b303141e398618552ef70ffca?sid=4c346eb4-2188-484b-a219-3f7460d9f715

Notice how the footer is pushed down once the widget is rendered. 

We can simply set a min-height on a container div for widgets as a simple solution. 

Might also be a good idea to set a min height in our `layout.tsx` to get our main content (header + main + footer) to be min-height: 100vh (i.e. full height of the viewport). Not related to the widgets loading but a good practice to avoid the weirdness with the footer appearing in the middle of the screen in the above demo.

### LINKED ISSUE #192: Widget background color does not match AC webpage

busbyk:
This addresses many of the bugs on #161 but it also skips the nac-widget-loader script and loads each widget type directly. This was partially to resolve #162 because the nac-widget-loader script doesn't do any cache busting on the widget scripts which was causing the browser to not re-run those initialization scripts on client side navigation. 

Resolves #161 with css to fix our font-family being overridden (loading the widget styles first to take the lowest precedence) and adding a reset for the widget container so our styles don't affect theirs. 

Resolves #163 by updating our app layout to have a minimum height of 100 viewport height and setting min height on forecast and observations widgets. 

Partially addresses #197 and resolves #192 by setting the background color to our --background variable and using our font. We could do more here by changing the value of the css variables used by the widgets to match things like button styling. 

Partially addresses #4 by hijacking the forecast links in the forecast widget's #all "zone" to use next client side navigation. The zone hash is still necessary in the url resulting in urls like: `/forecasts/avalanche/olympics#/olympics/` but the /forecasts/avalanche/[zone] route makes sure that hash exists if landing directly on that page from an external link. We still need to fetch the forecast zones and add them to the navigation. 

## Changes needed from this PR
1. We need to update the GitHub Actions secrets to use NEXT_PUBLIC_GOOGLE_MAPS_API_KEY instead of GOOGLE_MAPS_API_KEY. I've added this to Vercel for _just_ this branch so far. Will update prior to merge. 

Note: The iframe approach proved to have too many issues with the way the widgets are loaded. They expect to be in the parent document. This was particularly problematic for links when rendering the widgets in an iframe.


.env.example
```diff
@@ -16,8 +16,14 @@ NEXT_PUBLIC_SERVER_URL=http://localhost:3000
...
 # The API key for the Google Maps SDK we pass to the National Avalanche Center widgets.
-GOOGLE_MAPS_API_KEY=18514e92a54f6afc9a7fe83957be4fc3f1a89328
+NEXT_PUBLIC_GOOGLE_MAPS_API_KEY=18514e92a54f6afc9a7fe83957be4fc3f1a89328
 
 # Used to prefer redirecting to a tenant's custom domain or subdomain
 # you'll likely never want to set this to "true" in your local env
 REDIRECT_TO_CUSTOM_DOMAIN=false
+
+# The S3 bucket/CloudFront url for the NAC widgets
+NEXT_PUBLIC_NAC_WIDGETS_BUCKET=https://du6amfiq9m9h7.cloudfront.net
```
stevekuznetsov: This should be stored as a collection in the database - we need to allow the value to vary on its own timeline/lifecycle, and without our involvement. When it changes, we need to revalidate just the pages that are affected, and not require that a change to the version of the widgets go through a whole Vercel redeploy to land in users' hands. Landing in a collection also gives us a very clean way to let Chris self-serve manage this value if the automated process we develop has a failure. If you aren't ready to take this on during this PR, let's pull all of the loader stuff out and address it later. We're not in a rush to land this, have no users that could be impacted 
and therefore have the breathing room to invest in a more durable solution here.

.github/workflows/ci.yaml
```diff
@@ -54,15 +54,15 @@ jobs:
         env:
           DATABASE_URI: 'file:./dev.db'
           PAYLOAD_SECRET: ${{ secrets.PAYLOAD_SECRET }}
-          GOOGLE_MAPS_API_KEY: ${{ secrets.GOOGLE_MAPS_API_KEY }}
+          NEXT_PUBLIC_GOOGLE_MAPS_API_KEY: ${{ secrets.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY }}
```
stevekuznetsov, line 57: We need to use `NEXT_PUBLIC` since this value is used on the front-end, right?

src/app/(frontend)/[center]/nac-widgets.css
```diff
@@ -0,0 +1,36 @@
...
+#nac-app {
+  min-height: initial !important;
+}
+
+#nac-app #nac-weather-app {
+  min-height: initial !important;
+}
+
+#nac-app #nac-weather-map {
+  height: calc(100vh - 440px) !important;
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/nac-widgets.css]: What's the meaning of 440px?

src/app/(frontend)/[center]/observations/page.tsx
```diff
@@ -34,12 +34,19 @@ type PathArgs = {
...
-      <div className="container mb-16">
-        <div className="prose dark:prose-invert max-w-none" id="nac-widget-container">
-          <h1>Avalanche Observations For {center}</h1>
-          <NACWidget center={center} widget={'observations'} id="nac-widget-container" />
+    <div className="py-12">
+      <div className="container flex flex-col gap-4">
+        <div className="prose dark:prose-invert max-w-none">
+          <h1>Observations</h1>
         </div>
+        <p className="text-muted-foreground text-sm">
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/observations/page.tsx]: Do all centers use the same copy here, or should this be a RichText block in the settings/nav collection so it can vary per center? If it's the same, maybe we just store the forecaster contact email per center in settings and pull that in?

src/app/(frontend)/[center]/weather/stations/map/page.tsx
```diff
@@ -34,12 +34,14 @@ type PathArgs = {
...
-    <div className="pt-24 pb-24">
-      <div className="container mb-16">
-        <div className="prose dark:prose-invert max-w-none" id="nac-widget-container">
-          <h1>Weather Station Map For {center}</h1>
-          <NACWidget center={center} widget={'stations'} id="nac-widget-container" />
+    <div className="py-12">
+      <div className="container flex flex-col gap-4">
+        <div className="prose dark:prose-invert max-w-none">
+          <h1>
+            <span className="uppercase">{center}</span> Weather Station Map
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/weather/stations/map/page.tsx:45]: Maybe a question for Chris, I originally had put in these blurbs just to make it obvious to me during testing that routing was working and content changed. Since the only thing on the page is a map of weather stations, and we have the current nav highlighted when you're on the page, do we want to invest some valuable top-level screen real estate for this header or could we omit it?

src/components/NACWidget/index.tsx
```diff
@@ -1,24 +1,202 @@
...
+    containerRef.current.innerHTML = ''
+
+    const widgetId = `nac-widget-${widget}-${instanceId}`
+
+    const appDiv = document.createElement('div')
+    appDiv.id = widgetId
+    containerRef.current.appendChild(appDiv)
+
+    const awsBucket =
+      process.env.NEXT_PUBLIC_NAC_WIDGETS_BUCKET || 'https://du6amfiq9m9h7.cloudfront.net'
```
stevekuznetsov: Since you are doing this client-side, will our SSR know the script tag that needs to be fetched, or will we incur the time cost of loading+parsing client-side JS before the script begins? Could we server-side render a bunch of `<Script>`?

src/components/NACWidget/index.tsx
```diff
@@ -1,24 +1,202 @@
...
+      centerId: center.toUpperCase(),
+      devMode: false,
+      mountId: `#${widgetId}`,
+      baseUrl: baseUrl,
+    }
+
+    switch (widget) {
+      case 'map':
+        window.mapWidgetData = widgetData
+        Promise.all([
```
stevekuznetsov: nit: might be more concise and DRY to do something like

```ts
const loadersByWidget: Record<string, { script: string, loadData: (widgetData): void> = {
  'map': { script: 'danger-map', load: (widgetData) => { window.mapWidgetData = widgetData }},
  'forecast': { script: 'forecasts', load: (widgetData) => { window.forecastWidgetData = widgetData }},
  'warning': { script: 'warnings', load: (widgetData) => { window.warningWidgetData = widgetData }},
  'stations': { script: 'stations', load: (widgetData) => { window.stationWidgetData = widgetData }},
  'observations': { script: 'observations', load: (widgetData) => { window.obsWidgetData = widgetData }},
}

const { script, load } = loadersByWidget[widget]
load(widgetData)
Promise.all([
  loadJS(`${scriptUrl}${script}.js?t=${Date.now()}`),
  loadCSS(`${scriptUrl}${script}.css?t=${Date.now()}`),
]).catch((err) => console.error('Failed to load ${widget} widget:', err))
```

.github/workflows/ci.yaml
busbyk: Yea

src/app/(frontend)/[center]/nac-widgets.css
Comment (busbyk) [reply src/app/(frontend)/[center]/nac-widgets.css]: Ah meant to vary this on mobile using a media query. Will add that in. 440px is the height + padding of elements that are rendered above the map so that the map does not extend "below the fold".

src/app/(frontend)/[center]/observations/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/observations/page.tsx]: I wasn't sure this was even required but it was displayed on NWAC's observations page. 

This should definitely come from a collection. I think the likelihood that the copy needs to vary between center's is high enough that using a richtext field makes sense.

src/app/(frontend)/[center]/weather/stations/map/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/weather/stations/map/page.tsx:45]: I only included these on the weather station map and observations based on the Figma designs. 

A side note: active styling was taken out of the nav per discussion with Chris. Breadcrumbs provides that same functionality.

src/components/NACWidget/index.tsx
busbyk: It's the same number of lines of code after prettier reformats 😅

src/components/NACWidget/index.tsx
busbyk: This is _all_ client-side. 

We incur the cost of loading+parsing client side JS before the script begins. We have to do this because both the NAC widget loading scripts and our client component need to perform DOM manipulation which can't be done on the server. 

And we can't even let the browser cache these scripts because of the way they are loaded. Each widget script does not re-mount the app after a client side navigation which destroys the mount point in the DOM. We need to have the script re-run. 

Just re-adding the script to the DOM doesn't re-execute it because the browser realizes it's the same script, realizes it's cached, and doesn't re-execute it. That's why I added the `Date.now()` calls to the script loading so they don't get cached and they _will_ get re-executed to re-mount the widgets. This is actually why I had to load the widget scripts directly, bypassing the nac-widget-loader because I needed to add the `Date.now()` to bypass caching. 

This would be a good case for making changes in https://github.com/NationalAvalancheCenter/afp-public-widgets to modify the loader scripts to watch the DOM and re-mount the app when a new mount point is added.

.env.example
busbyk: Do we really want that though? I'm thinking about a situation where a new version contains breaking changes for us. Would we want to review how the new version displays on our sites before upgrading? I know that that's valuable dev time but allowing Chris to auto-upgrade might break things (like the hijacking of the forecast zone buttons to work with our routing. 

> and not require that a change to the version of the widgets go through a whole Vercel redeploy to land in users' hands 

Hmm that is a good point though. Would be nice to just revalidate the pages that have widgets on them and not have to rebuild the entire app. 

> If you aren't ready to take this on during this PR, let's pull all of the loader stuff out and address it later. 

See my [other reply](https://github.com/NWACus/web/pull/198#discussion_r2093497559): I can't take out the loader stuff. 

 > We're not in a rush to land this, have no users that could be impacted and therefore have the breathing room to invest in a more durable solution here.
 
I understand that sentiment to some degree but I'd rather make progress on this and iterate than hold off for a perfect solution. This is also a bit of a mixed signal. @yuliadub was questioning why `navigation-frontend` was not in `main` for so long for example. Is your concern mainly about amassing tech debt?

I'll move to storing the version and cloudfront url in a [global](https://payloadcms.com/docs/configuration/globals).

src/app/(frontend)/[center]/observations/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/observations/page.tsx]: I'm removing this for now. I'd like to create a settings collection in another PR.

src/components/NACWidget/index.tsx
stevekuznetsov: Funny! It's a nit-pick, feel free to ignore. I prefer the map-of-data approach since you are not repeating any logic - e.g. can't accidentally call `loadCSS` on one of the many JS file, etc. You might make it more concise as far as LoC and duplicate codepaths actually with

```ts
const loadersByWidget: Record<string, { script: string, dataKey: string> = {
  'map': { script: 'danger-map', dataKey: 'mapWidgetData'},
  'forecast': { script: 'forecasts', dataKey: 'forecastWidgetData'},
  'warning': { script: 'warnings', dataKey: 'warningWidgetData '},
  'stations': { script: 'stations', dataKey: 'stationWidgetData '},
  'observations': { script: 'observations', dataKey: 'obsWidgetData'},
}

const { script, dataKey } = loadersByWidget[widget]
window[dataKey] = widgetData
Promise.all([
  loadJS(`${scriptUrl}${script}.js?t=${Date.now()}`),
  loadCSS(`${scriptUrl}${script}.css?t=${Date.now()}`),
]).catch((err) => console.error('Failed to load ${widget} widget:', err))
```

src/components/NACWidget/index.tsx
stevekuznetsov: Oh, wow - thanks for explaining that. Makes a lot of sense. If we make the changes to the scripts you mentioned, would we be able to statically determine up-front the list of `<Script>` tags server-side? If so, let's see if Chris is up for it - that would be the ultimate outcome here, where browser could fetch the script in parallel to the main page loading. We would then be able to upgrade this component from this client-side impl to a server-side one once they are ready.

.env.example
stevekuznetsov: > Would we want to review how the new version displays on our sites before upgrading?

I understand the concern, but ultimately I think we're in a pickle - we do not own the widgets and the NAC publishes them as a safety-critical service, so them being in control is paramount. That was originally why getting rid of the loader was seen as unacceptable. If the NAC finds a critical bug and needs to do a hot fix rollout, they need to be able to do that with zero interaction necessary on our end, since we aren't able to commit to the staffing necessary to be able to reply to Chris on short notice at any point during the winter, so I think our imperative is keeping them in the driver's seat for now. Your concern on auto-upgrade breaking things is 100% correct and valid, and I think for planned rollouts we can sync with the NAC to upgrade dev.avy-fx.org first to check like you say, but ultimately I don't think we can be in the critical path for prod on a hotfix.

>  Is your concern mainly about amassing tech debt?

I'm sorry, I misunderstood the different parts here, it was not clear that using the loader also impeded the reload functionality. Given that I thought you were only doing the loading on our end to make the site faster, it seemed like we could also just do nothing at all and be OK. Now that I understand why your `?date=` query is needed, I think what you've got here is perfectly reasonable and, as you say, can be something we make better in the future. Sorry!

yuliadub: a perfect solution doesnt exist 😅 but seriously, thanks for digging into this and getting all the pieces figured out. 

I also just want to be clear that it is completely okay to not merge things quickly, sorry if that came of the wrong way earlier this week.

The version thing is tricky. At first your approach made perfect sense since I actually didnt think that widgets change all that often - but then I took a look at how many changes (and for what reasons) widget versions have gone through over time/forecast season and that got me a little worried. Given how quickly certain things can change on the NAC side based on various center requests, it is actually possible that they make a backend change that can bring one of (or at times all) the widgets down for older versions. If we would need to redeploy to bring them back up (or coordinate with Chris to do this smoothly in any way) this poses a challenge as you two have already discussed, especially since we may not necessarily have an engineer on call who can help right away. 

Part of me wishes there was a way for the NAC widget to default to newest version  if one is not provided. Another way to handle this is to have the version be updated automatically when NAC makes a version change via a github workflow (although those do sometimes randomly(or not) fail and that scares me). So I guess this is my very long comment where I come to the conclusion that maybe keeping the loader would be my suggestion, because even though the performance is SO much better when it comes to loading, the extra cost of managing this version up to date stuff just seems more complex and error prone given our situation.

yuliadub: Steve and I were just chatting and he brought up a good point that during the meeting with NAC this week we talked about dropping the loader and that was the consensus. But in my case I dont think I fully ever internalized why the loader existed, i.e. it was actually managing versions for us. So this is my bad.

src/components/NACWidget/index.tsx
```diff
@@ -1,24 +1,202 @@
...
+    const widgetId = `nac-widget-${widget}-${instanceId}`
+
+    const appDiv = document.createElement('div')
+    appDiv.id = widgetId
+    containerRef.current.appendChild(appDiv)
+
+    const awsBucket =
+      process.env.NEXT_PUBLIC_NAC_WIDGETS_BUCKET || 'https://du6amfiq9m9h7.cloudfront.net'
+    const version = process.env.NEXT_PUBLIC_NAC_WIDGETS_VERSION || '20250313'
+    const scriptUrl = `${awsBucket}/public/v2/${version}/`
```
yuliadub: could we ask Chris to publish the latest version in v2/latest and we can always load that? just an idea

.env.example
yuliadub: had a random thought [here](https://github.com/NWACus/web/pull/198/files#r2093759866)

src/components/NACWidget/index.tsx
busbyk: Yea...it's not ideal. Definitely eliminates the benefits of SSG for some key pages which is a bummer. 

I think we would, yea. We wouldn't be able to do any DOM manipulation prior to client hydration (as far as I understand things) but we could prioritize loading the nac widget script required for a given page which, yea, would help with performance. 

> We would then be able to upgrade this component from this client-side impl to a server-side one once they are ready.

You mean integrate with their APIs directly and build our own UI right? That will be nice. I do have questions about how we'll do revalidation and ISR once that's the case with forecasts expiring. Some potential approaches:
1. Don't use SSG - dynamically server side render all avalanche.org API powered data and use SWR based on the expiration dates of forecasts. 
2. Poll the avalanche.org APIs to determine which pages need to be revalidated.
3. Set up a webhooks system to have avalanche.org send us events when data needs to be revalidated. 

That's a problem for the future but will be a fun problem to figure out!

.env.example
busbyk: This is great discussion - thanks for this. I appreciate the background and the thought process explanation!

> ...ultimately I don't think we can be in the critical path for prod on a hotfix.

The reasoning here makes sense. Giving Chris the ability to go in and update the version values himself makes total sense if 
> we aren't able to commit to the staffing necessary to be able to reply to Chris on short notice at any point during the winter

Using a global to manage the version number also has the benefit of not needing to rebuild the entire app on an env var change but just revalidating the pages that render widgets. So I'm for it. 

> I'm sorry, I misunderstood the different parts here, it was not clear that using the loader also impeded the reload functionality. Given that I thought you were only doing the loading on our end to make the site faster, it seemed like we could also just do nothing at all and be OK. Now that I understand why your ?date= query is needed, I think what you've got here is perfectly reasonable and, as you say, can be something we make better in the future. Sorry!

I probably should have explained [this](https://github.com/NWACus/web/pull/198#discussion_r2093497559) in more detail in the description to begin with 😅 

> a perfect solution doesnt exist 😅 but seriously, thanks for digging into this and getting all the pieces figured out.
>
> I also just want to be clear that it is completely okay to not merge things quickly, sorry if that came of the wrong way earlier this week.

All good. I'm also of the opinion that I would rather get a working version in and iterate on it than hold off on it for a perfect solution. But there's a balance to be struck there and clearly it's good we're having this conversation!

> So I guess this is my very long comment where I come to the conclusion that maybe keeping the loader would be my suggestion, because even though the performance is SO much better when it comes to loading, the extra cost of managing this version up to date stuff just seems more complex and error prone given our situation.

Because of the way the Vue apps are mounted we need to load the widget scripts directly to enable re-rendering on client side navigation. It's not ideal but managing the widget version in a global in payload seems to be the best solution for us right now. We can put it on Chris to update this version number when new versions are pushed and educate him on the modifications that we rely on. 

Does that make sense?

We can then pursue improving the way the Vue apps are mounted so we can use cached scripts per https://github.com/NWACus/web/pull/198#discussion_r2093497559.

src/components/NACWidget/index.tsx
busbyk: That could be a nice solution too! It would be nice to have `latest` as a fallback and still have control over the specific version if we wanted to rollback to a previous version for some reason.

.env.example
yuliadub: It does make sense but I dont love it since in my experience extra steps like that can be forgotten or unexpectedly broken and I worry about it long term since we dont have a dedicated dev yet to come save the day if something weird happens. I am wondering if we could just have Chris edit the widget deployment to put the latest widget code into the same aws path anytime an update is made and then have the widget reload on the fly without anything needed to be done by him or us. That is what I was trying to convey in [this](https://github.com/NWACus/web/pull/198/files#r2093759866) comment.

src/components/NACWidget/index.tsx
busbyk: Sweet, I updated. Nice and concise!

src/app/(frontend)/[center]/nac-widgets.css
Comment (busbyk) [reply src/app/(frontend)/[center]/nac-widgets.css]: Updated to vary on mobile. This was in an effort to avoid layout shift and set the map to be the height of the main content area (i.e. full viewheight minus header and heading).

.env.example
busbyk: Pushed changes to use global.

src/endpoints/seed/index.ts
```diff
@@ -43,7 +43,20 @@ const collections: CollectionSlug[] = [
...
-const globals: GlobalSlug[] = ['footer']
+const globalsMap: Record<GlobalSlug, { requiredFields: any }> = {
+  footer: {
+    requiredFields: {
+      navItems: [],
+    },
+  },
+  nacWidgetsConfig: {
+    requiredFields: {
+      version: '20250313',
```
busbyk, line 54: This seems so silly 😅 but we can't set to an empty string because I set this field to be required (makes sense to IMO) and an empty string fails that validation.

src/endpoints/seed/index.ts
```diff
@@ -871,6 +882,13 @@ export const seed = async ({
         ],
       },
     }),
+    payload.updateGlobal({
+      slug: 'nacWidgetsConfig',
+      data: {
+        version: '20250313',
```
busbyk, line 892: This is another case for requesting Chris to use a `latest` tag. We're bound to forget to update this locally. 

Beyond that - do we think it makes sense to use an environment variable for this, if for nothing else, for our local environments?

.env.example
busbyk: > It does make sense but I dont love it since in my experience extra steps like that can be forgotten or unexpectedly broken and I worry about it long term since we dont have a dedicated dev yet to come save the day if something weird happens.

You mean having Chris log in to the AvyFx admin panel to update the version in the nacWidgetsConfig global?

> I am wondering if we could just have Chris edit the widget deployment to put the latest widget code into the same aws path anytime an update is made and then have the widget reload on the fly without anything needed to be done by him or us. That is what I was trying to convey in [this](https://github.com/NWACus/web/pull/198/files#r2093759866) comment.

I think this would be good. I would suggest we keep the ability to change the version manually too in case we need to for some reason but would normally keep it set to "latest".

yuliadub: > You mean having Chris log in to the AvyFx admin panel to update the version in the nacWidgetsConfig global?

yeah. Or even if Chris automates this step say with a github action, if that automated process fails and no one can remember how it is set since they havent touched it in a while or something like that. 

> I think this would be good. I would suggest we keep the ability to change the version manually too in case we need to for some reason but would normally keep it set to "latest".

great point! agreed

busbyk: Created a task to update some defaults to "latest" when that's available: https://github.com/NWACus/web/issues/206

stevekuznetsov (approved):

Issue #200, opened 2025-05-16
Add subscribe form to footer


rchlfryn:
Right now we are using [Form Builder Plugin](https://payloadcms.com/docs/plugins/form-builder) to control all of our forms. This seems like overkill to add a email field to the footer. 

TODO: We need to add a 'form' to the footer to collect emails to subscribe to the avalanche centers newsletter

rustynwac:
@rchlfryn will this allow AC's to plug in whatever email tool signup tool they use? ie. Mailchimp or Campaign Monitor? I know we need to allow them to use whatever tool they choose for marketing. Usually those tools have a form embed or something. I can chat through this pretty easily. I could see the argument for having more fields on email capture (first/last/zip).

busbyk:
This is another case for creating a generic embed block in Payload with some basic input sanitization. This would be useful for any third-party embed which I believe we'll need several of. A donation form embed would be another example.

PR #201, opened 2025-05-16
Footers for tenants

### LINKED ISSUE #102: Customizable footer
Create footer AC staff can edit to be on brand. Give them the ability to update...
* logo
* AC business info
* social
* copyright
* T&C and Privacy policy

Image

rchlfryn:
### Description
Creates a customizable footer for each tenant. This PR does not include the subscribe form (Issue created #200)

### Issue
Resolves #102 

### Screenshot
8052ac969eee6ae1452d8169db31f3b3


src/collections/Footer/config.ts
```diff
@@ -0,0 +1,160 @@
...
+          slug: '',
+          collection: 'pages',
+          tenant,
+          req,
+        })
+
+        return path
+      },
+    },
+  },
```
rchlfryn, line 53: Do we need the live preview? It just shows the home page. I think we can live without it.

src/Footer/Component.tsx
```diff
@@ -1,29 +1,72 @@
...
+  const { docs: data } = await payload.find({
+    collection: 'footer',
+    depth: 10,
+    where: {
+      'tenant.slug': {
+        equals: center,
+      },
+    },
+  })
+  const { address, email, hashtag, logo, name, phone, privacy, socialMedia, terms } = data[0]
```
stevekuznetsov: Let's check the length of `data` to make sure `[0]` doesn't error at runtime.

src/Footer/Component.tsx
```diff
@@ -1,29 +1,72 @@
...
+                {socialMedia.linkedin && <Icons.linkedin />}
+                {socialMedia.youtube && <Icons.youtube />}
+              </div>
+            )}
+            {hashtag && <p className="mb-4 text-secondary underline">{hashtag}</p>}
+          </div>
+        </div>
+      </div>
+      <div className="container text-center pb-8">
+        <p className="mb-2">All Content © 2017 – 2025 {name}</p>
```
stevekuznetsov: How did we come up with these dates, and who owns the copyright? Do we want the end date to be dynamic from `new Date()` so we don't need to update it?

src/collections/Footer/config.ts
stevekuznetsov: If it's hard to maintain, we can live without. If it's easy, I say why not! I think folks will appreciate having it everywhere.

src/Footer/Component.tsx
```diff
@@ -1,29 +1,72 @@
...
+        <div className="flex gap-x-2 justify-center">
+          {typeof terms === 'object' && (
```
stevekuznetsov: Could we use `tiny-invariant` on these like @busbyk did in the nav? If this is not `'object'` we just need to bump the `depth` on the query.

src/Footer/Component.tsx
```diff
@@ -1,29 +1,72 @@
...
+            <a href={`/${terms?.slug}`} className="underline">
```
stevekuznetsov: `terms` is a required reference to a page - why do we need the `?`? In what case would it be `null`?

src/collections/Footer/config.ts
```diff
@@ -0,0 +1,160 @@
...
+import { contentHashField } from '@/fields/contentHashField'
+import { tenantField } from '@/fields/tenantField'
+import { generatePreviewPath } from '@/utilities/generatePreviewPath'
+import type { CollectionConfig, TextFieldValidation } from 'payload'
+
+const validateHashtag: TextFieldValidation = (value: string | null | undefined): string | true => {
+  if (value === null || value === undefined) return true
+  else
+    return (
+      value?.match(/^#[A-Za-z0-9_](?:[A-Za-z0-9_]|(?:\.(?!\.))){0,28}[A-Za-z0-9_]$/)?.length ===
```
stevekuznetsov: Why `foo.match().length === 1` instead of `foo.test()`?

src/collections/Footer/config.ts
```diff
@@ -0,0 +1,160 @@
...
+        1 || `${value} is not a valid hashtag`
+    )
+}
+
+const validateTelephone: TextFieldValidation = (
+  value: string | null | undefined,
+): string | true => {
+  console.log(value?.match(/^(\+1\s?|1\s?)?(\(\d{3}\)|\d{3})[\s.-]?\d{3}[\s.-]?\d{4}$/))
```
stevekuznetsov: leftover?

src/collections/Footer/config.ts
```diff
@@ -0,0 +1,160 @@
...
+  if (value === null || value === undefined) return true
+  else
+    return (
+      value?.match(/^(\+1\s?|1\s?)?(\(\d{3}\)|\d{3})[\s.-]?\d{3}[\s.-]?\d{4}$/g)?.length === 1 ||
+      `${value} is not a valid hashtag`
```
stevekuznetsov:
```suggestion
      `${value} is not a valid telephone number`
```

src/collections/Footer/config.ts
```diff
@@ -0,0 +1,160 @@
...
+        })
+
+        return path
+      },
+    },
+  },
+  fields: [
+    tenantField({ unique: true }),
+    {
+      name: 'logo',
```
stevekuznetsov: Don't we already have a logo set in the theme collection?

src/collections/Footer/config.ts
```diff
@@ -0,0 +1,160 @@
...
+      type: 'upload',
+      relationTo: 'media',
+      // required: true,
+      filterOptions: {
+        mimeType: { contains: 'image' },
+      },
+    },
+    {
+      name: 'name',
```
stevekuznetsov, line 65: Some of these fields feel like good candidates for the tenant collection or settings collection instead of being specific to the footer. @busbyk I know you had some ideas kicking around for a settings collection - would you rather this PR start a proto-settings collection, or could we keep the footer as-is with the understanding that your future work will bring it all together?

src/endpoints/seed/index.ts
```diff
@@ -60,24 +59,6 @@ export const seed = async ({
 }): Promise<void> => {
   payload.logger.info('Seeding database...')
   if (!incremental) {
-    payload.logger.info(`— Clearing collections and globals...`)
```
stevekuznetsov: Why was this change made?

src/endpoints/seed/index.ts
```diff
@@ -399,6 +380,8 @@ export const seed = async ({
       .flat(),
   ])
 
+  payload.logger.info(`TYPEOF ${brandImages}`)
```
stevekuznetsov: Is this a leftover?

src/endpoints/seed/index.ts
```diff
@@ -838,40 +821,59 @@ export const seed = async ({
...
-            link: {
-              type: 'custom',
-              label: 'Payload',
-              newTab: true,
-              url: 'https://payloadcms.com/',
-            },
-          },
-        ],
+  payload.logger.info(`— Seeding footers...`)
+  const footerData: Record<Tenant['slug'], {}> = {
```
stevekuznetsov: Could the values of this record be typed with `RequiredDataFromCollectionSlug<'footer'>`?

src/endpoints/seed/index.ts
```diff
@@ -838,40 +821,59 @@ export const seed = async ({
...
+      socialMedia: {
+        instagram: 'https://www.instagram.com/nwacus',
+        facebook: 'https://www.facebook.com/NWACUS/',
+        twitter: 'https://x.com/nwacus',
+        linkedin: 'https://www.linkedin.com/company/nw-avalanche-center',
+        youtube: 'https://www.youtube.com/channel/UCXKN3Cu9rnnkukkiUUgjzFQ',
       },
-    }),
-  ])
+      contentHash: null,
```
stevekuznetsov: nit: you could omit these here

src/Footer/Component.tsx
rchlfryn: These were in the mocks which appears to be taken from the current website. And it's common to use a span to show the years of publication. I typically just use `new Date()` to show the current year, but this might be a Rusty question.

src/Footer/Component.tsx
rchlfryn: I have this as optional since I am not sure every AC has Terms & Conditions and a Privacy policy. I switched to have an initial null check rather than inline.

src/collections/Footer/config.ts
rchlfryn: No reason other than I've just always used `match`. Cleaned this area up a little

src/endpoints/seed/index.ts
rchlfryn: ~Thanks - I could not figure out the type for this one~
`Record<Tenant['slug'], RequiredDataFromCollectionSlug<'footer'>>` didn't work as the type.

rchlfryn:
```bash
src/endpoints/seed/index.ts(846,5): error TS2322: Type '{ address: string; email: string; socialMedia: {}; contentHash: null; }' is not assignable to type 'Omit<Footer, "createdAt" | "updatedAt" | "sizes" | "id"> & Partial<Pick<Footer, "createdAt" | "updatedAt" | "sizes" | "id">>'.
  Type '{ address: string; email: string; socialMedia: {}; contentHash: null; }' is missing the following properties from type 'Omit<Footer, "createdAt" | "updatedAt" | "sizes" | "id">': name, tenant
  ```

File: src/endpoints/seed/index.ts (hunk 11 above)
Comment (rchlfryn) [reply src/endpoints/seed/index.ts]: We don't have any globals anymore since we changed the footer to be a collection for each tenant. Is my understanding of the change incorrect?

File: src/collections/Footer/config.ts (hunk 9 above)
Comment (rchlfryn) [reply src/collections/Footer/config.ts]: Good call. Updated to `footerLogo`

File: src/endpoints/seed/index.ts (hunk 14 above)
Comment (rchlfryn) [reply src/endpoints/seed/index.ts]: I got errors from `upsert.ts` when I did. I think I need contentHash since it's a collection now?

Comment (stevekuznetsov) [reply src/endpoints/seed/index.ts]: Sorry, I meant that you could omit explicitly setting `contentHash: null` in this data, not that you could omit having the field on the collection itself. No biggie either way.

File: src/endpoints/seed/index.ts (hunk 13 above)
Comment (stevekuznetsov) [reply src/endpoints/seed/index.ts]: Ah, I see - unless you need to have the proto-data in its own variable for something later, this got worked around in the other calls to `upsert` by just inlining the data and mapping functions. No biggie either way, as it looks like all the fields being spread by `...footerData[tenant.slug]` are optional anyway.

File: src/Footer/Component.tsx (hunk 3 above)
Comment (stevekuznetsov) [reply src/Footer/Component.tsx]: Gotcha - yeah, let's check to see if we can be dynamic, that would be nice to have.

Comment (busbyk) [reply src/Footer/Component.tsx]: It would be nice to include this in a settings collection I think. Could default to _just_ current year if there isn't a start year set. I don't know how we would refer to that. Content copyright starting year?

File: src/collections/Footer/config.ts (hunk 10 above)
Comment (busbyk) [reply src/collections/Footer/config.ts:65]: Yea I'm really looking forward to getting a single settings collection created with tabs for organization. I'd like to discuss global collections as a larger topic (including which custom/structured pages will need their own global collection) among the team. 

So I think let's keep data in the footer global collection in this PR for now and they can get pulled into the larger settings collection when we move to that.

File: src/endpoints/seed/index.ts [hunk 15]
```diff
@@ -853,47 +849,56 @@ export const seed = async ({
...
-              label: 'Payload',
-              newTab: true,
-              url: 'https://payloadcms.com/',
-            },
-          },
-        ],
+  payload.logger.info(`— Seeding footers...`)
+  const footerData: Record<Tenant['slug'], {}> = {
+    nwac: {
+      address: '249 Main Ave. S, Suite 107-366\nNorth Bend, WA 98045\n(206) 909-0203',
```
stevekuznetsov: nit: should the phone number be in the other field?

src/endpoints/seed/index.ts
```diff
@@ -853,47 +849,56 @@ export const seed = async ({
...
+  }
+
+  const footer = (
+    tenant: Tenant,
+    brandImages: Record<Tenant['slug'], Record<string, Media>>,
+  ): RequiredDataFromCollectionSlug<'footer'> => {
+    return {
+      tenant: tenant.id,
+      footerLogo: brandImages[tenant.slug]['logo'].id,
+      name: tenant.name,
```
stevekuznetsov, line 891: Would we ever want the footer's name to be something other than the tenant's name? If not, we should imo remove the field and just look up the tenant's name for rendering.

rchlfryn: I am not sure - which is why I made this dynamic. I will ask Rusty and report back.

rchlfryn: It looks like we can just import the tenants name. I will update this

PR #202, opened 2025-05-16
Add ci job for generate:type and generate:importmap

rchlfryn:
Adding CI job for generate:type and generate:importmap since I accidentally forgot to run them and we should have a check for this.


rchlfryn:
I was able to confirm with a [removed commit ](https://github.com/NWACus/web/actions/runs/15076680553/job/42385620821?pr=202) that this works as expected

PR #204, opened 2025-05-17
Bug: Fixes load error on page or post without a slug

rchlfryn:
### Description
I kept randomly getting an error 
```bash
Error: A required parameter (slug) was not provided as a string received object in generateStaticParams for /[center]/[slug]
``` 

### Steps to recreate
Create a new `page`/`post` and navigate away.  
Visit and `page`/`post` with the `[slug]` pattern.
See the error above/ it wont compile

### Why this is an issue
When an admin user goes to create a `page`/`post` and navigates away without filling anthing in, the `page`/`post` will auto save both of these every 100ms (We chose these settings to optimize for live preview and can be found in the `page`/`post` config settings.
The error occurs when we are loading `pages` and `posts` with `payload.find()`. Although we set `draft: false`, according the the [payload website](https://payloadcms.com/docs/versions/drafts#reading-drafts-vs-published-documents) 
> **Important**: the draft argument on its own will not restrict documents with `_status: 'draft'` from being returned from the API. 

Yes - this means it will then return ghost pages/posts with no slug.

### The fix
For now, I added a check to when we mapping the `page`/`post` to see if a slug exists on any `[slug]` files.

### DIFF
--- a/src/app/(frontend)/[center]/[slug]/page.tsx
+++ b/src/app/(frontend)/[center]/[slug]/page.tsx
@@ -25,6 +25,11 @@ export async function generateStaticParams() {
       tenant: true,
       slug: true,
     },
+    where: {
+      _status: {
+        equals: 'published',
+      },
+    },
   })
 
   const params: PathArgs[] = []
@@ -33,7 +38,7 @@ export async function generateStaticParams() {
       payload.logger.error(`got number for page tenant: ${JSON.stringify(page.tenant)}`)
       continue
     }
-    if (page.tenant && page.slug) {
+    if (page.tenant) {
       params.push({ center: page.tenant.slug, slug: page.slug })
     }
   }
--- a/src/app/(frontend)/[center]/[slug]/page.tsx
+++ b/src/app/(frontend)/[center]/[slug]/page.tsx
@@ -18,7 +18,6 @@ export async function generateStaticParams() {
     collection: 'pages',
     draft: false, // does not remove posts with _status: 'draft'
     limit: 1000,
-    overrideAccess: true,
     pagination: false,
     depth: 2,
     select: {
--- a/src/app/(frontend)/[center]/[slug]/page.tsx
+++ b/src/app/(frontend)/[center]/[slug]/page.tsx
@@ -16,7 +16,6 @@ export async function generateStaticParams() {
   const payload = await getPayload({ config: configPromise })
   const pages = await payload.find({
     collection: 'pages',
-    draft: false, // does not remove posts with _status: 'draft'
     limit: 1000,
     pagination: false,
     depth: 2,
--- a/src/app/(frontend)/[center]/[slug]/page.tsx
+++ b/src/app/(frontend)/[center]/[slug]/page.tsx
@@ -16,7 +16,7 @@ export async function generateStaticParams() {
   const payload = await getPayload({ config: configPromise })
   const pages = await payload.find({
     collection: 'pages',
-    draft: false,
+    draft: false, // does not remove posts with _status: 'draft'
     limit: 1000,
     overrideAccess: true,
     pagination: false,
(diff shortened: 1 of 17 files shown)


src/app/(frontend)/[center]/[slug]/page.tsx
```diff
@@ -16,7 +16,7 @@ export async function generateStaticParams() {
   const payload = await getPayload({ config: configPromise })
   const pages = await payload.find({
     collection: 'pages',
-    draft: false,
+    draft: false, // does not remove posts with _status: 'draft'
```
Comment (busbyk) [inline src/app/(frontend)/[center]/[slug]/page.tsx]: Ah yea looks like this is expected behavior (unexpected to me though 😂): [draft state note in docs](https://payloadcms.com/docs/versions/drafts#reading-drafts-vs-published-documents:~:text=Important%3A%20the%20draft%20argument%20on%20its%20own%20will%20not%20restrict%20documents%20with%20_status%3A%20%27draft%27%20from%20being%20returned%20from%20the%20API.%20You%20need%20to%20use%20Access%20Control%20to%20prevent%20documents%20with%20_status%3A%20%27draft%27%20from%20being%20returned%20to%20unauthenticated%20users.%20Read%20below%20for%20more%20information%20on%20how%20this%20works.)

We can filter out unpublished documents by adding this to this find call: 

```
    where: {
      _status: {
        equals: 'published',
      },
    },
```

Then we wouldn't need to check for the page.slug since I believe you can't publish a document without a slug. 

I think we should also add this to the [queryByPageSlug](https://github.com/NWACus/web/blob/bug/fix-slugless-posts-pages/src/app/(frontend)/%5Bcenter%5D/%5Bslug%5D/page.tsx#L104) function.

src/app/(frontend)/[center]/[slug]/page.tsx
```diff
@@ -16,7 +16,7 @@ export async function generateStaticParams() {
...
     limit: 1000,
     overrideAccess: true,
```
Comment (busbyk) [inline src/app/(frontend)/[center]/[slug]/page.tsx]: Sort of unrelated but on the topic of API defaults: two things bug me in the way we have a lot of our local API calls written. We use `draft: false` and `overrideAccess: true`. 

These are the default values for these options. [draft defaults to false](By default, the draft argument is set to false) and [overrideAccess defaults to skipping access control](https://payloadcms.com/docs/local-api/access-control#default-behavior-access-control-skipped). 

@rchlfryn @stevekuznetsov how would you all feel about not passing these options? I'm fine with leaving them in if you think they make things clearer but whenever options are set I typically assume they are different than the defaults.

busbyk (changes requested):

src/app/(frontend)/[center]/[slug]/page.tsx
Comment (stevekuznetsov) [reply src/app/(frontend)/[center]/[slug]/page.tsx]: Omitting them SGTM

Comment (rchlfryn) [reply src/app/(frontend)/[center]/[slug]/page.tsx]: Sounds good to me as well! I can omit them in this PR

Comment (busbyk) [reply src/app/(frontend)/[center]/[slug]/page.tsx]: Thanks!

src/app/(frontend)/[center]/[slug]/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/[slug]/page.tsx]: Great Idea - DOn't know why I didn't think of that.

src/app/(frontend)/[center]/posts/[slug]/page.tsx
```diff
@@ -15,15 +15,19 @@ export async function generateStaticParams() {
   const payload = await getPayload({ config: configPromise })
   const posts = await payload.find({
     collection: 'posts',
-    draft: false,
+    draft: false, // does not remove posts with _status: 'draft'
```
Comment (busbyk) [inline src/app/(frontend)/[center]/posts/[slug]/page.tsx]: Can we remove this `draft: false` too? I don't think we need this comment anymore

src/app/(frontend)/[center]/[slug]/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/[slug]/page.tsx]: Can we remove this `draft: false` too? I don't think we need this comment anymore.

busbyk (approved):
Looks good pending two optional changes (removing the last two `draft: false` options. 

Thanks for removing all of those explicitly defined default options!

src/app/(frontend)/[center]/[slug]/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/[slug]/page.tsx]: Oh weird - this got adde back in with some poor conflict resolution

src/app/(frontend)/[center]/posts/[slug]/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/posts/[slug]/page.tsx]: This oddly got added back in I think with improperly resolved merge conflicts.

PR #205, opened 2025-05-19
Update blog permissions

rchlfryn:
### Description
This PR solves 2 issues I recognized when trying to create a blog post from `contributor@nwac.us`.
* **Add read access to roles**
    * When logging in as `contributor@nwac.us` `media`, `pages` and `posts` were missing from the admin panel. I added `read` permissions to the roles and the collections then showed in the admin panel. 
    * I updated the `roleAssignments` permissions since the same issue will occur. I also added `delete` permissions because it seems like a bad experience not to be able to delete a post. 
* **Add `tenant` to contributor permissions**
    * When uploading media as `contributor@nwac.us`, the upload would fail and show the error `The following field is invalid: Avalanche Center`. Upon further debugging, it appeared `tenantField` did not populate making [prefixFilename](https://github.com/NWACus/web/blob/main/src/collections/Media.ts#L23) error out since it is looking for `tenant in media` and `tenant` was not even passed as a property. When I updated the permissions to include `tenant` access in the role, and I debugged `prefixFilename`, the media passed had `tenant: 1`. 
    * This is confusing because we are giving [tenant read access to all users](https://github.com/NWACus/web/blob/main/src/collections/Tenants/index.ts#L49-L51), so it must not be behaving how we intend it to be? I also tried updating the tenant access config to include `create: () => true` and/or `update: () => true` but it still did not populate the tenant filed on media upload. 
    * I do not think this is the right solution but here is the problem since it would be very easy for someone to remove tenant access from permission, but I am out of my depth
    * I believe this is the same problem noted in the [TODO in the Media config.](https://github.com/NWACus/web/blob/main/src/collections/Media.ts#L18)

### DIFF
--- a/src/endpoints/seed/index.ts
+++ b/src/endpoints/seed/index.ts
@@ -288,8 +288,12 @@ export const seed = async ({
       name: 'Contributor',
       rules: [
         {
-          collections: ['posts', 'pages', 'media', 'tenants'],
+          collections: ['posts', 'pages', 'media'],
           actions: ['*'],
+        },
+        {
+          collections: ['tenants'],
+          actions: ['read'],
         },
       ],
     },
--- a/src/endpoints/seed/index.ts
+++ b/src/endpoints/seed/index.ts
@@ -288,7 +288,7 @@ export const seed = async ({
       name: 'Contributor',
       rules: [
         {
-          collections: ['posts', 'pages', 'media'],
+          collections: ['posts', 'pages', 'media', 'tenants'],
           actions: ['*'],
         },
       ],
--- a/src/endpoints/seed/index.ts
+++ b/src/endpoints/seed/index.ts
@@ -280,7 +280,7 @@ export const seed = async ({
       rules: [
         {
           collections: ['roleAssignments'],
-          actions: ['create', 'update'],
+          actions: ['create', 'read', 'update'],
         },
       ],
     },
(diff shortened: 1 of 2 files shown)


src/endpoints/seed/index.ts
```diff
@@ -280,16 +280,16 @@ export const seed = async ({
...
         },
       ],
     },
     {
       name: 'Contributor',
       rules: [
         {
-          collections: ['posts', 'pages', 'media'],
-          actions: ['create', 'update'],
+          collections: ['posts', 'pages', 'media', 'tenants'],
```
stevekuznetsov: Allowing mutating verbs on `tenants` would allow, for instance, the contributors to change the name of the AC or the websites associated with it - which is likely not what we would like. Do you know what specific permission over tenants was necessary? Perhaps just `read` would suffice?

rchlfryn: @stevekuznetsov Agreed. I updated the rules. But supposedly we have a global read on `tenants`.
>This is confusing because we are giving [tenant read access to all users](https://github.com/NWACus/web/blob/main/src/collections/Tenants/index.ts#L49-L51), so it must not be behaving how we intend it to be?

Do you think this is something we should look into?

stevekuznetsov: I don't quite understand the internals of Payload enough to discern the difference between the `() => true` access control and what you're doing here with the roles, and I agree that this is super surprising. Let's note it as a follow-up we should investigate - either to fix some bug on our end or report an issue upstream for the Payload folks.

PR #208, opened 2025-05-19
NAC capabilities and metadata + forecast zones in navigation

### LINKED ISSUE #172: Enable/disable features of the navigation based on NAC capabilities API

busbyk:
Resolves #172 and #4 

This PR introduces an NAC service and convenience functions for fetching an avalanche center's metadata and capabilities. 

The metadata is used to generate links to specific forecast zones in the navigation. 

The capabilities (i.e. platforms) are used to determine if:
1. to generate nav links for the respective widget page
2. allow a user to visit the respective widget page

This also adds an AvalancheCenter provider and context and associated useAvalancheCenter hook. This context/hook isn't used anywhere currently but I figured we'll likely want this at some point. We can certainly take this out and add it back in later if the team would prefer.


src/app/(frontend)/[center]/forecasts/avalanche/[zone]/page.tsx
```diff
@@ -36,7 +38,13 @@ type PathArgs = {
 
 export default async function Page({ params }: Args) {
   const { center, zone } = await params
-  // TODO: how do we pass zone to NACWidget? expects #/slug ...
+
+  const avalancheCenterPlatforms = await getAvalancheCenterPlatforms(center)
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/forecasts/avalanche/[zone]/page.tsx:44]: Is there some magic using this data in this page? I'm a bit lost as to what we do with this once we fetch it.

stevekuznetsov (approved):

src/app/(frontend)/[center]/forecasts/avalanche/[zone]/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/forecasts/avalanche/[zone]/page.tsx:44]: Hmm I'm just using it below to check to see if forecasts is enabled for this center and throwing a notFound response if not. 

That's all it's being used for here. The function's response is being cached via next.js's fetch implementation.

Comment (busbyk) [reply src/app/(frontend)/[center]/forecasts/avalanche/[zone]/page.tsx:44]: I probably could have left the // TODO in since I'm removing it in #198 if that was misleading

Comment (stevekuznetsov) [reply src/app/(frontend)/[center]/forecasts/avalanche/[zone]/page.tsx:44]: Ah, makes sense.

PR #212, opened 2025-05-22
Middleware fix

busbyk:
I had tested with media|thumbnail inside the negative lookahead in the matcher pattern and then decided it would look more logical to re-order them but accidentally added them outside of the negative lookahead and didn't do a full regression test of navigating from the root domain to a subdomain 🤦 

This moves them back inside the negative lookahead, fixing a bug where we can't navigate to subdomains from the root domain.


stevekuznetsov:
https://xkcd.com/1313/

PR #222, opened 2025-05-29
Footer: Make icons clickable and on brand color

rchlfryn:
#201 got merged early. Wrapping up #201


stevekuznetsov:
:upside_down_face:

Issue #206, opened 2025-05-19
Default NAC widgets version to "latest"


busbyk:
When the NAC widgets start getting published under a "latest" tag we should update our code in a few places:

1. The seed script should default the widget version to latest: [here](https://github.com/NWACus/web/blob/nac-widgets/src/endpoints/seed/index.ts#L54), [here](https://github.com/NWACus/web/blob/nac-widgets/src/endpoints/seed/index.ts#L892)
2. The version should fallback to latest if not set in global: [here](https://github.com/NWACus/web/blob/nac-widgets/src/utilities/getNACWidgetsConfig.ts#L6)

This issue should also add an API route for the nac-widget-loader build pipeline to hit to update the widget version and add documentation to the nac widget config global's description.

busbyk:
Closing this due to a Safari-specific bug that is a blocker to us using "latest" tag. We want to use rel="preload" but Safari aggressively caches any script loaded with that so that using "latest" will not pull new versions even when the ETag has changed. See https://mthadley.com/posts/safari-and-early-hints

PR #224, opened 2025-05-29
NAC widget loading adjusted for manual mount control

busbyk:
Dependent on afp-public-widget changes: https://github.com/NationalAvalancheCenter/afp-public-widgets/pull/187

Controls mounting/unmounting of NAC widgets in our own implementation by setting controlledMount to true in widget config objects. 

Uses next/script and onReady to mount widgets. We can't use the beforeInteractive strategy because we can't use onReady with that strategy. 

Updates the default widget version since that's now the minimum version that supports this controlled mounting change. 

I believe we could use a "latest" tag because S3/Cloudfront should update the ETag. The browser will only check the ETag/make another request for the script on page refresh/new session but I think that's fine. 

This issue should introduce the "latest" tag default and also an API route for the nac-widget-loader repo's pipeline to hit to update the widget version: https://github.com/NWACus/web/issues/206


src/components/NACWidget/index.tsx
```diff
@@ -16,30 +61,16 @@ export function NACWidget({
...
-    const safeBaseUrl = widgetsBaseUrl.replace(/\/+$/, '')
-
-    const scriptUrl = `${safeBaseUrl}/${widgetsVersion}`
-
-    // Base URL (used for Google Analytics)
-    const baseUrl = window.location.pathname
+  const { scriptName, widgetDataKey, widgetControllerKey } = loadersByWidget[widget]
 
+  useEffect(function loadWidgetCss() {
     const loadCSS = (url: string): Promise<string> => {
```
stevekuznetsov: Is there any downside/reason to declare this function definition inside of the useEffect as opposed to just at the top-level scope of this file? We're not closing over any state that I can tell.

src/components/NACWidget/index.tsx
```diff
@@ -65,91 +96,61 @@ export function NACWidget({
...
+
+  return (
+    <>
+      {/* container used for css selectors in src/app/(frontend)/[center]/nac-widgets.css */}
+      <div id="widget-container" data-widget={widget}>
+        <div id={`nac-widget-${widget}`} />
+      </div>
+      <Script
+        src={`${scriptUrl}/${scriptName}.js`}
+        strategy="lazyOnload"
```
stevekuznetsov: Unless my read is wrong, it's `onLoad` that doesn't work with `beforeInteractive`, not `onReady` [[ref](https://nextjs.org/docs/pages/api-reference/components/script#onload)] - if `beforeInteractive` functions as expected, I think we should prefer it since it will give users the best experience.

src/utilities/getNACWidgetsConfig.ts
```diff
@@ -3,7 +3,7 @@ import { getCachedGlobal } from './getGlobals'
 export async function getNACWidgetsConfig() {
   const nacWidgetsConfig = await getCachedGlobal('nacWidgetsConfig', 1)()
 
-  const version = nacWidgetsConfig?.version || '20250313'
+  const version = nacWidgetsConfig?.version || '20250602'
```
rchlfryn: These fallback are interesting. I wonder if it's worth it to have or is just a headache to remember to update it in 2 places.

busbyk: Yea it's not ideal to have a specific version as the fallback. We were going to use a `latest` tag (#206) but then realized that the browser would still cache this and not recognize a difference but then realized that the ETag should be different which would resolve that caching issue. So I'd like to set this to latest in #206. 

But an alternative would be to throw or warn.

src/components/NACWidget/index.tsx
busbyk: Nope, we could declare it outside of the component.

src/utilities/getNACWidgetsConfig.ts
stevekuznetsov: We'd only ever hit this if the DB had no entry for the `nacWidgetsConfig` global, right? Throw seems reasonable given that we can ensure that never happens, right?

stevekuznetsov: But if this just sits at `latest` and doesn't change after #206 then /shrug

busbyk: I'm down to update to throw if these values are falsy.

src/components/NACWidget/index.tsx
busbyk: Yea we're getting into some weird territory here. You're right that the docs don't mention `onReady` - they just mentioned `onLoad` but in my testing the `onReady` function will never be called when using `beforeInteractive`. 

This did prompt me to re-check my assumptions though. 

I tried the beforeInteractive approach: 
- onReady functions do not get called so we can't detect when the script is ready but it's pretty reasonable to assume that the `window.{widget}Widget` objects will be available with the beforeInteractive strategy. We could use an interval to continually check to see if this object is available for more reliability. 
- Since beforeInteractive scripts start loading before Next.js code we need to register all of the `window.{widget}WidgetData` config objects before any of the widget scripts load. So we can move that to an inline <Script strategy="beforeInteractive"/> in `src/app/(frontend)/[center]/layout.tsx`
- beforeInteractive scripts will only load/execute on a full page refresh. So if we have the next/script <Script> tags in the NACWidget component the script will not be loaded on client side navigation. So we can move _all_ of the widget scripts to `src/app/(frontend)/[center]/layout.tsx` which will _always_ load all of the widget scripts. 

I thought this might work but the forecast and observations widgets will not mount correctly on client side navigation for some reason. The other widgets are fine. I'm actually a little stumped here. Putting breakpoints in the widget mount functions indicates that they are in fact being called and the `window.{widget}Widget._mounted` flag indicates they have been mounted correctly but nothing is rendered. If I manually `window.{widget}Widget.unmount()` and then `window.{widget}Widget.mount()` in the console the widget re-mounts correctly. 

So it's possible there's an issue in my code which could be resolved but I don't love the beforeInteractive approach because we'd have to always load all of the widget scripts regardless of what page you're on because you might client side navigate to a widget page. 

It also feels like the incorrect usage of the beforeInteractive strategy in general. From the Next.js docs:
> beforeInteractive scripts must be placed inside the root layout (app/layout.tsx) and are designed to load scripts that are needed by the entire site (i.e. the script will load when any page in the application has been loaded server-side).

But we definitely can use afterInteractive (the default) instead of lazyOnLoad. I'm not sure what my thought process was for using lazyOnLoad here instead of afterInteractive to be honest.

stevekuznetsov: Thanks for digging in, this is definitely a sharp edge of Next.js it seems. My thought process was basically - how do we limit the number of server round-trips, JS parsing, execution, etc, that we do before we show the content to the user, given that the "whole point" of the site is to show them their forecasts. I agree that loading them all, always, is a huge shame but if that's what it takes to get this frankenstein to work I would be OK with it. It's a shame that the mounting function aren't working, and I trust you know when to cut the rabbit-hole. Given that this seems like a huge time-sink it seems like we should go with your gut instinct on what's best and note this down for reasons that doing the content first-class would be better in the future.

stevekuznetsov (approved):

Issue #218, opened 2025-05-23
Blog: filter for posts


rchlfryn:
Add "toolbar" to main blog page that will allow users to 
* filter through blog posts by category
* sort

Issue #217, opened 2025-05-23
Blog: sidebar


rchlfryn:
Ensure sidebar is populated on individual blog posts per the design
https://www.figma.com/design/vAJkcFrHt5gddho3ILBOEx/NWAC-UX-UI-Redesign---Work?node-id=395-23945&t=XVkk16qBbieuTQaB-1

rchlfryn:
Not needed for v1. Moving Removing from Summer 2025

PR #233, opened 2025-06-02
Blog - part 1

### LINKED ISSUE #30: AC staff can create blog entries
Make sure AC staff can create, edit, and delete posts from an admin or contributor role. The posts should follow the subtask requirements

### LINKED ISSUE #215: Blog: Individual post page
Be sure AC staff can add these items to a blog post:
- [x] Photo
- [x] Video embed
- [x] Rich Text

### LINKED ISSUE #216: Blog: author information & meta photo in admin
Date
Author

rchlfryn:
### Description
This is the first PR for updating blog posts. 
* Post Admin UI
    * Reuse post SEO fields on posts - change meta description --> description, meta title --> title and meta image --> featured image
    * Update styling and add fields to sidebar
    * Update rich text to include lists and link options
* Metadata functions
    * Split how we handle metadata for posts and pages
* Style website view for individual blog post and blog pages (blog main, search, pagination)
* Update shadcn avatar component --> `1.1.7`
* Add new component `AuthorAvatar` for avatar, author name and published date to posts

### Issue
Resolves #215 and #216 

### Still need to do...
* Remove Blog search #235
* Breadcrumbs #236
* Blog main toolbar including filter, sort, & search input #218
* Previous/ Next buttons on individual blog page #234
* Sponsors block #238

### DIFF
--- a/src/blocks/MediaBlock/Component.tsx
+++ b/src/blocks/MediaBlock/Component.tsx
@@ -34,7 +34,7 @@ export const MediaBlock = (props: Props) => {
   return (
     <div
       className={cn(
-        '',
+        'my-4',
         {
           container: enableGutter,
         },
@@ -43,7 +43,8 @@ export const MediaBlock = (props: Props) => {
     >
       {(media || staticImage) && (
         <Media
-          imgClassName={cn('border border-border rounded-[0.8rem]', imgClassName)}
+          className="my-0"
+          imgClassName={cn(imgClassName)}
           resource={media}
           src={staticImage}
         />
@@ -51,14 +52,13 @@ export const MediaBlock = (props: Props) => {
       {caption && (
         <div
           className={cn(
-            'mt-6',
             {
               container: !disableInnerContainer,
             },
             captionClassName,
           )}
         >
-          <RichText data={caption} enableGutter={false} />
+          <RichText data={caption} enableGutter={false} className="text-xs" />
         </div>
       )}
     </div>
--- a/src/components/AuthorAvatar/index.tsx
+++ b/src/components/AuthorAvatar/index.tsx
@@ -1,41 +1,54 @@
 'use client'
 
-import { Post } from '@/payload-types'
+import { Media, Post } from '@/payload-types'
 
 import { getAuthorInitials } from '@/utilities/getAuthorInitials'
+import { cn } from '@/utilities/ui'
 import { format, parseISO } from 'date-fns'
 import { Avatar, AvatarFallback, AvatarImage } from '../ui/avatar'
 
 export const AuthorAvatar = (props: { authors: Post['authors']; date: Post['updatedAt'] }) => {
   const { authors, date } = props
 
+  const combinedAuthorsNames: string[] = [],
+    combinedAuthorsInitials: string[] = [],
+    combinedAuthorsPhotos: Media[] = []
+  authors.forEach((author) => {
+    if (
+      author &&
+      typeof author !== 'number' &&
+      typeof author.name === 'string' &&
+      typeof author.photo !== 'number'
+    ) {
+      combinedAuthorsNames.push(author.name)
+      combinedAuthorsInitials.push(getAuthorInitials(author.name ?? ''))
+      combinedAuthorsPhotos.push(author.photo)
+    }
+  })
+
   return (
     <>
-      {authors?.map((author) => {
-        let initials, id, authorPhoto
-        if (author && typeof author !== 'number') {
-          initials = getAuthorInitials(author.name ?? '')
-          id = author.id
-          authorPhoto = typeof author.photo !== 'number' && author.photo.url
-        }
-
-        return (
-          <div className="flex items-center mb-6" key={id}>
-            <Avatar className="me-4 size-[60px]">
-              <AvatarImage src={authorPhoto || '/placeholder.svg'} />
-              <AvatarFallback>{initials}</AvatarFallback>
+      <div className="flex items-center mb-6">
+        <div className={cn(`${combinedAuthorsPhotos.length > 1 && 'flex -space-x-2'}`, 'me-4')}>
+          {combinedAuthorsPhotos.map((authorPhoto, index) => (
+            <Avatar className="size-[60px]" key={index}>
+              <AvatarImage src={authorPhoto.url || '/placeholder.svg'} />
+              <AvatarFallback>{combinedAuthorsInitials[index]}</AvatarFallback>
             </Avatar>
-            <div className="flex flex-col">
-              {author && typeof author !== 'number' && (
-                <p className="text-sm text-brand-600"> {author.name}</p>
-              )}
-              {date && (
-                <p className="text-xs text-brand-400">{format(parseISO(date), 'MMMM d, yyyy')}</p>
-              )}
-            </div>
-          </div>
-        )
-      })}
+          ))}
+        </div>
+        <div className="flex flex-col">
+          <p className="text-sm text-brand-600">
+            {combinedAuthorsNames.length > 1
+              ? combinedAuthorsNames.join(', ')
+              : combinedAuthorsNames[0]}
+          </p>
+
+          {date && (
+            <p className="text-xs text-brand-400">{format(parseISO(date), 'MMMM d, yyyy')}</p>
+          )}
+        </div>
+      </div>
     </>
   )
 }
--- a/src/components/AuthorAvatar/index.tsx
+++ b/src/components/AuthorAvatar/index.tsx
@@ -32,7 +32,7 @@ export const AuthorAvatar = (props: { authors: Post['authors']; date: Post['upda
         <div className={cn(`${combinedAuthorsPhotos.length > 1 && 'flex -space-x-2'}`, 'me-4')}>
           {combinedAuthorsPhotos.map((authorPhoto, index) => (
             <Avatar className="size-[60px]" key={index}>
-              <AvatarImage className="object-cover" src={authorPhoto.url || '/placeholder.svg'} />
+              <AvatarImage src={authorPhoto.url || '/placeholder.svg'} />
               <AvatarFallback>{combinedAuthorsInitials[index]}</AvatarFallback>
             </Avatar>
           ))}
--- a/src/components/AuthorAvatar/index.tsx
+++ b/src/components/AuthorAvatar/index.tsx
@@ -32,7 +32,7 @@ export const AuthorAvatar = (props: { authors: Post['authors']; date: Post['upda
         <div className={cn(`${combinedAuthorsPhotos.length > 1 && 'flex -space-x-2'}`, 'me-4')}>
           {combinedAuthorsPhotos.map((authorPhoto, index) => (
             <Avatar className="size-[60px]" key={index}>
-              <AvatarImage src={authorPhoto.url || '/placeholder.svg'} />
+              <AvatarImage className="object-cover" src={authorPhoto.url || '/placeholder.svg'} />
               <AvatarFallback>{combinedAuthorsInitials[index]}</AvatarFallback>
             </Avatar>
           ))}
--- a/src/components/AuthorAvatar/index.tsx
+++ b/src/components/AuthorAvatar/index.tsx
@@ -1,0 +1,41 @@
+'use client'
+
+import { Post } from '@/payload-types'
+
+import { getAuthorInitials } from '@/utilities/getAuthorInitials'
+import { format, parseISO } from 'date-fns'
+import { Avatar, AvatarFallback, AvatarImage } from '../ui/avatar'
+
+export const AuthorAvatar = (props: { authors: Post['authors']; date: Post['updatedAt'] }) => {
+  const { authors, date } = props
+
+  return (
+    <>
+      {authors?.map((author) => {
+        let initials, id, authorPhoto
+        if (author && typeof author !== 'number') {
+          initials = getAuthorInitials(author.name ?? '')
+          id = author.id
+          authorPhoto = typeof author.photo !== 'number' && author.photo.url
+        }
+
+        return (
+          <div className="flex items-center mb-6" key={id}>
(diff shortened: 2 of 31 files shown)


src/app/(frontend)/[center]/search/page.tsx
```diff
@@ -93,7 +93,7 @@ export default async function Page({
       </div>
 
       {posts.totalDocs > 0 ? (
-        <CollectionArchive posts={posts.docs as CardPostData[]} />
+        <CollectionArchive posts={posts.docs as unknown as PostPreviewHorizontalData[]} />
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/search/page.tsx]: I don't like this. I made a ticket #235 to update search and this type.

src/components/AuthorAvatar/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Avatar, AvatarFallback, AvatarImage } from '../ui/avatar'
+
+export const AuthorAvatar = (props: { authors: Post['authors']; date: Post['updatedAt'] }) => {
+  const { authors, date } = props
+
+  return (
+    <>
+      {authors?.map((author) => {
+        let initials, id, authorPhoto
+        if (author && typeof author !== 'number') {
```
rchlfryn, line 16: This type checking was really annoying for `author` and `date`. I am not sure if there is a better way

src/app/(frontend)/[center]/search/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/search/page.tsx]: Update: we are removing blog search for v1, so this will be removed

src/utilities/generateMeta.ts
```diff
@@ -19,16 +19,16 @@ const getImageURL = (image?: Media | Config['db']['defaultIDType'] | null) => {
   return url
 }
 
-export const generateMeta = async (args: {
-  doc: Partial<Page> | Partial<Post>
+export const generateMetaPage = async (args: {
```
stevekuznetsov: super-nit: `generateMetaForPage`? I started to wonder what a meta-page was :D

src/utilities/generateMeta.ts
```diff
@@ -47,3 +47,32 @@ export const generateMeta = async (args: {
     title,
```
stevekuznetsov, line 54: Is the handling of `url` from `doc.slug` here correct?

src/utilities/generateMeta.ts
```diff
@@ -47,3 +47,32 @@ export const generateMeta = async (args: {
...
+      description: doc?.description || '',
+      images: ogImage
+        ? [
+            {
+              url: ogImage,
+            },
+          ]
+        : undefined,
+      title,
+      url: Array.isArray(doc?.slug) ? doc?.slug.join('/') : '/',
```
stevekuznetsov: This might be a hold-over from the website template that doesn't work for our site-map.

src/utilities/generateMeta.ts
```diff
@@ -47,3 +47,32 @@ export const generateMeta = async (args: {
     title,
   }
 }
+
+export const generateMetaPost = async (args: {
+  customTitle?: string
+  center: string
+  doc: Partial<Post>
+}): Promise<Metadata> => {
+  const { customTitle, center, doc } = args || {}
```
stevekuznetsov: Why do we need `|| {}`? Aren't we guaranteed to get `args`?

src/utilities/generateMeta.ts
rchlfryn: it's like soup or salad being confused for super salad. I'll update it

src/utilities/generateMeta.ts
rchlfryn: This was already existent so I missed this, but updated since we should always have args

src/utilities/generateMeta.ts
rchlfryn: This is from the template. I will update it.

src/utilities/generateMeta.ts
rchlfryn: No it is not. I will update it.

src/components/CollectionArchive/index.tsx
```diff
@@ -12,12 +15,14 @@ export const CollectionArchive = (props: Props) => {
   return (
     <div className={cn('container')}>
       <div>
-        <div className="grid grid-cols-4 sm:grid-cols-8 lg:grid-cols-12 gap-y-4 gap-x-4 lg:gap-y-8 lg:gap-x-8 xl:gap-x-8">
+        {/* Grid classes */}
+        {/* <div className="grid grid-cols-4 sm:grid-cols-8 lg:grid-cols-12 gap-y-4 gap-x-4 lg:gap-y-8 lg:gap-x-8 xl:gap-x-8"> */}
```
busbyk: nit: Can we remove this commented out code?

tailwind.config.mjs
```diff
@@ -158,7 +158,7 @@ export default {
               p: {
                 marginTop: '1em',
                 marginBottom: '1em',
-                lineHeight: 1.2,
+                lineHeight: 1.4,
```
busbyk, line 162: This p tag specific styling was introduced in [this commit](https://github.com/NWACus/web/commit/ca25e14b36a6e68d14097aabcf0dd24916a20f9d#diff-aff1492b58cdb20ff9554dba377672d03b18f6f9ba80a56ff7d4cc6e60a4ae07) I think in part to set the line height in some of the blocks. 

Is this line-height changed intended to be for all use of the typography plugin or just for blog post copy? If it's intended for just blog post copy you can use tag specific classes in the JSX like `prose-p:{utility}`

src/components/ui/avatar.tsx
```diff
@@ -23,7 +23,7 @@ const AvatarImage = React.forwardRef<
 >(({ className, ...props }, ref) => (
   <AvatarPrimitive.Image
     ref={ref}
-    className={cn('aspect-square h-full w-full object-cover', className)}
+    className={cn('aspect-square h-full w-full', className)}
```
busbyk: I think we want to pull this object-cover class back in since avatar images can get/are getting distorted again now: 
CleanShot 2025-06-04 at 09 33 25

src/components/Media/ImageMedia/index.tsx
```diff
@@ -53,7 +53,7 @@ export const ImageMedia = (props: MediaProps) => {
         .join(', ')
 
   return (
-    <picture>
+    <picture className="my-2">
```
busbyk, line 56: This affects everywhere the ImageMedia component is used which is quite a few places, including the logo in the header as an example. Was this change for something blog-specific?

src/utilities/generateMeta.ts
rchlfryn: It's probably worth looking into how we handle meta data sine I don't think `generateMeta` works properly. I just did not want to conflate this PR

src/components/Card/index.tsx
```diff
@@ -34,11 +33,12 @@ export const Card = (props: {
...
-      <div className="relative w-full ">
-        {!metaImage && <div className="">No image</div>}
-        {metaImage && typeof metaImage !== 'string' && <Media resource={metaImage} size="33vw" />}
+      <div className="relative w-full">
+        {featuredImage && typeof featuredImage !== 'number' && (
+          <Media imgClassName="w-full" resource={featuredImage} size="33vw" />
+        )}
       </div>
-      <div className="p-4">
+      <div className="m-4 mt-2">
```
busbyk: I actually think the consistent padding on all sides looks nice 😅 

i.e. this: 
CleanShot 2025-06-04 at 09 40 09

vs. this:
image

Also what's your take on using margin vs. padding in cases like this? I typically use padding because the padding will stay the same even if you use classes like `space-` or `gap-`. I also like that padding is included in the height calculation of an element whereas margin is not. 

Using padding vs. margin doesn't matter for this particular use case - just a conversation starter.

src/blocks/MediaBlock/Component.tsx
```diff
@@ -34,7 +34,7 @@ export const MediaBlock = (props: Props) => {
   return (
     <div
       className={cn(
-        '',
+        'my-4',
```
busbyk, line 37: What is this for? Actually to my comments about padding vs. margin I'm not seeing this have any effect when I remove it haha which I believe is because this block is usually in an element with a prose class which is doing some positioning of it's children.

busbyk (commented):
Are you planning to handle the sidebar in another PR? 
CleanShot 2025-06-04 at 09 31 16

rchlfryn:
> Are you planning to handle the sidebar in another PR? CleanShot 2025-06-04 at 09 31 16

We are not doing the sidebar for v1. But there is an issue for it #217

tailwind.config.mjs
rchlfryn: I thought it was fine to make the global change since the font seemed kind of tight. I am personally a fan of using 1.2 but the mocks have everything at 1.4. Thoughts?

src/components/ui/avatar.tsx
rchlfryn: This was automatically removed  with the component upgrade from shadcn. I added the class on each declaration of `AvatarImage` so this wont happen again.
Maybe I will make a PR with shadcn.

src/components/CollectionArchive/index.tsx
rchlfryn: Yes - I just wanted a GH record of it

src/components/Card/index.tsx
rchlfryn: I agree - I like the secondary image. I will update this.

I still think of margin and padding with the old school border mentality. Margin = outside the border, padding = inside the border. In general I try to use margin as much as I can only use padding when trying to keep the background or enhance the height or width of an object. I still forget about `space` and `gap` which is silly, and probably why I probably overuse margin. 

In this example, I did not want to change the width of the card, but now I see how that could look weird. I can easily talk myself in circles on this topic ha!

src/components/Media/ImageMedia/index.tsx
rchlfryn: Yes this was an intentional universal change. Without it, tailwind puts a margin y `2em` on all photos through `prose` which make captions looks very strange (see below). I thought this was an easier change to add a class. 
| without class | with `my-2` change |
|--------|--------|
|Screenshot 2025-06-05 at 10 40 45 | Screenshot 2025-06-05 at 10 41 28 |

src/blocks/MediaBlock/Component.tsx
rchlfryn: This goes back to me probably using margin in a ridiculous way. I like using overlapping margins because then you don't have to rely on an order of objects. I assume the reason you don't see a change is because there is a `p` tag above the image, and it has the same `my-4` which gives a margin on the bottom, thus making it seem like nothing is happening. I added it to keep the uniform spacing and in case two images where back to back in a blog.

busbyk:
> We are not doing the sidebar for v1. But there is an issue for it #217

Aha OK!

src/blocks/MediaBlock/Component.tsx
busbyk: Ahh yea I tend to using padding and space/gap so I don't think about overlapping margins too much. I had a fun read about margin collapse behavior: https://www.joshwcomeau.com/css/rules-of-margin-collapse/ 😄

src/components/Media/ImageMedia/index.tsx
busbyk: Mm cool sounds good. Probably makes sense to keep that prose-picture default and just override it here.

src/components/ui/avatar.tsx
busbyk: Yea I figured. Yea I do think it's weird they don't have that as a default. Maybe there's a reason but it seems like a good default for an avatar component to me.

tailwind.config.mjs
busbyk: As long as it's intentionally global works for me. I'm good to go with the line height in the mocks!

src/components/Card/index.tsx
busbyk: Haha right? Cool, my tendency is definitely to use padding + gap/space but probably just a style/approach difference.

src/components/ui/avatar.tsx
busbyk: Why wouldn't we make this change in the `src/components/ui/avatar.tsx` component rather than on each declaration of `AvatarImage`?

rchlfryn: Because every time we update `Avatar` from shadcn it overwrites the file  (which is what happened here). Adding it as a class (while annoying) is safer

busbyk: I would prefer to make the change in `Avatar` personally since I don't think we'll update it from shadcn very often and we would see it being overwritten in the diff. I think it would be more likely that we forget to add `object-cover` to an instance of `Avatar`.

rchlfryn: For what it's worth, I saw that it was removed but did not readd it because I thought they fixed it in their update but then forgot to double check. I think maybe the best compromise will be to add it in `Avatar` and I will leave a comment.

busbyk: Cool, sounds good. I understand the concern. shadcn/ui is def a pretty different pattern than most UI/component libraries. 

In reality it's just starter code that provides nice default styles for Radix UI components. The intent is to customize it once copied to your repo. 

I think with that understanding it's reasonable to set the expectation that if we want to pull updates to shadcn components we first need to check if we made any changes that we'll need to persist.

busbyk (approved):
Thanks!

src/components/ui/avatar.tsx
rchlfryn: Oh okay - I guess I didn't realize it was intentionally meant to be altered. how did I miss that? I thought it was a component library itself that used Radix UI components. Maybe we can add this to the decision docs or README

PR #250, opened 2025-06-05
NAC widgets hash handling + observations form route

### LINKED ISSUE #19: User can view and submit observations
This will all happen through the NAC widget in v1

### LINKED ISSUE #220: Incorrect hash values in URLs crash widgets
Reported: https://docs.google.com/document/d/1VbgMbk_SKMAOQoKGSE7wzUQwAzpxF1JX2sKRZem7fxs/edit?disco=AAABkB0FMIM

> Got this fun error that didn’t make sense and it happened cause somehow in my clicking back and forth the browser decided to generate the following url (https://nwac.dev.avy-fx.org/forecasts/avalanche/olympics#/view/observations) I was previously on the obs page and it had a spinny in the widget. Then I went to forecasts and choose olympics. May be a one off but wanted to share it just in case

Image

I think a good solution would be to write a re-usable hook that manages the hash values allowed for our routes based on which NAC widgets they load. Right now there is some hash management happening for the all forecasts route: https://github.com/NWACus/web/blob/main/src/app/(frontend)/%5Bcenter%5D/forecasts/avalanche/%5Bzone%5D/ZoneHashHandler.client.tsx

busbyk:
Resolves #220 
Resolves #19 

This PR:
- Adds a WidgetHashHandler component which accepts an initialHash prop and will ensure that initial hash value is present in the url when visiting a route. Using the `.includes()` check allows navigating back to the /observations route to maintain the filtering selections to stay in the url. This resolves issues where using the browser's back behavior would leave the hash from a previous widget in the url causing errors with the widget's internal routing or visiting a page with an invalid hash that might mess with a widget's internal routing. 
- Adds a /observations/submit route for the observations widget submission form
- Adds a /forecasts/avalanche/archive route for the forecast archive section of the forecasts widget
- Removes the `nac-container` class from the observations form
- Removes the h1 tag from the observations form so we can use our own (should we do this on any other widgets? 🤔)
- Makes the observations widget full width as the afp-public-widgets documentation indicates that this widget should be full width. There are some elements that don't wrap correctly if we nest this within our own container, with our own breakpoints. 
- Adds a Submit Observation button/link on the /observations page
- Sets devMode to true in non production environments to use the AFP staging environment. I wonder if it would be better to control this with an environment variable instead or just manually change the value if we want to test against staging. I wanted to test observation submission and wanted to use staging but we still run into CORS errors so that was unsuccessful. 

/observations/submit page: 
CleanShot 2025-06-05 at 13 38 34

/observations page:
CleanShot 2025-06-05 at 13 37 46

The page heading and Submit Observation button are still wrapped in our `container` class. It just looked too weird to have them be full width with only a little horizontal padding. Open to ideas/suggestions here. Some variations:
CleanShot 2025-06-05 at 13 40 20
CleanShot 2025-06-05 at 13 40 40
CleanShot 2025-06-05 at 13 40 54


src/app/(frontend)/[center]/forecasts/avalanche/archive/page.tsx
```diff
@@ -0,0 +1,87 @@
...
+  return tenants.docs.map((tenant): PathArgs => ({ center: tenant.slug, zone: '' }))
+}
+
+type Args = {
+  params: Promise<PathArgs>
+}
+
+type PathArgs = {
+  center: string
+  zone: string
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/forecasts/avalanche/archive/page.tsx]: Do we need this? Didn't seem used

src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx
```diff
@@ -0,0 +1,38 @@
...
+      observer.observe(container, {
+        childList: true,
+        subtree: true,
+      })
+    }
+
+    return () => {
+      observer.disconnect()
+    }
+  }, [])
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx]: I think the new eslint will say we're missing dependencies in the array here

src/app/(frontend)/[center]/observations/submit/page.tsx
```diff
@@ -0,0 +1,91 @@
...
+      },
+    },
+  })
+  if (tenant.docs.length < 1) {
+    return {
+      title: `Submit Avalanche Observation`,
+    }
+  }
+  return {
+    title: `${tenant.docs[0].name} - Submit Avalanche Observation`,
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/observations/submit/page.tsx]: Nit: what you have is probably right but I just wanted to confirm - since an avalanche observation is a specific thing in the AFP. Is there a more generic term that's used for a field observation?

src/components/NACWidget/index.tsx
```diff
@@ -107,7 +107,7 @@ export function NACWidget({
     const widgetData = {
       googleMapsApiKey: process.env.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY || '',
       centerId: center.toUpperCase(),
-      devMode: false,
+      devMode: process.env.NODE_ENV !== 'production',
```
stevekuznetsov: The only reason this might not be what we always want as opposed to a setting in some collection. Is that the staging API for the AFP often has a bunch of junk data that doesn't really have any meaningful content most of the time.

stevekuznetsov (approved):
Looking good!

src/app/(frontend)/[center]/forecasts/avalanche/archive/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/forecasts/avalanche/archive/page.tsx]: Ope good catch! Removing.

src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx]: Ah yea - man I need to add linting to the pre-commit hook. Are you down for that? I'll do that in another PR.

src/app/(frontend)/[center]/observations/submit/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/observations/submit/page.tsx]: Excellent nitpick because you're right. This should just be "Submit Observation". Observations can have avalanches attached to them which are "Avalanche Observation"s. I'll also update the title of the observations page.

src/components/NACWidget/index.tsx
busbyk: Yea and it's also weird to use staging for this and not staging for the capabilities and metadata APIs. I'll remove this for now. Could always add an env var or multiple to control the use of staging environments for widgets and the capabilities and metadata apis.

src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx
Comment (stevekuznetsov) [reply src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx]: Hopefully also CI catches this now? If not, we should look into why

stevekuznetsov (approved):
LGTM. I don't know what's best for the obs submission button ... I'll let Rachel or Rusty weigh in!

src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx]: Well we just have it set up as a warning not an error so I don't think this would fail CI. We could make it an error.

Comment (stevekuznetsov) [reply src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx]: It melts my brain that `react-hooks/exhaustive-deps` would default to being a warning. Let's make that an error, I can't think of a good reason you would want to not trigger a well-formed effect based on all dependencies ...

Comment (busbyk) [reply src/app/(frontend)/[center]/observations/submit/NACContainerRemover.client.tsx]: There is one scenario where omitting dependencies makes sense: if you only ever want the useEffect to run one time after render. 

And that is the case here but I wrote the useEffect in such a way where it checks to see if the container has already been removed so it doesn't matter if it runs more than once. But there are scenarios where you will want to omit all/some dependencies. 

There are also cases where including certain dependencies will cause way too many re-renders but it would be easy to make the argument that that's a memoization issue and could be resolved elsewhere. 

But I'm fine with making it an error and adding an ignore comment if needed.

PR #253, opened 2025-06-06
in-editor eslint warnings fix

busbyk:
I wasn't getting ESLint error highlighting in VS Code. Was this the case for you too @rchlfryn? I don't think you use VS Code @stevekuznetsov 

`pnpm lint` was working for me but I was missing errors because it wasn't highlighting in editor. 

Also fixes a few eslint warnings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
   },
   "devDependencies": {
     "@eslint/eslintrc": "^3.2.0",
+    "@next/eslint-plugin-next": "^15.3.3",
     "@tailwindcss/typography": "^0.5.15",
     "@types/escape-html": "^1.0.4",
     "@types/node": "22.5.4",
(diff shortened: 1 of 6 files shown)


stevekuznetsov:
I'm set up to consume `eslintrc` - so I get whatever errors CI does I think

package.json
```diff
@@ -92,6 +93,7 @@
     "eslint": "^9.17.0",
     "eslint-config-next": "15.1.0",
     "eslint-config-prettier": "^10.1.1",
+    "eslint-plugin-react-hooks": "^5.2.0",
```
stevekuznetsov, line 96: This is such a silly thing to forget 🤦

stevekuznetsov (approved):

package.json
busbyk: What's so weird to me is that 1) the errors were so hidden in the VS Code output tab nested under an eslint dropdown and 2) that running `pnpm lint` ran without complaining about missing packages. I didn't spend too much time digging as I was doing other things but very strange. Maybe available as peer dependencies of other packages for the cli but not for VS Code's extension.

PR #257, opened 2025-06-09
*: tighten eslint rules, run in CI

### LINKED ISSUE #256: make `react-hooks/exhaustive-deps` error
It melts my brain that `react-hooks/exhaustive-deps` would default to being a warning. Let's make that an error, I can't think of a good reason you would want to not trigger a well-formed effect based on all dependencies ...

_Originally posted by @stevekuznetsov in https://github.com/NWACus/web/pull/250#discussion_r2134224652_

stevekuznetsov:
Not clear why `next lint` doesn't seem to pick up all the rules from the eslint config, so just run both.

Fixes #256


tailwind.config.mjs
```diff
@@ -1,12 +1,13 @@
...
-export default {
+const config = {
   content: [
     './pages/**/*.{ts,tsx}',
     './components/**/*.{ts,tsx}',
     './app/**/*.{ts,tsx}',
     './src/**/*.{ts,tsx}',
   ],
   darkMode: ['selector', '[data-theme="dark"]'],
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
```
stevekuznetsov, line 10: Didn't have time to figure out if there was some other import that could work here.

stevekuznetsov:
blah, can't be from my fork

Issue #187, opened 2025-05-12
MediaBlock Images Broken in preview environments


stevekuznetsov:
[Naive](https://github.com/NWACus/web/blob/d7249ef41e25f95f819e5836a76210b82eaa7cf2/src/blocks/Biography/Biography.tsx#L35) images like:

```jsx
<AvatarImage src={biography.photo.url || '/placeholder.svg'} alt={name} />
```

Render to 

```html
<img class="aspect-square h-full w-full object-cover" alt="Dallas Glass" src="/api/media/file/dallas_glass.jpg">
```

& load fine.

---

Correct images like:

```jsx
        <Media
          imgClassName={cn('border border-border rounded-[0.8rem]', imgClassName)}
          resource={media}
          src={staticImage}
        />
```

Render to:

```html
<img
	alt="imageMountain"
	loading="lazy"
	width="1920"
	height="1080"
	decoding="async"
	data-nimg="1"
	class="object-cover w-full h-full group-hover:scale-105 transition-transform duration-300 ease-in-out overflow-hidden"
	style="color: transparent;"
	sizes="(max-width: 1920px) 3840w, (max-width: 1536px) 3072w, (max-width: 1280px) 2560w, (max-width: 1024px) 2048w, (max-width: 768px) 1536w, (max-width: 640px) 1280w"
	srcset="/_next/image?url=https%3A%2F%2Fnavigation-frontend.preview.avy-fx.org%2Fapi%2Fmedia%2Ffile%2Fimage-post3-1.webp&amp;w=16&amp;q=100 16w,
			/_next/image?url=https%3A%2F%2Fnavigation-frontend.preview.avy-fx.org%2Fapi%2Fmedia%2Ffile%2Fimage-post3-1.webp&amp;w=32&amp;q=100 32w,
			/_next/image?url=https%3A%2F%2Fnavigation-frontend.preview.avy-fx.org%2Fapi%2Fmedia%2Ffile%2Fimage-post3-1.webp&amp;w=48&amp;q=100 48w,
			/_next/image?url=https%3A%2F%2Fnavigation-frontend.preview.avy-fx.org%2Fapi%2Fmedia%2Ffile%2Fimage-post3-1.webp&amp;w=64&amp;q=100 64w,
			/_next/image?url=https%3A%2F%2Fnavigation-frontend.preview.avy-fx.org%2Fapi%2Fmedia%2Ffile%2Fimage-post3-1.webp&amp;w=96&amp;q=100 96w,
[...]

PR #265, opened 2025-06-11
ImageMedia component enhancements

### LINKED ISSUE #144: Updated or new Payload Image component or helpers
There is an existing [component](https://github.com/NWACus/web/blob/main/src/components/Media/ImageMedia/index.tsx) that accepts a payload `Media`, i.e. image object from Payload. We might be able to re-use this with some modifications or it might make more sense to create a new one. 

This task this:
1. Update or remove the data uri placeholder
2. Update or create a new version of this re-usable image component that accepts a payload Media object
3. Export some helper functions that would allow us to generate image props from a payload object if we ever need to bail out of using the component (i.e. in the approach described below). 
4. Update any existing usages of ImageMedia accordingly

              I like this approach [here](https://github.com/shadcn-ui/ui/issues/368#issuecomment-2678650485) using `getImageProps` from next/image. 

But we'd need to do some work on the image we get from payload first similar to what is happening in ImageMedia. 

Like generating the right sizes: https://github.com/NWACus/web/blob/e8c040517dc9cb133613ea135f510d10ea84bddf/src/components/Media/ImageMedia/index.tsx#L49

And potentially everything else in this if block: https://github.com/NWACus/web/blob/e8c040517dc9cb133613ea135f510d10ea84bddf/src/components/Media/ImageMedia/index.tsx#L36

FYI this is where that placeholder data uri is defined: https://github.com/NWACus/web/blob/e8c040517dc9cb133613ea135f510d10ea84bddf/src/components/Media/ImageMedia/index.tsx#L16

At the very least we should remove that haha. 

I actually think we should write our own image wrapper or modify this one to suit our needs a little better.

_Originally posted by @busbyk in https://github.com/NWACus/web/pull/137#discussion_r2059160514_

### LINKED ISSUE #189: Image Block Images Don't Handle Un-Loaded Images Correctly
Our handling for the fallback doesn't seem to correctly handle image sizes.

Image

busbyk:
Partially addresses #98 (better image loading placeholders)
Resolves #189 
Resolves #144 

#187 will be addressed separately.

This PR:
- Generates tiny blur data urls to use for placeholder images while the image is loading instead of the default blur data url for all images (at least all images used by the ImageMedia component)
- Uses blur data url placeholder as a fallback image if image loading fails + text indicating the the image is unavailable
- Fixes the display of images that fail to load in specific components
- Fixes a typo in seed data (I just happened to notice it)
- Removes className prop from ImageMedia and VideoMedia since it's not used but keeps it on Media block since it's used there
- Pulls in a website template change for ImageMedia: https://github.com/payloadcms/payload/commit/18693775e43e70e01e99b8581a48e699fa75fdc5

You can test images failing to load by blocking specific file type network requests. In Chrome dev tools -> ⌘ + Shift + P -> "Show network request blocking" -> Add `*.png` and `*.webp`

Before:
CleanShot 2025-06-11 at 10 15 23
CleanShot 2025-06-11 at 10 15 54
Blur data url placeholders on fast 4g: 
CleanShot 2025-06-11 at 10 20 24

After:
CleanShot 2025-06-11 at 10 17 57
CleanShot 2025-06-11 at 10 18 16
Blur data url placeholders (generated from actual images) on fast 4g: 
CleanShot 2025-06-11 at 10 20 51

Not the best demonstration of the blur data urls since the images we use for seeding this page are the Payload graphics but it demonstrates the idea.


src/app/(frontend)/globals.css
```diff
@@ -134,6 +134,6 @@
     @apply border-border;
   }
   body {
-    @apply bg-background text-foreground min-h-[100vh] flex flex-col;
```
busbyk, line 137: I add this on all my projects. It's easy to have some content expand the `body` beyond the screen widget which results in the user being able to scroll horizontally into some whitespace and the header doesn't expand to fit. We pretty much always want to keep the entire app/site to a max width of the viewport width.

busbyk:
Ah two issues I'm aware of:
1. ~~Images are failing to load in the preview env~~ this is current behavior that will be addressed by #187 
2. ~~To fully address #144 I need to update AvatarImage to use either our ImageMedia component or to extract the sizes generation logic out of ImageMedia to be able to pass srcset prop to AvatarImage (and probably blur data url, etc.)~~ added!

next.config.js
```diff
@@ -4,7 +4,7 @@ import redirects from './redirects.js'
 
 const domain = process.env.SERVER_DOMAIN || process.env.VERCEL_URL
 const url = domain
-  ? `https://${domain}`
+  ? `${domain.includes('localhost') ? 'http' : 'https'}://${domain}`
```
stevekuznetsov, line 7: If incoming `domain` is coming in with `localhost`, do we need the `|| 'http://localhost:3000'` at the end of this expression?

src/utilities/getMediaUrl.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Processes media resource URL to ensure proper formatting
+ * @param url The original URL from the resource
+ * @param cacheTag Optional cache tag to append to the URL
+ * @returns Properly formatted URL with cache tag if provided
+ */
+export const getMediaUrl = (url: string | null | undefined, cacheTag?: string | null): string => {
+  if (!url) return ''
+
+  // Check if URL already has http/https protocol
+  if (url.startsWith('http://') || url.startsWith('https://')) {
```
stevekuznetsov: nit: can we use `new URL(url)` and look at the scheme using structured data, append query parameters using the methods, etc, instead of treating this like a string? I tried to do that in `middleware.ts` and I think (or hope?) that it's a more robust approach that makes fewer assumptions / is better-formed in the face of failures, cache tags that need url-encoding, etc.

stevekuznetsov (approved):
Two tiny comments, otherwise this is awesome :sunglasses:

src/app/(frontend)/[center]/theme-preview/page.tsx
```diff
@@ -387,14 +387,16 @@ export default async function Page({ params }: Args) {
...
-              <Avatar>
-                <AvatarImage src="http://www.gravatar.com/avatar/?d=mp " alt="User Avatar" />
-                <AvatarFallback>CN</AvatarFallback>
-              </Avatar>
-              <Avatar>
-                <AvatarImage src="http://www.gravatar.com/avatar/?d=mp " alt="User Avatar" />
-                <AvatarFallback>AN</AvatarFallback>
-              </Avatar>
+              <MediaAvatar
+                src="http://www.gravatar.com/avatar/?d=mp"
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/theme-preview/page.tsx:391]: ⛏️ Do we want a static fallback we have control over incase gravatar is down?

src/blocks/LinkPreview/Component.tsx
```diff
@@ -38,10 +38,10 @@ export const LinkPreviewBlock = (props: LinkPreviewBlockProps) => {
                 )}
                 key={index}
               >
-                <CardHeader>
+                <CardHeader className="overflow-hidden">
```
rchlfryn, line 41: why is overflow hidden needed here?

src/collections/Media/hooks/generateBlurDataUrl.ts
rchlfryn, line 1: 🥳

src/collections/Media/hooks/prefixFilenameWithTenant.ts
```diff
@@ -0,0 +1,45 @@
...
+
+type BeforeOperationHook = Exclude<
+  Exclude<CollectionConfig['hooks'], undefined>['beforeOperation'],
+  undefined
+>[number]
+export const prefixFilenameWithTenant: BeforeOperationHook = async ({ req }) => {
+  if (req.file) {
+    req.payload.logger.debug(`media: have data ${JSON.stringify(req.data)}`)
+    const media = req.data
+    let tenantSlug: string | undefined = undefined
```
rchlfryn: ⛏️  keep the type consistent `Tenant['slug']`

src/components/Media/index.tsx
```diff
@@ -5,8 +5,9 @@ import type { Props } from './types'
 import { ImageMedia } from './ImageMedia'
 import { VideoMedia } from './VideoMedia'
 
-export const Media = (props: Props) => {
-  const { className, htmlElement = 'div', resource } = props
+export const Media = (props: Props & { className?: string }) => {
+  const { className, ...restProps } = props
```
rchlfryn: ⛏️  `restProps` to `optionalProps` ?

next.config.js
busbyk: Hm in the scenario where the SERVER_DOMAIN env var isn't set locally? The SERVER_DOMAIN env var _should_ be just the hostname without a protocol but obv there's no guarantee of that. I'm planning to improve our domain handling in another PR that's in progress.

src/utilities/getMediaUrl.ts
busbyk: Yea no problem 🤘

src/app/(frontend)/[center]/theme-preview/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/theme-preview/page.tsx:391]: Def not a bad idea but I think I'd be fine with the fallback appearing here since it's just the theme preview page.

src/blocks/LinkPreview/Component.tsx
busbyk: Without it images that failed to load were still expanding beyond the size of the CardHeader.

src/components/Media/index.tsx
busbyk: Hmm they're not optionalProps exactly - it's just that VideoMedia and ImageMedia don't use the className prop. 

childComponentProps?

busbyk: Hmm they're not optionalProps exactly - it's just that VideoMedia and ImageMedia don't use the className prop.

childComponentProps?

src/blocks/LinkPreview/Component.tsx
rchlfryn: I should have pulled this down to look. I thought the overflow would have cut off the image but we should be okay since we are using `object-cover`. I just get nervous with overflow hidden

rchlfryn (approved):

src/components/Media/index.tsx
rchlfryn: `combinedProps`? otherwise just leave it. Not worth a discussion :)

Issue #63, opened 2025-03-18
Shared links on social media apps give nice previews


rustynwac:
per Andy's request: Add support for attaching appropriate og meta (https://ogp.me/) tags to content including forecasts and observations. See source on this page for examples: https://www.sierraavalanchecenter.org/forecasts/#/central-sierra-nevada

Best scenario: 
We have the zone in the URL as opposed to behind the # fragment so the web server knows which zone is getting requested. Dependent on #4

Secondarily: We could embed the current widget into a page that's zone-specific but the widget has a zone picker in it so it's kinda weird

busbyk:
I've been referencing this issue when talking about center-scoped metadata. Right now we don't generate metadata correctly for center-scoped pages - it falls back to some Avy-branded title, description, OG image. 

This issue should update metadata for center-scoped pages. 

Just a note for future me or whoever takes this issue.

busbyk:
Although I suppose this might be a duplicate of https://github.com/NWACus/web/issues/191

busbyk:
I think we're good here. #191 is closed by #395 and I can't think of anything else we wanted to do here before we move away from the widgets. Closing for now but let me know if I'm not thinking of something.

PR #270, opened 2025-06-18
Better domain handling

### LINKED ISSUE #49: Next.js Middleware Used For Domain Multiplexing Causes Hydration Errors
We use Next.js [middleware](https://github.com/NWACus/web/blob/cf46ffd0ebdcf7e0e17018b60e18309fd99a4023/src/middleware.ts#L39) to dynamically re-write sub-domains on the main site to path-prefixes so we can serve many tenants from one domain. However, this causes a hydration issue for the URIs of media:

```
Console Error

Hydration failed because the server rendered HTML didn't match the client. As a result this tree will be regenerated on the client. This can happen if a SSR-ed Client Component used

- A server/client branch `if (typeof window !== 'undefined')`.
- Variable input such as `Date.now()` or `Math.random()` which changes each time it's called.
- Date formatting in a user's locale which doesn't match the server.
- External changing data without sending a snapshot of it along with the HTML.
- Invalid HTML tag nesting.

It can also happen if the client has a browser extension installed which messes with the HTML before React loaded.

See more info here: https://nextjs.org/docs/messages/react-hydration-error

+ srcSet={"/_next/image?url=http%3A%2F%2Fnwac.localhost%3A3000%2Fapi%2Fmedia%2Ffile%2..."}
- srcSet={"/_next/image?url=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fmedia%2Ffile%2Fnwac..."}
+ src={"/_next/image?url=http%3A%2F%2Fnwac.localhost%3A3000%2Fapi%2Fmedia%2Ffile%2Fnw..."}
- src={"/_next/image?url=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fmedia%2Ffile%2Fnwac-im..."}
```

### LINKED ISSUE #91: Set up tests
Let's carve it out as a follow-up, I'll get a GitHub Action set up to run Jest and then the lift to add a test here and there will be smaller.

_Originally posted by @stevekuznetsov in https://github.com/NWACus/web/pull/90#discussion_r2036314541_

### LINKED ISSUE #155: dev.avy-fx.org redirects to production domains in middleware
Our logic to determine when we redirect to "real" domains is:

https://github.com/NWACus/web/blob/ac9fa21f156a304d08caee6975fc4ef859c5c747/src/middleware.ts#L23-L25

Today, this has the unintended effect of redirecting dev.avy-fx.org paths to e.g. nwac.us since that's a "production" deployment of the `main` branch.

In the future, it will have the unintended effect of not allowing us to support new centers with a preview of what their site will look like on the prod site before they switch over.

We should amend the logic to have some environment-driven precondition (which we only set for avy-fx.org) and then have a per-tenant opt-in for redirecting to custom domains.

### LINKED ISSUE #187: MediaBlock Images Broken in preview environments

### LINKED ISSUE #223: Landing Page Links To Centers Broken
Today, previews like https://footer.preview.avy-fx.org/ will have broken links to individual centers' pages:

```html
<a 
  class="p-6 border rounded-lg hover:bg-gray-100 dark:hover:bg-gray-800 transition"
  href="https://nwac.https/">
	<h2 class="mb-2">
		Northwest Avalanche Center
	</h2>
	 <p class="no-underline">
	 	Visit site ➡️
	 </p>
 </a>
```

The logic for generating this is here:

https://github.com/NWACus/web/blob/4905ff73d6859e7d710dd1241cc833c9225d953d/src/app/(frontend)/page.tsx#L40

So it looks like `getServerSideURL` is doing something wrong:

https://github.com/NWACus/web/blob/4905ff73d6859e7d710dd1241cc833c9225d953d/src/utilities/getURL.ts#L3-L8

Which likely means some environment variables are not being set correctly for the Vercel deployments. 

cc @busbyk

busbyk:
Resolves #49 
Resolves #187 
Resolves #155 
Resolves #223 
Resolves #91 

The initial intent of this PR was to resolve the hydration errors we get when images are requested from the root domain from tenant subdomains. I made some other changes while I was in there that (IMO) made sense to do all together. 

- Combines getServerSideURL and getClientSideURL into one function that determines where it's being called. It also accepts a hostname argument to enable tenant custom and subdomain hostnames to be used where appropriate.
- Adds a tenant context/provider/hook mainly to be used for the getURL hostname arg but could be helpful elsewhere too. 
- Moves protocol, root domain, and root site url into a single file and uses a single environment variable, NEXT_PUBLIC_ROOT_DOMAIN instead of one for the server and one for the client.
- Changes the tenants collection's `domains` array field to a single `customDomain` text field since I believe we only ever expect a tenant to have a single custom domain. 
- Uses environment variable PRODUCTION_TENANTS per #155 to control whether a tenant's customDomain or their subdomain is used in urls and in the middleware.
- Cleans up the `.next` directory after seeding because next build warned about it being present in CI

There's still some work needed on sitemaps and on page metadata but this PR already got big so I'd like to handle those on #50 and #63 (this issue might not fully capture what we need to do here though tbh, leaving some comments).


next.config.js
```diff
@@ -2,18 +2,17 @@ import { withPayload } from '@payloadcms/next/withPayload'
...
+const ROOT_SITE_URL = `${PROTOCOL}://${ROOT_DOMAIN}`
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   crossOrigin: 'anonymous',
   images: {
     unoptimized: false,
     remotePatterns: [
-      ...[url]
+      ...[ROOT_SITE_URL]
```
stevekuznetsov: If we expect this to be a single-item thing in the future, could we also do a small refactor here to not have the array + .map.flat?

src/collections/Tenants/index.ts
```diff
@@ -22,26 +22,17 @@ export const Tenants: CollectionConfig = {
...
-      type: 'array',
-      fields: [
-        {
-          name: 'domain',
-          type: 'text',
-          required: true,
-        },
-      ],
-      index: true,
+      name: 'customDomain',
```
stevekuznetsov, line 26: Nice, thanks - this had been bugging me for a while.

src/middleware.ts
```diff
@@ -15,18 +16,16 @@ export const config = {
...
-  { slug: 'sac', domain: 'sierraavalanchecenter.org', id: '2' },
-  { slug: 'snfac', domain: 'sawtoothavalanche.com', id: '3' },
-]
+const PRODUCTION_TENANTS = getProductionTenantSlugs()
 
-const REDIRECT_TO_CUSTOM_DOMAIN =
-  process.env.REDIRECT_TO_CUSTOM_DOMAIN?.toLowerCase() === 'true' ||
-  process.env.NODE_ENV === 'production'
+const TENANTS: { slug: string; customDomain: string; id: string }[] = [
+  { slug: 'nwac', customDomain: 'nwac.us', id: '1' },
```
stevekuznetsov: We need the ID because of how Payload sets the cookie, right? If we end up doing the "multi-tenant admins on a tenant-scoped site only see content for one tenant" do we need to do the cookie handling ourselves again? Only reason I ask is - we may need to just pre-populate tenants in the DB and make this list have them all, otherwise registering a new tenant (and generating its ID) will require a redeploy as we change this as well.

stevekuznetsov (approved):
Would be cool to see some unit tests for the different possibilities for `getUrl` and `getHostnameFromTenant` - which might influence how we factor the env access bits - but otherwise this looks great!

next.config.js
busbyk: Yea I'd love to. I wasn't sure if this was set up as an array to account for multiple domains being used in the future or something so I left it but I'm down to take this out of an array for sure.

src/middleware.ts
busbyk: Thanks for bringing this up, that's a really good point. And now that I'm looking at the tenant cookie part of this again I don't see the point of it. 

I just did some basic testing with this but I don't think the tenant cookie part of this is necessary which is the only thing using the id like you've called out so I think we're good to remove that.

busbyk:
> Would be cool to see some unit tests for the different possibilities for `getUrl` and `getHostnameFromTenant` - which might influence how we factor the env access bits - but otherwise this looks great!

Added some tests!

stevekuznetsov (approved):
:muscle:

PR #271, opened 2025-06-18
Blog: Add tags & add sort for posts

### LINKED ISSUE #218: Blog: filter for posts

rchlfryn:
#### Description
* Add tags collection
* Give user the ability to tag posts
* Add tag filter to main blog post page
* Add ability to sort posts by `published_at` value

#### Issues
Fixes #218

#### Open questions
1. How do we want to handle the main blog page on tablet and mobile? Show a menu with the sidebar options?


src/components/CollectionArchive/index.tsx
busbyk, line 1: Can we call this something else? CollectionArchive is confusing to me. This is really just a list of posts right? Archive makes me think it's...not sure actually - old posts? I think this is just leftover from the website template?

src/endpoints/seed/index.ts
```diff
@@ -178,6 +179,24 @@ export const seed = async ({
...
+    {
+      name: 'Contributor',
+      rules: [
+        {
+          collections: ['posts', 'pages', 'media', 'tags'],
+          actions: ['*'],
+        },
         {
           collections: ['navigations', 'tenants'],
           actions: ['read'],
```
busbyk: Hmm I think this merge is slightly off. In #287 I removed User Administrator and Contributor. And the read on navigations and tenants should be on the Admin role.

busbyk (commented):
The approach you took here uses client side data fetching. Two thoughts about this:

1. There's an error with fetching posts because I removed the permissive read on tenants.slug in #273. Steve and I had a conversation on that PR about the intent of permissive reads. And I think that would be worth a decision doc and/or team discussion tbh. But for this PR to work for me locally I had to add permissive read to tenants.slug and tags.slug. 
2. Client side data fetching obviously works but I think it misses out on the benefits of Next.js and static site generation. 

A more Next.js-y (technical term) pattern would be:
- Fetch posts based on filters server side by using `router.push(`/posts?${filters.toString()}`)`
- The server component (i.e. `src/app/(frontend)/[center]/posts/page.tsx`) fetches data using the Payload Local API based on the query params that represent the search filters (i.e. sort by and tags)

The pagination example from the website template is a good example of this pattern: 
- https://github.com/NWACus/web/blob/blog/tags/src/app/(frontend)/%5Bcenter%5D/posts/page/%5BpageNumber%5D/page.tsx
- router.push: https://github.com/NWACus/web/blob/blog/tags/src/components/Pagination/index.tsx#L32

If you could only have one tag selected at a time you could also use `generateStaticParams` to pre generate each of the tag pages so they would always just get served up from the CDN and you could actually make the tag buttons Next Links. Depends on what we want for functionality there but I actually think that would be nice - only one tag selected at once. 

You could even use Suspense to add a loading state for filtering requests after the initial page load: https://nextjs.org/docs/app/getting-started/fetching-data#with-suspense

Using this server side data fetching approach would resolve the permissions issue with 1. (even though we may want more permissive reads for the AvyApp down the road) and should be more performant since we'll be able to statically generate more and serve more from the CDN. 

I think vertically stacking the filters and tag buttons on mobile is a good mobile approach. Maybe make the buttons smaller and use flex-wrap on them though or use another dropdown menu for those on mobile. 

Happy to discuss further/anything I'm not considering here.

rchlfryn:
@busbyk I agree with you about making this a server component. However when I initially tired this, I was unfamiliar with Next.js patterns so thank you for pointing these out. 
While rewriting this I could not figure out why `searchProps` was empty. It turns out having `force-static` will render a static pages which cannot react to query parameters at runtime, so they will always be empty. So I removed that and just wanted to note it for a knowledge share.

> Depends on what we want for functionality there but I actually think that would be nice - only one tag selected at once.

I left the ability to select multiple tags. I can bring this up with Rusty to see if we want to implement it with only single tag selection and then we can use `generateStaticParams`

>You could even use Suspense to add a loading state for filtering requests after the initial page load: https://nextjs.org/docs/app/getting-started/fetching-data#with-suspense

I saw this too and added it along with a loading skeleton.

src/app/(frontend)/[center]/posts/page.tsx
```diff
@@ -1,76 +1,96 @@
...
+      <div className="container mb-16 flex flex-col-reverse md:flex-row flex-1 gap-6">
+        <div className="grow">
+          {posts.totalDocs > 0 ? (
+            <Suspense fallback={<PostSkeleton />}>
+              <PostCollection posts={posts.docs} />
+            </Suspense>
+          ) : (
+            <h3>There are no posts matching these results.</h3>
+          )}
+        </div>
```
Comment (busbyk) [inline src/app/(frontend)/[center]/posts/page.tsx]: I'm never seeing the loading skeleton even when throttling my network speed down to 3G. I think because the Suspense component isn't being rendered until the posts promise has resolved and totalDocs can be truthy. 

I think you'd want to put the totalDocs check inside of the Suspense boundary. Unless you were able to see the PostSkeleton and I'm missing something here.

src/app/(frontend)/[center]/posts/posts-sort.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  useEffect(() => {
+    setSortOption(initialSort)
+  }, [initialSort])
+
+  useEffect(() => {
+    if (isFirstRender.current) {
+      isFirstRender.current = false
+      return
+    }
```
Comment (busbyk) [inline src/app/(frontend)/[center]/posts/posts-sort.tsx:32]: Nice

busbyk (changes requested):
Excellent! The new approach is great.

Three UI requests:
1. On mobile: can you stack the sort and tag filters vertically?
CleanShot 2025-07-02 at 08 32 47
2. On mobile: can you stack the blog post image and the rest of the blog details vertically?
CleanShot 2025-07-02 at 08 34 31
3. The selected state of tags isn't super clear and depending on what the center's theme's secondary and callout colors are it could be even more confusing. What do you think about adding a checkbox to the toggle? I got similar feedback in AvyObs and used a checkbox and I think it looks much clearer. Could use a gray color for selection and the theme's secondary color for the checkbox (or callout whatever you think makes more sense):
CleanShot 2025-07-02 at 08 37 59@2x

src/app/(frontend)/[center]/posts/posts-tags.tsx
```diff
@@ -0,0 +1,63 @@
...
+      prev.includes(slug) ? prev.filter((s) => s !== slug) : [...prev, slug],
+    )
+  }
+
+  useEffect(() => {
+    if (!hasUserInteracted.current) return
+
+    const params = new URLSearchParams(searchParams.toString())
+    if (selectedTags.length > 0) {
+      params.set('tag', selectedTags.join(','))
```
Comment (busbyk) [inline src/app/(frontend)/[center]/posts/posts-tags.tsx]: Nit: naming this `tags` instead of `tag` might be more clear since it's treated as an array of tag slugs right?

src/app/(frontend)/[center]/posts/page.tsx
```diff
@@ -1,76 +1,96 @@
...
-    select: {
-      authors: true,
-      description: true,
-      featuredImage: true,
-      meta: true,
-      publishedAt: true,
-      slug: true,
-      title: true,
-    },
+    limit: 10,
```
Comment (busbyk) [inline src/app/(frontend)/[center]/posts/page.tsx]: Not necessary: to test pagination I had to change this limit in three places (here, in page/[pageNumber], and in the PageRange component's props). Might be nice to use an exported constant for this value so they're all always in sync.

busbyk:
> @busbyk I agree with you about making this a server component. However when I initially tired this, I was unfamiliar with Next.js patterns so thank you for pointing these out. While rewriting this I could not figure out why `searchProps` was empty. It turns out having `force-static` will render a static pages which cannot react to query parameters at runtime, so they will always be empty. So I removed that and just wanted to note it for a knowledge share.

Good knowledge share, I didn't know that!

src/app/(frontend)/[center]/posts/posts-tags.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/posts/posts-tags.tsx]: Updated

src/app/(frontend)/[center]/posts/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/posts/page.tsx]: I was also not able to see it even when I switched <Suspense> to be outside of the the totalDocs check. 
I even tried wrapping the Page function to something like this...
```tsx
<Suspense fallback={<div>Loading...</div>}>
  <PostSection center={center} sort={sort} selectedTag={selectedTag} />
</Suspense>
```
...  and having the `await` in a separate file like the docs suggested, but still the data is rendered too quickly. We do have a good loading state with the blurry images so I will remove for now.

src/app/(frontend)/[center]/posts/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/posts/page.tsx]: Good call. Added this as a variable. I was only able to find 2 files that changed/ used this const. Can you confirm I am not missing one?  [079e0b4](https://github.com/NWACus/web/pull/271/commits/079e0b43987d6a3f62128e7c7619a2555b906dfc)

busbyk (approved):
Love it. Changes look good!

PR #273, opened 2025-06-18
Use overrideAccess param default value

busbyk:
We've talked about this a little bit before but the access.read === true on the tenants collection's slug field got me thinking about this again. 

The Payload Local API [defaults to overriding access control](https://payloadcms.com/docs/local-api/overview#:~:text=Skip%20access%20control.%20By%20default%2C%20this%20property%20is%20set%20to%20true%20within%20all%20Local%20API%20operations.) since we're only using it on the server. We have control over when it's called so it makes sense to bypass access control. 

If we use that default of bypassing access control I think we can avoid the `read: () => true`s on some fields/collections. 

The only one I'm unsure about is the global read access on roles and globalRoles - what was the purpose of that? I removed and tested and I couldn't see a change in behavior in the admin panel across the different roles. 

Is there any reason we shouldn't make these changes that I'm not thinking of?


stevekuznetsov:
Any content we want to display to users on the app needs to be readable without auth. IIRC there was some downside (did the roles not get added to the JWT?) of not having them readable, I can't remember - but if you removed without side-effect, we can nix it.

busbyk:
> Any content we want to display to users on the app needs to be readable without auth.

Yea I think we want to accomplish this by using the Payload Local API which bypasses auth. At least that's how I'm thinking about things but wanted a sanity check. 

> IIRC there was some downside (did the roles not get added to the JWT?) of not having them readable, I can't remember - but if you removed without side-effect, we can nix it.

Hmm yea I could see that being the case for the roles not being added t the JWT but I can't find any issues after removing these haha so I think we're good.

stevekuznetsov:
> Yea I think we want to accomplish this by using the Payload Local API which bypasses auth. At least that's how I'm thinking about things but wanted a sanity check.

Sorry, when I said "the app" I meant Avy the mobile app - which needs to make normal API calls out through the network. I know we haven't yet started to integrate the two, but unless it's costing us a bunch of time and effort to handle it, keeping the "published content needs to be readable at the API without auth" boundary would help us not have a larger lift when we do the integration.

busbyk:
> > Yea I think we want to accomplish this by using the Payload Local API which bypasses auth. At least that's how I'm thinking about things but wanted a sanity check.
> 
> Sorry, when I said "the app" I meant Avy the mobile app - which needs to make normal API calls out through the network. I know we haven't yet started to integrate the two, but unless it's costing us a bunch of time and effort to handle it, keeping the "published content needs to be readable at the API without auth" boundary would help us not have a larger lift when we do the integration.

Ahhh yea I hadn't considered that! 

Will AvyApp be accessing roles though?

stevekuznetsov:
>Will AvyApp be accessing roles though?

Nope! Just the pages/events/blogs likely.

Issue #229, opened 2025-05-29
Tenant can update privacy policy/terms of use


rustynwac:
Tenant can add privacy policy/terms links for their organization

rustynwac:
NWAC to provide content

rustynwac:
@busbyk I may not have been clear on this. The super admin will update the privacy policy and terms of use for all ACs. So one privacy policy + one terms of use for every center that I will manage.

busbyk:
> [@busbyk](https://github.com/busbyk) I may not have been clear on this. The super admin will update the privacy policy and terms of use for all ACs. So one privacy policy + one terms of use for every center that I will manage.

Oh no you were clear. The privacy policy/terms of service being in the website settings was just a hold over from the footer fields being merged into website settings.

rustynwac:
Frig. I was wrong on how this works! A board member lawyer thought we could cover everyone in one Privacy/Policy + Terms of Use but after talking to our actual lawyers I learned this is by organization. So each tenant will need control of these links.

stevekuznetsov:
Does it need to be 100% different or can it be some kind of form/fill adapted between ACs? Will each AC need a laywer to draft it?

rustynwac:
To do it right, each AC should have a lawyer draft one up and they should be posted and maintained by them. They could steal language but it's technically up to that lawyer to decide whether or not to steal the language from NWAC or another AC.

busbyk:
Ah bummer. Well this means that we don't have to do anything on our side but unfortunate for ACs. Sharing a single one would have been nice for them. 

We currently have the privacy policy and terms of service links controlled in the Website Settings and those are just Pages. So the only thing we'll really want to do is make sure skeleton pages get created as part of onboarding.

Issue #72, opened 2025-03-20
Website team can preset theming and styles for an AC


rustynwac:
When onboarding a new center, we need the ability to set the AC's colors logo and branding

stevekuznetsov:
Happened to be chatting about this with @busbyk today - we will likely want to differentiate between the following

- a 1:1 aspect ratio logo/favicon, perhaps without text, like [NWAC's](https://files.nwac.us/wp-content/uploads/2020/10/06090559/cropped-NWAC-rectangle-blue-logo-32x32.jpg)
- a "compact" logo, like [NWAC's](https://github.com/NWACus/avy/blob/main/assets/logos/NWAC.png)
- a "full" or "banner" logo, like [NWAC's](https://files.nwac.us/wp-content/uploads/2020/10/08140810/nwac-logo-usfs.png)
  - need to determine our guidelines for appending the USFS logo

The collection where users upload this stuff needs to enforce size requirements and aspect ratios - likely best for us to ask for a vector graphic if possible or a high-resolution raster if not, and we can generate the correct sizes for different use-cases.

rustynwac:
Not all ACs will have USFS branding but I can look into requirements. I think this exists somewhere.

rustynwac:
Helpful? [USFS branding guidelines](https://www.fs.usda.gov/sites/default/files/fs_media/fs_document/Branding-Standards.pdf)

stevekuznetsov:
Great link - looks like we'd fall under partner products with financial assistance. Key points:

>USE OF INSIGNIA BY PARTNERS
In order for a cooperator or partner to use the Forest Service
insignia on its web, social media, large publications, or
videos, a formal relationship must exist. Relationships may be
established through an agreement, such as a memorandum
of understanding, cooperative agreement, or an interagency
agreement (Forest Service Manual [FSM] 1632.2, FSM 1641.1, and
FSM 1660 [see chapter 1666 for guidance specific to websites].
The insignia should only be used for identification and not to
suggest the agency endorses one product or organization over
another. Use caution when identifying the agency when working
with for-profit organizations. If the Forest Service is not the
primary organization, the insignia may be used alone but should
not be placed before the lead or other organization’s logos.
The insignia should be the same size or smaller than the lead
partner’s and other logos. The product should not appear to be
a Forest Service product. It should be very clear that the product
was produced by the partner.

> Financial Assistance
If the partner is a recipient of Federal financial assistance,
ensure they include a public notification of nondiscrimination
[...]

Issue #114, opened 2025-04-15
Website team can onboard a new AC


rustynwac:
Rusty + co need to be able to easily add a new AC. Some features worth considering
- templates for certain pages
- a script to copy over a site or template of a site

busbyk:
A few additional things we need for onboarding (which might make sense to create separate issues for but just wanted to get these written down somewhere):

- A script to generate seed pages for the template tenant so we have a starting point for building some pre-built pages. This would be similar to our seed script but for just a single tenant. 
- A script with or without some custom UI for copying all pages from a tenant to another tenant. @rchlfryn, @rustynwac said you may already be working on this?
- Make sure we include at least a skeleton page for the Privacy Policy and Terms of Service and relate those to a tenants' respective Website Settings fields. #229

rchlfryn:
@busbyk moved notes to #353

PR #277, opened 2025-06-19
Website template cleanup

### LINKED ISSUE #98: Payload website template cleanup and updates

### LINKED ISSUE #263: add description to tenant or a settings global collection for og generation
Let's file an issue?

_Originally posted by @stevekuznetsov in https://github.com/NWACus/web/pull/258#discussion_r2136830109_

busbyk:
Resolves #98 
Resolves #263 

- Upgrades Payload to version 3.43.0
- Creates a Settings collection (appears as Website Settings in admin panel) which is intended to be a single place for a center admin to control website-wide settings.
- Changes "Globals" group to "Settings". I think this will make more sense to users. 
- Removes Themes and Palettes collections since we've hardcoded these for the first few centers
- Removes ColorField used by Palettes
- Removes Brands collection. Moves those fields into Settings. 
- Adds separate icon asset and usfsLogo asset. Separates the Banner and the USFS Logo if the center uploads one. Left as optional since it's possible a center wouldn't have an affiliation? Or they could just use a single banner file if they prefer. 
- Pulls in a [small change from the website template](https://github.com/payloadcms/payload/commit/5b554e5256b2e59043280185723fee54eca7745d) - why not? 
- Moves the Footers fields into the Settings global collection
- Removes revalidateFooter hook since it's no longer a global and not queried using the getCachedGlobal function
- Moves the Footer component under @/components to match other components
- Adds a max height + object-fit: contain to footer logo (uses same logo as the settings.logo image)
- Uses a column layout for footer elements on mobile

This mostly addresses the graphics discussion in #72. I'd prefer to enforce an aspect ratio and dimensions for icon, logo but I _think_ that we would need to create separate collections to define different imageSizes and validations. I don't believe you can override those attributes on an upload relation field. See [my Discord question](https://discord.com/channels/967097582721572934/1385344002680688771/1385344002680688771) (no answers yet). 

Also worth calling out that the settings tab has a legal pages section now because they were pulled over from the footer collection but per #229 we'll remove those and use globals for every tenant.


busbyk:
I was also considering moving fields from the Footers global collection into Settings. I'm thinking that it will be logical to keep settings in one place. Some of these things aren't necessarily footer-specific either. 

These could go under the Settings General tab: 
CleanShot 2025-06-20 at 15 56 02@2x
We might as well use the tenant name here instead of having a user type the name in multiple locations. We also might as well use the same logo right?

These could go under a Settings -> Social Media tab: 
CleanShot 2025-06-20 at 15 57 26@2x

And then the terms of service and privacy policy page links - not sure where I'd nest those in Settings. Maybe a Footer tab?

The only complexity would be updating the revalidateFooter hook to run on the Settings collection and only revalidate the footer if fields from settings used in the footer changed. But I think that's fine. 

What are everyone's thoughts?

src/fields/ColorField/index.ts
```diff
@@ -1,45 +0,0 @@
...
-  return {
-    name: name,
-    type: 'text',
-    required: true,
-    // TODO: admin config for custom view that shows off the color, picker or tailwind panels for choosing
-    validate: validateHSLValue,
-  }
-}
-
-const validateHSLValue: TextFieldValidation = (value: string | null | undefined): string | true => {
```
rchlfryn, line 13: Seeing this reminded me we still need to do some work on the "new" color picker. This is probably not a part of this PR but I made an issue (#282) and mentioned this PR to be inspired by this code.

rchlfryn (approved):
I think it's a good idea to move the footer info under settings. It never really felt useful in its own section. However, I could imagine it would be confusing for a user to find this on their own for the first time. Is there a way we can make a note or guide them to this? Might be a good user testing question.

> We might as well use the tenant name here instead of having a user type the name in multiple locations. We also might as well use the same logo right?

Yes. I believe we discussed this. Feel free to remove the name.

> These could go under a Settings -> Social Media tab:

👍 
 
> And then the terms of service and privacy policy page links - not sure where I'd nest those in Settings. Maybe a Footer tab?

Or `Legal Policies`?

busbyk:
> Is there a way we can make a note or guide them to this? Might be a good user testing question.

It might be nice to add some UI for onboarding in the BeforeDashboard component which would direct a tenant admin to their website settings among other things. 

Other things we may want to encourage users to do during a semi-guided onboarding would be filling out their staff and biographies, creating their first blog post, setting up their donate button, etc. etc. We can figure out what those things might be on #114

busbyk:
> I think it's a good idea to move the footer info under settings. It never really felt useful in its own section. However, I could imagine it would be confusing for a user to find this on their own for the first time. Is there a way we can make a note or guide them to this? Might be a good user testing question.
> 
> > We might as well use the tenant name here instead of having a user type the name in multiple locations. We also might as well use the same logo right?
> 
> Yes. I believe we discussed this. Feel free to remove the name.
> 
> > These could go under a Settings -> Social Media tab:
> 
> 👍
> 
> > And then the terms of service and privacy policy page links - not sure where I'd nest those in Settings. Maybe a Footer tab?
> 
> Or `Legal Policies`?

Updated if you'd like to take a look!

PR #278, opened 2025-06-19
Allow a basic user to view their profile page

### LINKED ISSUE #48: GET /api/users/me 403
Something (looking up your own user?) is causing a 403 in the /users/me endpoint.

yuliadub:
I did a thing ❗ 😄 (hopefully in a way that will make sense to not just me) 

fixes https://github.com/NWACus/web/issues/48

### DIFF
--- a/src/collections/Users/access/byGlobalRoleOrTenantDomain.ts
+++ b/src/collections/Users/access/byGlobalRoleOrTenantDomain.ts
@@ -19,6 +19,7 @@ export const byGlobalRoleOrTenantDomain: (method: ruleMethod) => Access =
     }
 
     const roleAssignments = roleAssignmentsForUser(args.req.payload.logger, args.req.user)
+
     const matchingTenantDomains = roleAssignments
       .filter(
         (assignment) =>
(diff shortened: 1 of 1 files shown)


src/collections/Users/access/byGlobalRoleOrTenantDomain.ts
```diff
@@ -8,7 +8,10 @@ import { Access, CollectionConfig } from 'payload'
 export const byGlobalRoleOrTenantDomain: (method: ruleMethod) => Access =
   (method: ruleMethod): Access =>
   (args) => {
-    if (!args.req.user) {
+    const { req } = args
```
stevekuznetsov: nit: I prefer not doing this since it can be hard to follow, and hard to ensure everyone uses the new variables - to wit, you forgot to update `args.req.payload` to `req.payload` on L24. If you prefer having more tightly scoped variables, please destructure in the `(args) => {` signature so `args` isn't in scope at all and it's not possible to make that mistake!

stevekuznetsov (commented):
Nice! One comment, otherwise this looks good.

stevekuznetsov (approved):

Issue #230, opened 2025-06-02
Role management enhancements


busbyk:
There's a few things we want to do here:

1. Determine the final set of roles that we need in production and make sure those are available in prod and that our seed script is updated. 
2. Make the flow of creating new roles and/or assigning roles more intuitive. 

The issues for 2. are:
- The role assignments table having "Untitled" labels for roles and users. 
- Content Hash field shouldn't be visible for global role assignments 
![Image](https://github.com/user-attachments/assets/661aa2c6-afc2-4ad2-81dd-60dcaf5a2b85)
- A center admin cannot currently see their own users. Center admins should be able to see their users and assign roles via the user.
- A center admin should be able to easily create new roles.

busbyk:
Relevant discussion: https://github.com/NWACus/web/pull/281

A few other issues/things to consider:
- If a user is created without any roles and without an email domain matching a tenant they are in this tenant-less state where they can see things that are global read. Right now, that's plugin collections (media, forms, form submissions), and biographies and staffs because those collections have global read (there was some desire in #273 to keep this). 
- ~~A tenant admin can't currently create roles just for their tenant. I would imagine this is something that we want but that's a requirements question.~~
- Do we need to use join tables for roles assigned to a user? Could we just have a relationTo field to make the UI feel more intuitive?
- The way things are setup right now, if we want to have a "super admin" role we can't hide that from admin roles who can read roles.
- ~~We don't currently have a way of limiting access based on a user's own content. Limiting a user to editing only their own posts for example.~~
- If you don't have any tenant-scoped roles the baseListFilter does not apply so you can see content from all tenants.

stevekuznetsov:
> because those collections have global read (there was some desire to keep this)

Unless we have a better option, we need unauth read to the API to support consuming this content in the mobile app. I don't really want to punt this decision for later since it's going to be painful one way or the other. App bundles are just tarballs so any secrets we put in there are not obfuscated and effectively give global read to everyone anyway.

>A tenant admin can't currently create roles just for their tenant. I would imagine this is something that we want but that's a requirements question.

Is this a limitation outside of the tenant admin role definition? Or is it just a matter of factoring that role correctly.

> Do we need to use join tables for roles assigned to a user? Could we just have a relationTo field to make the UI feel more intuitive?

[...]

PR #281, opened 2025-06-20
Use tenant scoped role assignments to determine users permissions

### LINKED ISSUE #276: Tenant admins should be able to manage their own center's users
At one point this functionality was working. This was potentially broken during the multi tenancy plugin updates or during some RBAC change. 

We need a center admin to be able to manage users with access to their tenant.

busbyk:
Resolves #276 

The existing domain check wasn't working because customDomain (or previously domains) wasn't in the defaultPopulate for the tenants collection. And I think that checking the domain is pretty reliable since we know that at least the first three centers use their website domain for their emails and it's very likely this will be the case. But I couldn't help but think there could be a scenario where a tenant's customDomain value is different from the emails they want to allow or they wanted to allow a user with a different email domain register as a user. 

This changes the permissions check to use role assignment tenants instead of domains. So we determine the role assignments on the logged in user that apply to the users collection and determine which tenants they have user collection rules for (might be multiple). Then we filter the permissions to apply for users with role assignments (any) for the same tenants. 

This means that a multicenter user (i.e. multicenter@avy.com in our seed data) can see users from all the centers they have a users collection rule. 

I think this is a more reliable approach but what do you think?


busbyk:
Targeting the website-template-cleanup branch because I've upgraded the Payload version there and this change was causing LibSQL errors on the version we're on in main 😬

stevekuznetsov:
I like the direction but the one downside is that for the "add a new user to the team" flow - which may be the only time an admin interacts with users, realistically speaking - they need to see the candidates for whom they will add tenant-scoped roles, which are (usually) members of their domain and (always) users with no existing role assignments yet.

busbyk:
> I like the direction but the one downside is that for the "add a new user to the team" flow - which may be the only time an admin interacts with users, realistically speaking - they need to see the candidates for whom they will add tenant-scoped roles, which are (usually) members of their domain and (always) users with no existing role assignments yet.

Very good point. 

A few ideas, could combine some:
1. Require at least one role for new users (probably a good idea since it's a safe bet that we want all users to have at least one role). 
2. Check both 1) user's email domains against tenant.customDomain and 2) tenant scoped roles
3. Add a tenant.emailDomains array field on tenants to handle the scenario where a tenant wants to have a user with an email different from their customDomain. This feels unlikely so maybe not. 
4. Add a createdBy attribute on the users collection and make it so you can always see the users you created. 

Thoughts? Feel free to hit me up in Slack too if that's easier. 

Option 2 might be the best of both worlds.

stevekuznetsov:
I'd vote for 2 - the downside with 1 is I really want the entire "onboard a new member" flow to be 100% delegated to some tenant admin on their own time. Unless we can automate some tenant scoped permission for user creation via SSO, 1 seems like it would regress that. Also would be possibly bad ux for the admin to accidentally remove the last role from a user and then not see the user any longer to un-do it.

busbyk:
> I'd vote for 2 - the downside with 1 is I really want the entire "onboard a new member" flow to be 100% delegated to some tenant admin on their own time. Unless we can automate some tenant scoped permission for user creation via SSO, 1 seems like it would regress that. 

I like option 2. 

But worth calling out: a tenant admin has permission to add a role to a user when they're creating them so onboarding a new member would still be fully delegated to a tenant admin. 

> Also would be possibly bad ux for the admin to accidentally remove the last role from a user and then not see the user any longer to un-do it.

Yea definitely wouldn't want that! If we required at least one role to exist on a user they wouldn't be able to remove the last role though so that would avoid the user disappearing scenario. 

We may want to consider implementing option 1 in another PR. But I'll just implement option 2 for now.

yuliadub:
> I like the direction but the one downside is that for the "add a new user to the team" flow - which may be the only time an admin interacts with users, realistically speaking - they need to see the candidates for whom they will add tenant-scoped roles, which are (usually) members of their domain and (always) users with no existing role assignments yet.

I am not sure I understand the downside  (maybe Monday brain lol) - are you saying if I add a user with no roles and then try to find them later to now add a role it would be harder to find them since the email domain may not match the tenant I want to add them to? (like I'd need to remember their name to search for them vs us being able to display by default?) Or are you referring to the possibility of assigning roles automatically based on email domain when users request accounts? 

Overall I agree that it is possible that we will have at least one center that will not use the same domain in their email as the short name of the center so figuring a graceful way to handle both scenarios (where domain matches and where it doesn't) is important. 

Just brainstorming: For when the users who are added, could we tag those users with who added them or something similar? And that way they can pop up even if a role did get removed based on the user who added them and their tenant? (not full proof but maybe a start of an idea that could work...?) or maybe assume that the user who added them tenant is the one this user can be associated with.  (this is along option 4 Kellen is suggesting)

Going with 2 currently makes sense to me.

busbyk:
> I am not sure I understand the downside (maybe Monday brain lol) - are you saying if I add a user with no roles and then try to find them later to now add a role it would be harder to find them since the email domain may not match the tenant I want to add them to? (like I'd need to remember their name to search for them vs us being able to display by default?) Or are you referring to the possibility of assigning roles automatically based on email domain when users request accounts?

The issue with the current approach is that if a user doesn't have any roles a tenant admin can't see them. To be more specific any role applied to a user with a users collection rule will not apply to a user without any roles (I feel like that's still confusing). 

So a user logs into the admin dashboard. This user has a role that allows them to create and read on the users collection. With the change I was introducing in this PR (prior to implementing option 2) that logged in user would be able to read all user collection documents that have roles with the same tenant(s) as the logged in user. 

But if the logged in user creates a user collection document but doesn't assign a role to them, they will save that user and then they will disappear because that new user doesn't have any roles with the same tenant. So that's the current issue that option 2 largely resolves. 

> Overall I agree that it is possible that we will have at least one center that will not use the same domain in their email as the short name of the center so figuring a graceful way to handle both scenarios (where domain matches and where it doesn't) is important.
> 
> Just brainstorming: For when the users who are added, could we tag those users with who added them or something similar? And that way they can pop up even if a role did get removed based on the user who added them and their tenant? (not full proof but maybe a start of an idea that could work...?) or maybe assume that the user who added them tenant is the one this user can be associated with. (this is along option 4 Kellen is suggesting)

Yea I think we could implement several of these options depending on what we want. 

Options 3 & 4 also make sense to me.

I think this would be good to include in a technical discussion for #230

stevekuznetsov:
> add a role to a user when they're creating them

When we're on SSO, I don't think anyone will actually create User objects, will they?

busbyk:
Option 2 implemented!

busbyk:
> > add a role to a user when they're creating them
> 
> When we're on SSO, I don't think anyone will actually create User objects, will they?

Just depends what we want to allow. When we implement SSO we'll likely need to restrict who can sign up based on email domain (potentially solving some of our domain concerns). 

But what if a center wants to hire a contractor to write blog posts for them (or something along those lines) and that contractor doesn't have/get an email with their domain? Do we build some invite flow? Or do we let an admin create a user for them and have them go through the password reset flow?

stevekuznetsov (approved):

PR #284, opened 2025-06-23
Breadcrumbs

### LINKED ISSUE #207: Breadcrumbs
Add a breadcrumbs component to all (all?) pages:

Image

busbyk:
Resolves #207

Uses useSelectedLayoutSegments hook plus a few hardcoded overrides to generate breadcrumbs. 

This hook returns nested folders without page or layout files as valid routes so I had to add those as `knownPathsWithoutPages` so those crumbs wouldn't be clickable (since we don't have pages at those routes). 

I also updated our vertical padding to be consistent for pages with breadcrumbs to match the Figma designs.


package.json
```diff
@@ -54,7 +54,7 @@
     "@radix-ui/react-navigation-menu": "^1.2.5",
     "@radix-ui/react-popover": "^1.1.10",
     "@radix-ui/react-select": "^2.1.4",
-    "@radix-ui/react-slot": "^1.1.1",
+    "@radix-ui/react-slot": "^1.2.0",
```
busbyk, line 57: From installing breadcrumbs from shadcn

src/components/Breadcrumbs/Breadcrumbs.client.tsx
```diff
@@ -0,0 +1,105 @@
...
+  BreadcrumbItem,
+  BreadcrumbLink,
+  BreadcrumbList,
+  BreadcrumbPage,
+  BreadcrumbSeparator,
+} from '../ui/breadcrumb'
+
+const knownPathsWithoutPages = ['/forecasts', '/weather', '/weather/stations']
+const knownPathTitleOverrides: Record<string, string> = {
+  '/posts': 'Blog',
```
stevekuznetsov: Can we just move the files/routing around?

src/components/Breadcrumbs/Breadcrumbs.client.tsx
```diff
@@ -0,0 +1,105 @@
...
+  name: string,
+  href: string | null,
+  isLast: boolean,
+): BreadcrumbType => ({
+  name: knownPathTitleOverrides[href || ''] ?? name.replace(/-/g, ' '),
+  href: href && knownPathsWithoutPages.includes(href) ? null : href,
+  isLast,
+})
+
+const processNestedSegments = (
```
stevekuznetsov, line 34: Excellent candidate for some unit tests

src/components/Breadcrumbs/Breadcrumbs.client.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return [...prependedSegments, mainSegment]
+}
+
+export function Breadcrumbs() {
+  const segments = useSelectedLayoutSegments()
+
+  if (segments.length === 0) return null
+
+  const breadcrumbItems: BreadcrumbType[] = segments.flatMap((segment, index) => {
```
stevekuznetsov, line 57: and/or unit test this flatMap

stevekuznetsov:
Looks great, just one question on the path title overrides.

src/components/Breadcrumbs/Breadcrumbs.client.tsx
busbyk: I would definitely be happy renaming /posts to /blog. @rchlfryn you're working on the blog now - any objections to renaming?

I think /posts is just leftover naming from the website template right?

stevekuznetsov: Yeah, I think it's just a leftover from there!

rchlfryn: Yes it is leftover from the template. Change it!

rchlfryn (approved):
LG pending test

yuliadub:
I know I likely am starting to sounds like a broken record but I noticed that in breadcrumbs, for example if you go to About us, you have "home->about-> about us", home is clickable and so is about us but about is not. That feels a bit broken to me, is it just me or anyone else have any thoughts about it? 

a possible proposed solution: each main nav that is not clickable can become clickable and all it would display would be the list of links to the pages present under it.

busbyk:
> I know I likely am starting to sounds like a broken record but I noticed that in breadcrumbs, for example if you go to About us, you have "home->about-> about us", home is clickable and so is about us but about is not. That feels a bit broken to me, is it just me or anyone else have any thoughts about it?
> 
> a possible proposed solution: each main nav that is not clickable can become clickable and all it would display would be the list of links to the pages present under it.

I think we should revisit the "should top level nav items be clickable" discussion. I think the behavior of the breadcrumbs in this PR matches the current behavior of the top level nav item _not_ being clickable. I think we should have the conversation under #288 or #3 

If we make a change to top level nav items being clickable the breadcrumbs would be updated to match that behavior. 

We'll still have some situations where some of the breadcrumbs are not clickable though in cases like the weather stations map which is currently nested under /weather/stations where weather and stations don't have pages/routes.

src/components/Breadcrumbs/Breadcrumbs.client.tsx
busbyk: @rchlfryn since you're changing things with the blog right now I'm going to make that change in a separate PR so I don't give you an annoying merge to do after renaming files haha. 

I'll just remove this title override and we can change the naming/routing of the blog in another PR.

src/components/Breadcrumbs/Breadcrumbs.client.tsx
busbyk: tests added

src/components/Breadcrumbs/Breadcrumbs.client.tsx
busbyk: component tests added

Issue #309, opened 2025-06-26
Cannot provide name when Biography linked to User


busbyk:
This is working as intended according to the code but this behavior was confusing to a user. 

![Image](https://github.com/user-attachments/assets/005a78f8-6589-4ce0-8d26-718dd384b6f8)

If a Biography doesn't have a name it can't be looked up in relationship type fields. We should either require a name field value or use some calculated field with the fallback as the id for lookups. Probably just makes sense to require the name field. 

If we want to keep the user relationship we could try some of the options in my comment below.

busbyk:
@rustynwac What do you think the behavior should be here? We could probably make the name field readOnly and set to the user's name if a user relationship is added. 

Or we could add a little helper text that says that if the name field is set this will be used instead of the user's name.

rchlfryn:
We will also want to figure out a way to only show users from the selected tenant. `filterOptions` will not work since a user is not assigned to a `tenant` so we will need to get creative about reading this through `roleAssignments`

busbyk:
> We will also want to figure out a way to only show users from the selected tenant. `filterOptions` will not work since a user is not assigned to a `tenant` so we will need to get creative about reading this through `roleAssignments`

Good call. We can adapt the logic from here for that: https://github.com/NWACus/web/blob/main/src/collections/Users/access/byGlobalRoleOrTenantRoleAssignmentOrDomain.ts

busbyk:
> [@rustynwac](https://github.com/rustynwac) What do you think the behavior should be here? We could probably make the name field readOnly and set to the user's name if a user relationship is added.
> 
> Or we could add a little helper text that says that if the name field is set this will be used instead of the user's name.

@rustynwac did either of these options sound good to you?

rustynwac:
Helper text seems easy to me. Not sure this is critical for Sierra launch though now that I understand it more.

busbyk:
> Helper text seems easy to me. Not sure this is critical for Sierra launch though now that I understand it more.

Agreed, it's an easy change but let's kick it out of the Sierra milestone to make sure we're only working on things that are necessary right now.

busbyk:
Related to #295

rchlfryn:
We decided to just remove this behavior

PR #312, opened 2025-06-28
Fix relations fields to be tenant exclusive on posts

### LINKED ISSUE #305: Change default column for Role Assignments
Under roles assignments - change column to show name instead of id

add to `admin` `useAsTitle: 'name',`

rchlfryn:
#### Description
This PR solves three issues I noticed when working on blogs.
* The first issue I noticed was as a super admin, when navigating to a post, if you try to select an author or related post, it would show all options from all tenants. This was especially confusing because we have the same title for post across multiple tenants. I think the simple fix was to add a filter on the fields.
* The second issue I noticed was the authors selected in the seed script for Sierra and sawtooth of were NWAC biographies. This is a bit more involved in touches the `populateAuthor` hook, which I am not familiar with, but it does properly select an author within the avalanche center.
* Finally this resolves #305


src/collections/Posts/hooks/populateAuthors.ts
rchlfryn, line 1: This change doesn't seem correct based on the comments, but I tested it and it work. I am not really sure what this was doing so someone feel free to explain.

stevekuznetsov: I think it just depends on what we want the outcome to be here - if we want authors on a post to be `User`, and we want it to display nicely, we need to have `User` records with avatars, etc. We know we can't associate every `Biography` with a `User` (not only do we have separate bios for the same user across multiple centers, but we also have emeritus/post-humous bios). We might also ask that every `User` have a `Biography` in the tenants they are writing for, which would let you search this.

src/collections/Posts/index.ts
```diff
@@ -130,6 +130,13 @@ export const Posts: CollectionConfig<'posts'> = {
       hasMany: true,
       relationTo: 'biographies',
       required: true,
+      filterOptions: ({ data }) => {
```
busbyk, line 133: It's weird that the baseListFilter on biographies doesn't apply here by default. But good to know!

src/collections/Posts/hooks/populateAuthors.ts
busbyk: Wait, I don't think we're even using populatedAuthors anywhere. It's only used by the https://github.com/NWACus/web/blob/main/src/utilities/formatAuthors.ts utility but that's not used anywhere. 

The Posts collection is already using the biographies collection for author relationships. 

We should remove populateAuthors and formatAuthors.

busbyk (changes requested):
> Finally this resolves #305 

Did you mention the correct issue here? This PR doesn't change anything related to the role assignments collection.

rchlfryn:
@busbyk `populateAuthors` is being run `afterRead` on posts
https://github.com/NWACus/web/blob/main/src/collections/Posts/index.ts#L214

I am fine removing it, I just want to know if we should or not. I will bring this up at stand today

busbyk:
> @busbyk `populateAuthors` is being run `afterRead` on posts https://github.com/NWACus/web/blob/main/src/collections/Posts/index.ts#L214
> 
> I am fine removing it, I just want to know if we should or not. I will bring this up at stand today

Yea that's there to allow for displaying user fields on the frontend since read access for the users collection is restricted. 

But we don't need that for two reasons:
1. We don't respect access control in our server components. We use the local API which overrides access by default. 
2. We're not using the users collection as authors anymore, we're using the biographies collection. #321 will make it so that biographies no longer has permissive read but since our server components use the local API and we don't use the REST API at all right now, this isn't an issue. populatedAuthors isn't used in the admin panel anywhere. 

Although I suppose there may be a little more to this if we want to use the user related to a biography for the name instead of the biography name https://github.com/NWACus/web/issues/309.

So worth considering that.

busbyk (approved):
🤘

rchlfryn:
I will merge this and we can remove `populateAuthors` as needed in #321

PR #316, opened 2025-07-07
Global roles separation + role assignment UI enhancements

### LINKED ISSUE #230: Role management enhancements

### LINKED ISSUE #305: Change default column for Role Assignments

busbyk:
Resolves #230
Resolves #305

- Separates roles used for tenant-scoped role assignments and global roles
	- Global role assignments use global role rules for access control
- Adjusts access control to allow for readable roles and role assignments in join field UIs
	- Role assignments access control is by tenant roles with the ability to read a user's own roles
	- Roles access control is by tenant roles with the ability for any authenticated user to read 
	- If a user only has read access to roles or role assignments they will not see the list view for those collections. This is to allow the display of role names and role assignments on a user's account page or in the users list view. 
- Adjusts role assignments to have a single role instead of multiple because we can use multiple role assignments for users with more than one role. This simplifies making the role name readable in the admin panel UI. 
- Renames byTenant to byTenantRole for clarity

Forecaster user:
CleanShot 2025-07-06 at 22 14 54@2x

Multi center admin (applies to any tenant admin):
CleanShot 2025-07-06 at 22 15 44@2x
CleanShot 2025-07-06 at 22 16 22@2x (The default columns exclude the avalanche center because for a non-admin user they would get an "Untitled - ID:{number}" value since they don't have read on that collection. It seems fine to have to add this column manually if needed for multi tenant users / super admins)
CleanShot 2025-07-06 at 22 18 16@2x


src/access/byGlobalRole.ts
```diff
@@ -1,11 +1,11 @@
...
 import { ruleCollection, ruleMatches, ruleMethod } from '@/utilities/rbac/ruleMatches'
-import type { Access, CollectionConfig } from 'payload'
+import type { CollectionConfig, FieldAccess } from 'payload'
 
 // byGlobalRole walks the global roles bound to the user to determine if they have permissions
 // to take the specified action on a resource of the collection type at the global scope
-export const byGlobalRole: (method: ruleMethod, collection: ruleCollection) => Access =
-  (method: ruleMethod, collection: ruleCollection): Access =>
+export const byGlobalRole: (method: ruleMethod, collection: ruleCollection) => FieldAccess =
+  (method: ruleMethod, collection: ruleCollection): FieldAccess =>
```
busbyk, line 8: I used this in a field level access control config too and field level access only accepts Boolean as a return value, not a query. This already only returned Boolean so I felt fine updating this type.

src/collections/Users/index.ts
```diff
@@ -20,21 +21,28 @@ export const Users: CollectionConfig = {
...
       saveToJWT: true,
       maxDepth: 3,
+      admin: {
+        defaultColumns: ['role'],
+      },
+    },
+    {
+      name: 'globalRoles',
+      type: 'relationship',
+      relationTo: 'globalRoles',
```
stevekuznetsov: Shouldn't this be a relationship to assignments? There's no direct link between users and roles.

busbyk: I removed the assignments collection / join table because there is no contextual data needed on the relationship between a user and a global role. The relationship between a user and a tenant scoped role needs contextual data on it (tenant) so it makes sense to use an explicit join table. 

I say an explicit join table because of [this advice in the docs](https://payloadcms.com/docs/fields/join#using-the-join-field-to-have-full-control-of-your-database-schema:~:text=For%20typical%20polymorphic,junction%20table%20design.) and because we could change this to be a join type, still on globalRoles, to enable bi-directional updates without our own assignments collection / join table. I suppose we might as well do that so we could see which users have Super Admin by visiting the global role rather than filtering the users collection by global role. 

I suppose that the argument for a global role assignments collection is the argument for flexibility. What if we want to have contextual data on the relationship between a user and a global role in the future? 

The choice to remove the explicit join table was just for simplicity. 

What are you thinking?

stevekuznetsov: Having an assignments collection allows these two to be decoupled, not necessarily for the fact that it allows some metadata. I don't feel strongly enough to push it, but if you're going to remove the concept of role assignments, let's do it both for global and tenant-scoped ones.

busbyk: We do need the assignments collection for tenant-scoped roles because we need a tenant associated with the relationship. So we need to keep that. 

What is the benefit of decoupling global roles and users by using a global role assignments collection? (genuine curiosity question)

stevekuznetsov: Part of it (and in this case likely immaterial since our ORM is doing something for us here) is having the correct normal form in the database with the join table. The other part is if the references to roles and users in the join table are not strictly enforced to be foreign keys, it turns the relationship into an eventually consistent one where you can, say, pre-allocate a number of role assignments for new users before they are done doing the signup flow and have their permissions work immediately when the user object ends up being created and they finish the signup. I don't know if that is easy to achieve in the ORM. If we require the assignment collection for tenant-scoped roles there's a small argument to keeping the same structure for global-scoped ones to not surprise users, but if you would prefer to not do that, it's ok.

busbyk: Thanks for explaining 🙏 

> The other part is if the references to roles and users in the join table are not strictly enforced to be foreign keys, it turns the relationship into an eventually consistent one where you can, say, pre-allocate a number of role assignments for new users before they are done doing the signup flow and have their permissions work immediately when the user object ends up being created and they finish the signup. I don't know if that is easy to achieve in the ORM. 

Ooh yea that's an interesting one. Yea I think that could be tricky with the way our ORM is set up to work with Payload but conceptually that makes sense to me. 

> If we require the assignment collection for tenant-scoped roles there's a small argument to keeping the same structure for global-scoped ones to not surprise users, but if you would prefer to not do that, it's ok.

I think the case for consistency is the strongest argument for me. I like the simplicity of just selecting one or many global roles directly on the user document for simplicity but I can get behind consistency. I'll update.

stevekuznetsov: I'll be honest, I have not worked with ORMs much and don't really trust a layer to be the thing deciding how things get laid out in the DB, but I can appreciate the simplicity, especially if you don't have performance needs. Likely why I ended up making so many different collections vs the single "settings" that you have invented - which is by far the nicer approach for how users will interact with the admin panel.

busbyk: > I'll be honest, I have not worked with ORMs much and don't really trust a layer to be the thing deciding how things get laid out in the DB, but I can appreciate the simplicity, especially if you don't have performance needs.

Yea that's pretty fair!

So do you think the argument for consistency is strong enough to justify adding the global role assignments collection back or should we just stick with the simplicity of globalRoles being a direct relationship on users?

stevekuznetsov: Personally, yes - but I defer to your choice here. I imagine having to teach the "roles" and "assignments" distinction is likely going to be a higher barrier for most and not having to remember two flows will be nice, but I don't know for sure.

yuliadub: having internalized the details of this (appreciate you running through this yesterday Kellen) and played with the UI, I think consistency is a strong enough argument for me to add the global role assignments collection back.

busbyk: Consistency it is! Added that back. Thanks y'all 🙏

yuliadub (approved):
a lot in here, thanks for getting this figured out!

PR #320, opened 2025-07-09
updating /posts/ routes to be /blog/ routes

busbyk:
I left a few things named with "posts" internally since that's still the collection name and I think it makes sense to leave it that way. I only changed naming related to routing.


rchlfryn (approved):
👍

PR #321, opened 2025-07-10
Remove permissive read

### LINKED ISSUE #294: baseListFilter not working as expected for teams and bios
This is potentially an issue with our permissive read access control functions. 

When viewing Teams and Biographies users can see all tenants when they should only see tenants they have tenant-scoped roles for. 

Needs further investigation.

busbyk:
Resolves #294 

Removing permissive reads means that we don't need to change the filterByTenant baseListFilter since we're now letting roles dictate read access control which kicks in when there isn't a tenant cookie set to filter list results. 

I also: 
- added access control via byTenantRole to the redirects and form builder plugins
- removed the unused payload cloud plugin that we don't intend to use

### DIFF
--- a/src/access/filterByTenant.ts
+++ b/src/access/filterByTenant.ts
@@ -1,4 +1,4 @@
-import { getTenantListFilter } from '@payloadcms/plugin-multi-tenant/utilities'
+import { getTenantFromCookie } from '@payloadcms/plugin-multi-tenant/utilities'
 import { BaseListFilter } from 'payload'
 
 // filterByTenant implements per-tenant data filtering from the 'payload-tenant' cookie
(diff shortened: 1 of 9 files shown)


src/access/filterByTenant.ts
```diff
@@ -1,10 +1,20 @@
...
+
+  if (selectedTenant) {
+    return {
+      tenants: {
+        equals: selectedTenant,
+      },
+    }
+  }
+
+  return null
```
busbyk, line 19: This is from https://github.com/payloadcms/payload/blob/main/packages/plugin-multi-tenant/src/list-filters/filterDocumentsBySelectedTenant.ts

yuliadub (approved):
this makes sense to me!

Issue #243, opened 2025-06-04
Add template tenant


rustynwac:
- Create a "template" tenant and use their Pages collection and Navigation global collection as the source for generated pages.
- We need to use this approach because many of our blocks expect related collections to exist. We get previews without extra routes. We get shareable links (template.avy-fx.org or something similar).
- This tenant's site would actually be publicly accessible at template.avy-fx.org (or some similar subdomain) but we would never enable a custom domain for it. We think having this URL publicly accessible isn't a problem and is actually a nice way to share examples or demo template page designs. Might even be a nice way to promote AvyWeb
- Just need to filter out this "special" tenant from some locations like the root domain's avalanche center listings.
- Widget's will be broken for this tenant so we might want to add some fallback UI if widget loading fails (which might be nice anyways).

rchlfryn:
@rustynwac this should be easy enough. What other pages did you want to add?

rustynwac:
There's quite a few marked in the spreadsheet @rchlfryn. We can add more if those are really easy.

rchlfryn:
Done in [Add template tenant #322](https://github.com/NWACus/web/pull/322)

PR #322, opened 2025-07-11
Add template tenant

rchlfryn:
### Description
There is a request to help avalanche centers build there pages. The way we decided to do this is build a "template tenant" that will not be seen by anyone but super admins, create the pages under that tenant, and duplicate any pages to the respective avalance center that wants the page. This PR:
* Adds a new tenant Death Valley Avalance Center (dvac)
* Adds a `Duplicate page for...` button in the edit view of pages

We still need to alter permissions, but that will be a follow up PR with #316 out in the wild at the same time.
We also need to build the pages which will be a separate PR

### Issue
Addressed part of of #243


src/app/(payload)/custom.scss
```diff
@@ -6,3 +6,7 @@
   width: 24px;
   height: 24px;
 }
+
+#duplicate-page-drawer .drawer__content {
+  max-width: 600px;
+}
```
busbyk, line 12: I have no issues with this class but just FYI we can use tailwind classes in Payload because we are importing 
```
@tailwind components;
@tailwind utilities;
```
in this file

src/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
```diff
@@ -0,0 +1,157 @@
...
+  useFormModified,
+  useModal,
+  useRouteTransition,
+} from '@payloadcms/ui'
+import { useRouter } from 'next/navigation'
+import { formatAdminURL } from 'payload/shared'
+import { useCallback, useState } from 'react'
+
+// TODOs
+// - Remove photos from blocks or use a global photo?
```
busbyk: Ah because photos are tenant scoped right. Hadn't thought about that. 

I think using a global, fallback photo might be a nice UX. Might be worth calling this out in some helper text in the duplicate to drawer.

src/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
```diff
@@ -0,0 +1,157 @@
...
+        <PopupList.Button onClick={() => openModal(drawerSlug)}>
+          Duplicate page to...
+        </PopupList.Button>
+      </PopupList.ButtonGroup>
+
+      <Drawer slug={drawerSlug} title="Duplicate page to..." gutter={true}>
+        <div className="p-6">
+          {modified && (
+            <Banner>
+              <b>Warning:</b> this page has been modified Continuing with duplication will not apply
```
busbyk: I think you're just missing a period here:

```suggestion
              <b>Warning:</b> this page has been modified. Continuing with duplication will not apply
```

src/components/DuplicatePageFor/index.tsx
```diff
@@ -0,0 +1,8 @@
+import { EditMenuItemsServerProps } from 'payload'
+import { DuplicatePageForDrawer } from './DuplicatePageForDrawer'
+
+export const DuplicatePageFor = ({ user }: EditMenuItemsServerProps) => {
+  const isSuperAdmin = (user?.globalRoles?.docs?.length ?? 0) > 0
```
busbyk: This doesn't necessarily mean that they are a super admin. I think you'd want to use `byGlobalRole('*', '*')` for that.

src/components/DuplicatePageFor/index.tsx
```diff
@@ -0,0 +1,8 @@
...
+  if (!isSuperAdmin) return null
+  return <DuplicatePageForDrawer />
```
busbyk: Should users who have create access on the pages collection for multiple avalanche centers (via role assignments) be able to duplicate pages across their tenants? Or any other scenarios?

src/components/ui/button.tsx
```diff
@@ -25,7 +25,7 @@ const buttonVariants = cva(
         link: 'text-primary items-start justify-start underline-offset-4 hover:underline',
         outline: 'border border-border bg-background hover:bg-card hover:text-accent-foreground',
         secondary: 'bg-secondary text-secondary-foreground hover:bg-secondary/80',
-        callout: 'bg-callout text-callout-foreground hover:bg-callout/90',
+        callout: 'bg-callout text-callout-foreground hover:bg-callout-hover',
```
busbyk: If we want to keep the same functionality as before I think we'd need our :root `--callout-hover` to be a slightly different color (i.e. 90% opacity I believe) of the `--callout` but in globals.css you set it to be the same color as `--callout`. 

I realize that you wanted to have a different callout hover color for dvac and I'm onboard with breaking these out.

src/endpoints/bootstrap/index.ts
```diff
@@ -49,6 +49,11 @@ export const bootstrap = async ({
   payload.logger.info(`— Seeding tenants...`)
 
   const tenantData: RequiredDataFromCollectionSlug<'tenants'>[] = [
+    {
+      name: 'Death Valley Avalanche Center',
```
busbyk, line 58: I love this

src/services/nac/nac.ts
busbyk, line 1: Instead of `fallbackCenter` and `replaceCenter` what about something like `safeCenter`? 

fallback and replace are a little confusing. safe isn't _super_ obvious either but 🤷

busbyk (changes requested):
I like it! 

We talked about some UI for duplicating all off the template tenant's (i.e. dvac) pages to a newly created tenant. I assume this is a precursor to that and that will be in another PR?

src/app/(payload)/custom.scss
rchlfryn: Unfortunately I can't get access to `.drawer__content` since it is a payload component, and I can't pass it, so it needs to be css

src/components/DuplicatePageFor/index.tsx
rchlfryn: Thanks for sharing. I think I need to have a list of all secret payload functions because 9 times out of 10 they are what I am looking for. I also agreed this would not work but thought it was "fine for now". 
I did try using this but it was still visible for tenant admins in the UI. Is `byGlobalRole` able to be called in a component? I think this should work after reading the code. I will commit and see if I am doing something wrong?

src/components/DuplicatePageFor/index.tsx
rchlfryn: Good shout... we can add a check for that as well.

rchlfryn: I added a check, but waiting on the comment above to be able to test

src/services/nac/nac.ts
rchlfryn: I went with `actualAvyCenter` 🤷

src/components/ui/button.tsx
rchlfryn: I undid this.

I initially added it to debug why the callout color was not working on hover, and it was because I left `,` in between the hsl colors when picking a color.

src/components/DuplicatePageFor/index.tsx
stevekuznetsov: `byGlobalRole` is our invention - so you can check it out in the repo and de-mystify!

src/app/(payload)/custom.scss
busbyk: Ahhh I see, makes sense!

src/components/DuplicatePageFor/index.tsx
busbyk: Ah yea byGlobalRole uses currying so you need to call the result of byGlobalRole with a request object. Sorry I didn't include the full implementation in my original comment. There are a couple examples in #323 too but:

```
  const mockedPayloadReq = {
    user,
    payload,
  } as PayloadRequest

  const isSuperAdmin = byGlobalRole('*', '*')({ req: mockedPayloadReq })
```

src/components/ui/button.tsx
busbyk: Cool 👍

src/services/nac/nac.ts
busbyk: I like it!

src/components/DuplicatePageFor/index.tsx
busbyk: Cool. I think you'd probably want to do something similar this roles check in `byTenantRole` https://github.com/NWACus/web/blob/main/src/access/byTenantRole.ts#L31

This works:

```
  const roleAssignments = roleAssignmentsForUser(payload.logger, user)
  const matchingTenantIds = roleAssignments
    .filter(
      (assignment) =>
        assignment.role &&
        typeof assignment.role !== 'number' && // captured in the getter
        assignment.role.rules.some(ruleMatches('create', 'pages')),
    )
    .map((assignment) => assignment.tenant)
    .filter((tenant) => typeof tenant !== 'number') // captured in the getter
    .map((tenant) => tenant.id)
  const canCreatePagesInMultipleTenants = matchingTenantIds.length > 1
```

I think the create action on the pages collection is what we'd want to check for this right?

src/components/ui/button.tsx
busbyk: You left `--callout-hover` in globals.css but it's not used anywhere. If we're not using it anywhere I'd lean towards removing it 🤷

src/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
```diff
@@ -0,0 +1,157 @@
...
+} from '@payloadcms/ui'
+import { useRouter } from 'next/navigation'
+import { formatAdminURL } from 'payload/shared'
+import { useCallback, useState } from 'react'
+
+// TODOs
+// - Remove photos from blocks or use a global photo?
+
+export const DuplicatePageForDrawer = () => {
+  const { initialData: pageData } = useDocumentInfo()
```
busbyk: I ran into a bug where if I made changes to a document, published them, and then duplicated the page the newly added (and published) page data was omitted.

When I changed `initialData` to `savedDocumentData` the newly added and published data was duplicated.

src/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
```diff
@@ -0,0 +1,157 @@
...
+        const { slug, layout, title } = cleaned
+
+        const now = new Date()
+        const timestamp = now.valueOf()
+
+        const newPage = {
+          layout,
+          title,
+          tenant,
+          slug: `${slug.replace(/^\/([^/]+)\//, `/${tenant.slug}/`)}-${timestamp}`,
```
busbyk: When I duplicate a page to a tenant the slug is generated with the timestamp suffix. When viewing it in the pages list view I see that timestamped slug but in the edit view it shows as the slug determined from the `fieldToUse` which defaults to `'title'` in https://github.com/NWACus/web/blob/template-tenant/src/fields/slug/SlugComponent.tsx
CleanShot 2025-07-16 at 08 59 51

Then saving this results in a conflict:
CleanShot 2025-07-16 at 09 00 40@2x

I guess I have a few thoughts here:
1. Whoa our slug field should probably allow us to modify the slug to be different than a slug determined from the configured `fieldToUse` which is title for the pages collection. I think the behavior here is a bug with the slug field separate from this PR but does impact the functionality here. 
2. Rather than generating a slug with a suffixed timestamp it would be nice to check if the slug is already taken and then append an incrementing number. I know that's more effort than just a guaranteed unique slug via the timestamp method. It would be nice to do this in a server action to be able to use the local API.

src/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
```diff
@@ -0,0 +1,157 @@
...
+          Duplicate page to...
+        </PopupList.Button>
+      </PopupList.ButtonGroup>
+
+      <Drawer slug={drawerSlug} title="Duplicate page to..." gutter={true}>
+        <div className="p-6">
+          {modified && (
+            <Banner>
+              <b>Warning:</b> this page has been modified. Continuing with duplication will not
+              apply changes.
```
busbyk: What about: 

```suggestion
              <b>Warning:</b> this page has been modified. Please save or revert your changes before duplicating this page.
```

src/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
rchlfryn: Great catch - I will update this.

src/components/DuplicatePageFor/index.tsx
rchlfryn: Thanks for teaching. This was neat to dig into. I don't know how I haven't run into this before.

src/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
rchlfryn: @busbyk I took the shortcut but agree adding a sequential/incremental number is so much better. I added a new endpoint to do all the heavy lifting for us in [d554444](https://github.com/NWACus/web/pull/322/commits/d55444475808505639a1da820d58377f881472cc). 

I also side stepped the "slug bug" in payload where it is rendered based on the title since I added the sequential number to the title. This seemed like the cleanest approach.

src/endpoints/duplicatePageToTenant.ts
```diff
@@ -0,0 +1,64 @@
...
+      tenant: {
+        equals: tenantId,
+      },
+    },
+  })
+  if (pageResults.totalDocs < 1) {
+    return `${i}`
+  }
+  return getUniqueSlug(baseSlug, tenantId, i + 1)
+}
```
busbyk: I like this update. But it this will always append an integer on the title and slug even if there is no matching page slug. I.e. "Test Page" will be copied as "Test Page 1" even if there is no "Test Page" in the destination tenant. 

What about:

```suggestion
export async function duplicatePageToTenant(req: PayloadRequest) {
  const selectedTenantId = req.routeParams?.selectedTenantId as number

  const { newPage } = await req.json?.()
  const newPageSansIds = removeIdKey(newPage)

  const uniqueSlugResult = await getUniqueSlug(newPageSansIds.slug, selectedTenantId)

  const payload = await getPayload({ config: configPromise })
  const tenant = await payload
    .find({ collection: 'tenants', where: { id: { equals: selectedTenantId } } })
    .then((res) => res.docs[0])

  return await payload.create({
    collection: 'pages',
    data: {
      ...newPageSansIds,
      tenant,
      title: uniqueSlugResult.index ? `${newPage.title} ${uniqueSlugResult.index}` : newPage.title,
      slug: uniqueSlugResult.slug,
    },
  })
}

async function getUniqueSlug(baseSlug: string, tenantId: number, i = 1): Promise<{ slug: string; index: number | null }> {
  const slug = i === 1 ? baseSlug : `${baseSlug}-${i}`

  const payload = await getPayload({ config: configPromise })
  const pageResults = await payload.find({
    collection: 'pages',
    depth: 0,
    where: {
      slug: {
        equals: slug,
      },
      tenant: {
        equals: tenantId,
      },
    },
  })
  if (pageResults.totalDocs < 1) {
    return { slug, index: i === 1 ? null : i }
  }
  return getUniqueSlug(baseSlug, tenantId, i + 1)
}
```

src/app/(frontend)/globals.css
```diff
@@ -50,6 +50,7 @@
     --footer-foreground-highlight: 240 3.8% 46.1%;
 
     --callout: 240 4.8% 95.9%;
+    --callout-hover: 240 4.8% 95.9%;
```
busbyk: Do you want to leave this in if we're not using it?

busbyk (approved):
I like the changes! Breaking this out into it's own endpoint will probably be helpful for future the script to copy all pages from the template tenant to a new tenant during onboarding. 

Just one suggested change and a thought:

I've been colocating things (components, hooks, server actions, etc.) within the collection's folder if they are specific to a collection. What do you think about that? Do you like that approach? 

It might be nice to colocate the DuplicatePageFor components and the duplicatePageToTenant handler inside of the pages collection for consistency if you think that makes sense. 

i.e. `./src/collections/Pages/component/DuplicatePageFor` and `./src/collections/Pages/endpoints/duplicatePageToTenant`

Looks good pending the incrementing slug change (feel free to take a different approach than my suggestion of course but I do think we should handle copying the title and slug without an integer suffix if there is no conflict).

src/app/(frontend)/globals.css
rchlfryn: No I do not. Removing this!

src/endpoints/duplicatePageToTenant.ts
rchlfryn: Ah yes --- good call. I was initially returning the slug from `getUniqueSlug` (hence the name) but moved to just returning the number to be able to add it to the title as well. I might add the title to this function as well. I will commit this and might amend a few things like the name and possibly passing the title.

rchlfryn:
> I've been colocating things (components, hooks, server actions, etc.) within the collection's folder if they are specific to a collection. What do you think about that? Do you like that approach?

> It might be nice to colocate the DuplicatePageFor components and the duplicatePageToTenant handler inside of the pages collection for consistency if you think that makes sense.

I like this approach better as well. If we do end up using it for posts or something else, we can unnest it from Pages.

Issue #313, opened 2025-07-03
Role assignments escalation check


busbyk:
For both global roles and roles (i.e. tenant scoped roles) add a check for their respective assignments collection to ensure that the user assigning a role has the permissions of the role being assigned. 

So if the logged in user is going to assign a user (could be themselves) a role with update on the biographies collection but the logged in user doesn't have update on biographies, they won't be allowed to assign that role.

busbyk:
Make sure the escalation check is also respected here in the user invite flow. 

Reference comment: https://github.com/NWACus/web/pull/323#discussion_r2208922776

PR #323, opened 2025-07-15
User invite flow

### LINKED ISSUE #289: Email adapter
We will inevitably want to send transactional emails: https://payloadcms.com/docs/email/overview

At the very least this will be helpful for user management flows like password resets. 

I'm personally a fan of Resend in terms of ease of use, setup and pricing.

### LINKED ISSUE #318: Provide avalanche center users with a good login flow
Pre standup on 7/3/25, we decided to move forward with Payload's user management. We need to ensure that the login, invite, password, password resets experience is seamless.

busbyk:
Resolves #289 
Resolves #318 

#### User invite flow
This PR introduces a workflow for inviting new users to AvyFx with preset roles (tenant scoped and global) that uses a very similar mechanism to Payload's forgot password / reset password flow. 
- Invite-based user creation. The Payload create operation for the edit view is disabled because that create button cannot be replaced without replacing the entire edit view. Create permission is used to control access to inviting. The invite operation essentially replaces user creation. 
- Inviting a user creates an inviteToken and inviteExpiration (defaults to 10 day expiration)
- A button to resend an invitation is available when a user's invite token expires
- A user status UI field was added which is either: accepted, invited, or invite expired
- A custom view for accepting invites was added at /admin/accept-invite which requires a token query param

#### Email integration
Email was required for this so this PR wires up both local and production email adapters: nodemailer in local env and Resend.com in production. react-email is used for custom email templates. See README for details on that setup. 
- custom invite user email
- custom forgot password email
- email previews
- instructions for using mailtrap.io for local email testing
- customized email components for future email development: EmailLayout, EmailLogo, EmailButton

Demo / overview: https://www.loom.com/share/2144da5d0ce84c238b32d57ae7f51b94?sid=6809bacb-f2ed-49ab-884b-adacd82d241f


package.json
```diff
@@ -74,18 +81,20 @@
     "pluralize": "^8.0.0",
     "prism-react-renderer": "^2.4.1",
     "qs-esm": "^7.0.2",
-    "react": "^19.0.0",
-    "react-dom": "^19.0.0",
+    "react": "^19.0.1",
+    "react-dom": "^19.0.1",
```
busbyk, line 88: This is for compatibility with react-email

package.json
```diff
@@ -108,6 +117,7 @@
     "postcss": "^8.4.49",
     "prettier": "^3.4.2",
     "prettier-plugin-organize-imports": "^4.1.0",
+    "react-email": "4.0.8",
```
busbyk, line 124: The latest version of react-email has a bug in the preview server so had to use an older version (latest is 4.1.3). Issue: https://github.com/resend/react-email/issues/2276

src/collections/Users/components/inviteUserAction.ts
```diff
@@ -0,0 +1,228 @@
...
+                role: assignment.roleId,
+                tenant: assignment.tenantId,
+              },
+            })
+            createdRoleAssignmentIds.push(roleAssignment.id)
+          }
+        }
+      }
+
+      const canCreateGlobalRoleAssignments = !!byGlobalRole(
```
stevekuznetsov, line 130: This is also susceptible to escalation, right?

stevekuznetsov (approved):
Nice. Reviewed the data flows, looks good!

src/collections/Users/components/inviteUserAction.ts
busbyk: Yea definitely. There's nothing stopping a user from creating a user with a higher level of permission at the moment. 

I'll leave a note on #313 to make sure to handle escalation here as well. Good callout.

src/views/AcceptInvite/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  if (user) {
+    return (
+      <MinimalTemplate className={`${baseClass}`}>
+        <div>
+          <div className="form-header">
+            <h1>Already Logged In</h1>
+            <p>You are already logged in.</p>
+          </div>
+          <Button buttonStyle="secondary" el="link" size="large" to={adminRoute}>
```
rchlfryn, line 87: ⛏️ Might be nice to add a logout button here

src/views/AcceptInvite/index.scss
```diff
@@ -0,0 +1,16 @@
+.accept-invite {
```
rchlfryn, line 1: ⛏️  Can you flatten this to be  `.accept-invite__brand` or nest  the css below within `.accept-invite` to be something like:
```scss
accept-invite {
  &__brand { 
  // Styles
  }
  
  &__wrap {
    .inputWrap {
    // Styles
    }
  }
}
```

src/views/AcceptInvite/AcceptInviteForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+      initialState={{
+        ...staticInitialState,
+        email: {
+          initialValue: email,
+          valid: true,
+          value: email,
+        },
+        name: {
+          initialValue: name,
+          valid: name ? true : false,
```
rchlfryn: Curious why you have the `true/false` here. what happened if you just left `name`?

src/collections/Users/index.ts
```diff
@@ -46,6 +63,25 @@ export const Users: CollectionConfig = {
...
+      type: 'text',
+      hidden: true,
+    },
+    {
+      name: 'inviteExpiration',
+      type: 'date',
+      hidden: true,
+    },
+    {
+      name: 'userStatus',
```
rchlfryn, line 79: Can you confirm we dont see this on create first user sign up?

src/collections/Users/components/inviteUserAction.ts
```diff
@@ -0,0 +1,228 @@
+'use server'
```
rchlfryn, line 1: Is this necessary? I thought use server was default unless use client was used

src/collections/Users/components/resendInviteActions.ts
```diff
@@ -0,0 +1,176 @@
+'use server'
```
rchlfryn, line 1: Same comment as above

rchlfryn (commented):
Some light comments

src/views/AcceptInvite/AcceptInviteForm.tsx
stevekuznetsov: Or `!!name`?

src/collections/Users/components/inviteUserAction.ts
stevekuznetsov: Can't hurt to be explicit

src/views/AcceptInvite/index.scss
busbyk: Yea sure, updated ✅

src/views/AcceptInvite/index.tsx
busbyk: I like it. Added ✅

src/views/AcceptInvite/AcceptInviteForm.tsx
busbyk: I simplified to `!!name`

src/collections/Users/components/inviteUserAction.ts
busbyk: `'use server'` is required in server actions because they are imported in client components so Next.js needs this directive to indicate that this is actually server code that shouldn't be included in the client bundle. 

I also thought this was weird at first.

src/collections/Users/index.ts
busbyk: Confirmed!

This field is a UI field so it's not actually stored in the database. But based on the lastLogin bug, good callout haha.

.env.example
```diff
@@ -24,3 +24,21 @@ NAC_HOST=
 
 # AFP wordpress API base url. Defaults to https://forecasts.avalanche.org. Not required.
 AFP_HOST=
+
+# The default from address for transactional emails. Defaults to support@avy-fx.org. Not required.
```
yuliadub, line 28: what happens if someone replies to the default address?

busbyk: Right now, nothing. The email would bounce. 

We would be able to set up receiving for this email in the future though without messing with sending via Resend. 

I do think we should set up a shared inbox for a support email. 

I'll get a discussion going in Slack about this.

src/app/(payload)/api/users/accept-invite/route.ts
```diff
@@ -0,0 +1,103 @@
...
+    // parse the request in the way the @payloadcms/ui Form component expects it to be
+    const req = await createPayloadRequest({
+      canSetHeaders: true,
+      config,
+      request,
+    })
+    await addDataAndFileToRequest(req)
+
+    const token = req.data?.token
+    const password = req.data?.password
```
yuliadub: looks like right now you can make a very weak password like "test" - I feel strongly that we should prevent passwords like that and at least have some basic reqs in place. Took a quick look if payload has a built in something to do this but didnt see one, weird!

yuliadub:
> * A button to resend an invitation is available when a user's invite token expires

would it be possible to resend if the token is not expired. People loosing emails is more common than it should be 😄 

When I fill in my name and password from accepting invite, the first time I try to submit, I get an "unexpected error occurred". If I push the submit button again, it works. Maybe it's just me? I tried twice and had the same experience. 

random thoughts, dont feel strongly about these: 
- if I accept the invite and make an account but then click the invite link again, it says it has expired and request a new one. Someone might get confused and ask for it again even thought they made an account, maybe we say "you already made an account" instead of generic "has expired"?
- looks like when I logged in my gmail that has a photo, the system adapted it. do we want that?
image
-  Before navigating away from the accept invite page after I fill it out and push submit, it brings back the original name vs what I entered. Very unlikely case since people probably wont be changing their name but could cause confusion for some.

src/app/(payload)/api/users/accept-invite/route.ts
busbyk: Ooh very good call. We can easily add this using a hook. Not a perfect solution because client side validation would be really nice for this but server side validation should be good enough for now. I'll add this. 

I'll probably use an environment flag so we can still use weak passwords in our local envs.

busbyk:
> would it be possible to resend if the token is not expired. People loosing emails is more common than it should be 😄

Yea for sure

> When I fill in my name and password from accepting invite, the first time I try to submit, I get an "unexpected error occurred". If I push the submit button again, it works. Maybe it's just me? I tried twice and had the same experience.

Oooh! You were able to recreate it. That's awesome. Did you notice any logs in the console? I haven't been able to recreate this on my machine. I'm seeing 400 errors in the BetterStack logs but the actual validation error isn't logged -_-

I'll see about logging the actual validation error, at least temporarily. 

> * if I accept the invite and make an account but then click the invite link again, it says it has expired and request a new one. Someone might get confused and ask for it again even thought they made an account, maybe we say "you already made an account" instead of generic "has expired"?

Yea I think that would be nice. Just need to make two queries instead of one. Happy to update. 

> * looks like when I logged in my gmail that has a photo, the system adapted it. do we want that?

Payload's [default](https://payloadcms.com/docs/admin/overview#the-admin-user-collection:~:text=Set%20account%20profile%20picture.%20Options%3A%20gravatar%2C%20default%20or%20a%20custom%20React%20component.) is to use gravatar for profile photos. We'd need to add an image upload to the users collection if we wanted users to upload their own. I think it's fine to use gravatar - I typically try to prefill user's avatars using gravatar in applications I build but I'd be fine with setting this to "default" to remove this instead. 

> Before navigating away from the accept invite page after I fill it out and push submit, it brings back the original name vs what I entered. Very unlikely case since people probably wont be changing their name but could cause confusion for some.

Hmm yea I do see that. Likely a state initialization thing - maybe Payload's Form component resets the form before waiting for success. I'll look into it.

busbyk:
@yuliadub updates made. 

Only thing I didn't do was 

> I'll see about logging the actual validation error, at least temporarily.
> 
> > * if I accept the invite and make an account but then click the invite link again, it says it has expired and request a new one. Someone might get confused and ask for it again even thought they made an account, maybe we say "you already made an account" instead of generic "has expired"?

I would have had to store the inviteToken that a user did finally accept with in order to still be able to look up a user by their token. I don't think that's worth it. Could modify the language though?

FYI If a user is logged in they will get a message indicating that they're already logged in with a link to the dashboard. 


I also refactored the form to attempt to resolve the stale state during submission and weird 400 error on first submit error. I discovered that there was a re-render happening a few seconds after initial render (could not for the life of me figure out what was causing it) and after that re-render the token would not be included in the request. 

Since I couldn't figure out what was causing it I refactored to use the same approach I used in the invite form - a server action and that seemed to resolve the issue but I'd love if you can try to recreate since I wasn't able to recreate in preview in the first place.

yuliadub:
> I would have had to store the inviteToken that a user did finally accept with in order to still be able to look up a user by their token. I don't think that's worth it. Could modify the language though?

just modifying the language sounds good! We can say something "its invalid due to it being expired or account has already been created. I agree not worth refactoring just to support this

> I also refactored the form to attempt to resolve the stale state during submission and weird 400 error on first submit error. I discovered that there was a re-render happening a few seconds after initial render (could not for the life of me figure out what was causing it) and after that re-render the token would not be included in the request.
> 
> Since I couldn't figure out what was causing it I refactored to use the same approach I used in the invite form - a server action and that seemed to resolve the issue but I'd love if you can try to recreate since I wasn't able to recreate in preview in the first place.

I will try to re-create today and let you know if it happens again or not!

src/collections/Users/components/inviteUserAction.ts
rchlfryn: I just learned this yesterday too. Cool cool

yuliadub (approved):
looks good to me! I see a re-render happening (where accepting goes back to "accept invite" briefly before it redirects me to be the logged in user) but no errors or anything on my end this time around.

PR #324, opened 2025-07-15
Last login

busbyk:
Sets a user's lastLogin field when they log in.

This was a request from Rusty.


stevekuznetsov:
Implementation looks fine but not really clear what the intent of this field is.

busbyk:
> Implementation looks fine but not really clear what the intent of this field is.

Just records when the user's last login was. @rustynwac had interest in this and it was really quick to add.

Issue #57, opened 2025-03-15
Integrate Sentry, PostHog, Log drain


stevekuznetsov:
We need to integrate our Sentry and PostHog accounts that we use for the app into this site.

- [x] Sentry
- [x] PostHog
- [x] Log drain (provider to be determined)

busbyk:
Whoever works on this task should make sure that Sentry captures server side exceptions successfully. Example of us not getting any logs for a server side exception: https://docs.google.com/document/d/1VbgMbk_SKMAOQoKGSE7wzUQwAzpxF1JX2sKRZem7fxs/edit?disco=AAABkB0FMMk

busbyk:
Let's also add a log drain with this issue.

PR #325, opened 2025-07-16
Sentry integration

busbyk:
Partially addresses #57 

Next.js Sentry integration + [Payload Sentry plugin](https://payloadcms.com/docs/plugins/sentry)


rchlfryn (approved):
Looks good! Just confirming we want this both on the client and server side? I cant think of any reason not to.

busbyk:
> Looks good! Just confirming we want this both on the client and server side? I cant think of any reason not to.

Yep, confirmed! Thanks 🙏

PR #326, opened 2025-07-16
PostHog integration

### LINKED ISSUE #227: Integrate Avy analytics
Eventually we'll want to add Avy insights to the dashboard

busbyk:
Resolves #227 

PostHog integration with:
- tenant property on auto capture
- custom capture function with tenant property included
- identification of admin panel user on login


busbyk:
~~Visiting the root page (i.e. just `/`) results in an endless loop of calls to /ingest and errors being thrown because that gets matched to the [center] dynamic param. #333 should resolve this.~~

#333 fixed this

src/providers/PostHogProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+import posthog from 'posthog-js'
+import { PostHogProvider as PHProvider } from 'posthog-js/react'
+import React, { useEffect } from 'react'
+
+export function PostHogProvider({ children }: { children: React.ReactNode }) {
+  useEffect(function initPostHog() {
+    if (process.env.NEXT_PUBLIC_POSTHOG_KEY) {
+      posthog.init(process.env.NEXT_PUBLIC_POSTHOG_KEY, {
+        api_host: `${getRootDomainURL()}/ingest`,
+        ui_host: 'https://us.posthog.com',
```
yuliadub: any specific reason to not use NEXT_PUBLIC_POSTHOG_HOST here?

src/providers/PostHogProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+        defaults: '2025-05-24',
```
yuliadub, line 15: not sure how often or if this changes but may make sense to pull into the config as well

busbyk: I think we could not ever change this. As far as I understand, the [defaults](https://posthog.com/docs/libraries/js/config#:~:text=defaults%0A%0AType%3A,most%20recent%20defaults.) argument just specifies a set of defaults for this config. So if we're happy with the default values for options right now, we could just leave this value as is indefinitely. And if we ever need to change a value in the config we would just set that explicitly. 

This was set by the PostHog setup wizard and is the recommended value now. I'm assuming this matches the current docs' specified default values for fields.

src/providers/PostHogProvider.tsx
busbyk: Nope! Updated.

rchlfryn (approved):

PR #328, opened 2025-07-16
hiding readOnly lastLogin field during create operations

### LINKED ISSUE #327: Remove last time login field
When you navigate to `admin/create-first-user`, you are able to see `Last login` field. We should hide this. I am guessing this is a side effect of #324 

Image

busbyk:
Resolves #327 

Their create first user view explicitly sets readOnly to false in their RenderFields function which receives all fields for the users collection so it was showing up during create operations. I added a condition that hides the field if the operation is create.


yuliadub (approved):
this makes sense 👍

PR #329, opened 2025-07-16
Adding permissive read back for the media collection

busbyk:
Well, that was a silly mistake on my part. Removing permissive read for the media collection results in the frontend not being able to load images because they are client-side requests. 

#321 introduced this.


yuliadub (approved):
good catch. I was trying to think about something like this when I was looking at the previous PR but did not connect the dots

busbyk:
> good catch. I was trying to think about something like this when I was looking at the previous PR but did not connect the dots

clearly I didn't either 🤦 haha

PR #333, opened 2025-07-17
setting only routes generated by generateStaticParams to match

### LINKED ISSUE #297: Non-center paths are matched by the [center] dynamic route resulting in errors
Paths that don't match any other route will get matched by the [center] dynamic route. Since the path values are not centers they cause errors when we query the NAC and AFP APIs (and search for a tenant). I added a tenant check so we don't call the NAC and AFP APIs with invalid center values but this is still problematic. 

Setting [dynamicParams](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamicparams) to `false` will help with this so non-center paths will 404.

### LINKED ISSUE #298: Non-tenant scoped layout does not have a custom 404 page
Just add a 404 page for non-tenant scoped routes (i.e. above the [center] dynamic route).

busbyk:
Resolves #297
Resolves #298 

Setting [dynamicParams to false](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamicparams) results in paths not generated by generateStaticParams to 404 (as expected). 

This also adds a global not-found via a workaround required for apps with multiple root layouts per: https://github.com/vercel/next.js/issues/59180#issuecomment-2828177089


stevekuznetsov:
image

Did you want styling as part of this PR or another?

stevekuznetsov:
Also not sure we want the breadcrumbs on 404

src/app/[...segmentsNotFound]/layout.tsx
```diff
@@ -0,0 +1,115 @@
+import { cn } from '@/utilities/ui'
+import localFont from 'next/font/local'
+import '../(frontend)/globals.css'
+
+const lato = localFont({
```
Comment (rchlfryn) [inline src/app/[...segmentsNotFound]/layout.tsx:5]: Interesting, any reason we are not using google as a host?

rchlfryn (approved):
We can style this later if we have time. Might be fun to do something related to being off trail or backcountry related.

Approved pending removing breadcrumbs on this page

src/app/[...segmentsNotFound]/layout.tsx
Comment (busbyk) [reply src/app/[...segmentsNotFound]/layout.tsx:5]: Yea see this PR: https://github.com/NWACus/web/pull/194 

I had found a link to a years-old GH issue explaining why Google doesn't host all weights for Lato but I couldn't find it again easily

busbyk:
> Did you want styling as part of this PR or another?

> We can style this later if we have time. Might be fun to do something related to being off trail or backcountry related.

I considered designing a fun 404 page but I lost to the feeling that it felt like a waste of time given the amount of work we still have to be ready for the season. 

But I threw some basic styling in. I'd love to make it a little more fun but I don't want to spend a lot of time on it now. 

center-scoped 404
CleanShot 2025-07-20 at 12 58 24

global 404
CleanShot 2025-07-20 at 12 57 57

Also it's weird that there is no better way to detect if the not-found page is being rendered in Next.js besides using context 🤷 : https://github.com/vercel/next.js/discussions/61823#discussioncomment-10649483

PR #334, opened 2025-07-17
Migrations + production readiness

### LINKED ISSUE #311: Persistent production environment

busbyk:
Resolves #311 

Database Changes:
- Moved from dropping/recreating databases to persistent production environment
- Added database migrations instead of re-seeding on prod deploys
- Created database sanitization for dev environment (sanitize-db.ts)

Config Changes:
- Updates CORS rules to account for production tenant custom domains
- Adds CSRF prevention by generating allowed domains

Workflow Updates:
- Production workflow now just runs migrations instead of recreating DB
- Development workflow clones prod DB, sanitizes it, syncs files, and runs migrations
- Added new sync-prod-to-dev.yml workflow for daily prod-to-dev syncing

File Storage:
- Added environment-specific prefixes for Vercel Blob storage
- Added scripts for syncing blobs between environments (sync-blob-storage.ts)
- Added prefix field to Media collection to track environment

New Scripts:
- sanitize: Sanitizes database (changes emails/passwords)
- sync-blobs: Syncs files between environments
- update-media-prefix: Updates media prefix in database
- migrate: Runs database migrations

Test Updates:
- Updated tenant tests to use PRODUCTION_TENANTS constant instead of mocked function

Dependencies:
- Added @vercel/blob, dotenv, tsx packages
- Removed ts-node in favor of tsx

This PR establishes persistent environments with proper data management between prod/dev/preview environments.


.github/workflows/production.yaml
```diff
@@ -59,14 +43,13 @@ jobs:
...
-      - name: Seed the database
-        run: pnpm seed
+      - name: Run database migrations
+        run: pnpm migrate
         env:
           DATABASE_URI: '${{ steps.connection.outputs.uri }}'
           DATABASE_AUTH_TOKEN: '${{ steps.connection.outputs.token }}'
           PAYLOAD_SEED_PASSWORD: '${{ secrets.PAYLOAD_SEED_PASSWORD }}'
           PAYLOAD_SECRET: '${{ secrets.PAYLOAD_SECRET }}'
-          NEXT_PUBLIC_GOOGLE_MAPS_API_KEY: '${{ secrets.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY }}'
```
stevekuznetsov, line 69: Why?

.github/workflows/production.yaml
```diff
@@ -14,22 +14,6 @@ jobs:
         run: curl -sSfL https://get.tur.so/install.sh | bash
         env:
           TURSO_INSTALL_SKIP_SIGNUP: 'true'
-      # TODO: once we're ready for a persistent database, don't delete anything, just run migrations
```
stevekuznetsov, line 17: Gotta love a deleted TODO

src/migrations/index.ts
```diff
@@ -0,0 +1,9 @@
+import * as migration_20250717_013057 from './20250717_013057'
```
stevekuznetsov: Not sure what I was expecting but certainly not a 10kLoC JSON file ... what am I even looking at? Are we able to store the SQL migrations instead, or something that could be reviewed by humans?

busbyk: Payload uses Drizzle as their ORM under the hood but it seems they wrote their own migration approach. Payload generates this .json metadata file + typescript with up/down functions file. Apparently, this is so they can be database agnostic. Which makes sense. 

The migration's .ts file is the closest you're going to get to the SQL migrations without building our own migrations system. Looks like they effectively are just running sql commands in there anyways - just wrapped in db.run calls. Interesting. 

Migrations should be much smaller going forward. This is effectively the initialization of our db schema since we've just been using push mode up to this point.

busbyk:
This is ready for a review but I'd like to hold off on merging this until we get other open PRs merged.

We'll need to edit a row in the migrations table to get the initial migration appear to have been run already (since we've been operating in push mode and all of those changes have already been made). 

So the steps will be:
- Merge other open PRs (i.e. deploy to dev)
- Essentially mark init migration as already run in dev db (https://discord.com/channels/967097582721572934/1280169446924615844)
- Merge this PR (i.e. deploy to dev)
- Mark init migration as already run in prod db
- Fast forward `release` (i.e. deploy to prod)

src/scripts/sanitize-db.ts
```diff
@@ -0,0 +1,66 @@
...
+      depth: 0,
+    })
+
+    console.log(`Found ${users.docs.length} users to sanitize`)
+
+    for (const user of users.docs) {
+      const originalEmail = user.email
+      const sanitizedEmail = createEmailAlias(originalEmail, emailDefaultReplyToAddress)
+
+      console.log(`Updating user: ${originalEmail} -> ${sanitizedEmail}`)
```
yuliadub: since this is gonna run in a GAH - this would mean we will print all of user emails in the github actions logs, right? if that is so, since our repo is public lets remove this log statement to avoid exposing all the emails/users.

src/scripts/sanitize-db.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const payload = await getPayload({ config })
+
+  if (!process.env.PAYLOAD_SEED_PASSWORD && process.env.ALLOW_SIMPLE_PASSWORDS !== 'true') {
+    payload.logger.fatal(
+      "$PAYLOAD_SEED_PASSWORD missing and ALLOW_SIMPLE_PASSWORDS not set to 'true' - either opt into simple passwords or provide a seed password.",
+    )
+    process.exit(1)
+  }
+  const password = process.env.PAYLOAD_SEED_PASSWORD || 'localpass'
```
yuliadub: thinking more about our chat in standup even if we did different passwords for different users, in reality if someone got one password to log into our site, they can likely figure out all of them even in they were different. So I think sticking with all accounts having one password for now is fine. 

I do think it would be good to refresh the one password each time we do a sanitize so it doesnt remain static, ideally by fully generating a new one and storing it somewhere (like a key vault that we can access if we needed it) or at least by changing some part of our existing password based on something like a date or something similar (so you can easily adjust it if you know the base password formula)

creating a password wall with vercel for dev env as we talked about would be good too on top of this.

src/scripts/sanitize-db.ts
```diff
@@ -0,0 +1,66 @@
...
+    console.log('Database sanitization completed successfully!')
+  } catch (error) {
+    console.error('Error during sanitization:', error)
+    process.exit(1)
+  }
+
+  process.exit(0)
+}
+
+sanitizeDatabase()
```
yuliadub, line 72: part of me wonders if it is worth it to just drop users table and recreate it with the few basic accounts we use now, but I guess that likely would mean some broken connections between content users created in other collections which would not be ideal

src/utilities/createEmailAlias.ts
```diff
@@ -0,0 +1,35 @@
+export function createEmailAlias(
+  originalEmail: string,
```
yuliadub: I think I would prefer to just convert original emails to something that didnt expose the persons name or parts of their real email unless we have a specific need to keep a form of the original email (i.e. just generate a random 10 chars for each user and put then after developer+ahsjdhkjghskdjfhskjdf@nwac.us or something like that)

yuliadub (approved):
left my two cents! I think it would be good to:
- hide dev behind a login in vercel
- refresh the db dev password at least occasionally vs it being fully static
- take out all parts of original user emails unless we have a specific need for them

src/scripts/sanitize-db.ts
busbyk: Yea I think the empty references wouldn't be ideal.

Issue #209, opened 2025-05-20
Cached API route used by middleware for tenant config


busbyk:
Right now we hard code the tenant data in the middleware which will be problematic for onboarding new centers without dev changes being required. 

https://github.com/NWACus/web/blob/nac-capabilities-and-metadata/src/middleware.ts#L17

This would introduce an API route that provides the same data but is cached and then served from the CDN.

stevekuznetsov:
If we find this has any serious impact on load times, the list of centers is fairly static and their mapping to the domains is even more so, so it's not the end of the world to just slap a constant into there.

busbyk:
Yea agreed. The only desire for this is to avoid a code change for onboarding. But if we're OK with that then we don't need to do this. The way it's currently hardcoded works.

stevekuznetsov:
We can also avoid code changes for onboarding by pre-populating all centers ;)

I think having it pull from the DB keeps things from drifting and is 100% the more correct approach, so let's give it a shot and know if it costs too much in performance/load time we have a good-enough backstop.

busbyk:
> We can also avoid code changes for onboarding by pre-populating all centers ;)
> 
> I think having it pull from the DB keeps things from drifting and is 100% the more correct approach, so let's give it a shot and know if it costs too much in performance/load time we have a good-enough backstop.

Both excellent points!

PR #336, opened 2025-07-21
scoping the admin panel to a tenant based on subdomain or custom domain

### LINKED ISSUE #203: Admin panel on tenant-scoped subdomain/custom domain
Right now if you have access to multiple tenants through our RBAC you will always see the tenant picker regardless of if you're on a subdomain or not. 

It might make sense to restrict the user to the tenant of the subdomain/custom domain regardless of if they have roles with rules that give them access to multiple tenants. 

So if a super admin visits `nwac.avy-fx.org/admin` they would be restricted to _only_ nwac documents even though they have access to all tenants. If that user wants to see multiple tenants in one admin panel they should visit `www.avy-fx.org/admin`. 

This bypasses our RBAC a bit but I think that's OK if this seems like more of an expected behavior.

What are everyone's thoughts?

busbyk:
Resolves #203 

This PR implements tenant domain scoping for the admin panel. Visiting a tenant scoped subdomain or custom domain sets the `payload-tenant` cookie to the id of that tenant which applies our tenant base list filter on top of permissions. 

Key Changes

- When accessing the admin panel via a tenant's subdomain (e.g., nwac.localhost) or custom domain (e.g., nwac.us/admin), the admin panel is scoped to that tenant
- The users collection now filters to only show users with a tenant scoped role matching the domain's tenant or if users with a global role if the logged in user has a global role granting read on the users collection.
- The tenant selector is hidden when the admin panel is scoped to a specific tenant
- New context provider to track current admin view types across components (required for hiding the tenant selector). Surprisingly this isn't available to server components and in some contexts. The beforeNavLinks components for example do not receive this in their props.  

Technical changes

- Added isTenantDomainScoped utility to detect when a domain corresponds to a specific tenant
- Implemented filterByTenantScopedDomain base list filter for the Users collection
- Refactored TenantSelector into client/server components for conditional rendering
- Using middleware to set tenant cookie if on a tenant scoped subdomain or custom domain
- Removed the domain-based login hook in favor of middleware-based tenant scoping


busbyk:
Using the middleware to set the `payload-tenant` cookie is a very simple solution but since we use the tenant id in that cookie (the multi tenant plugin expects that I believe - at least how we've currently got it configured). But this means I had to add back the hardcoded ids in the middleware which could become out of sync with the db. 

So I'm leaving this as a draft while I try:
1. Changing the `payload-tenant` cookie to use the slug instead of the id of a tenant. Just need to make sure the multi tenant plugin can be configured to expect this too. 
or
2. The cached API route we've discussed for a while #209

busbyk:
> Using the middleware to set the `payload-tenant` cookie is a very simple solution but since we use the tenant id in that cookie (the multi tenant plugin expects that I believe - at least how we've currently got it configured). But this means I had to add back the hardcoded ids in the middleware which could become out of sync with the db.
> 
> So I'm leaving this as a draft while I try:
> 
> 1. Changing the `payload-tenant` cookie to use the slug instead of the id of a tenant. Just need to make sure the multi tenant plugin can be configured to expect this too.
>    or
> 2. The cached API route we've discussed for a while [Cached API route used by middleware for tenant config #209](https://github.com/NWACus/web/issues/209)

I'm just going to work #209 separately. Hardcoding the tenant list in middleware with ids is fine - that's what we used to have.

yuliadub:
just to make sure I am following, in theory if I go change the id of the cookie from tenant 1 to 2 I still wouldn't be able to see that tenants data because of the tenant that my user is assigned to, right?

busbyk:
> just to make sure I am following, in theory if I go change the id of the cookie from tenant 1 to 2 I still wouldn't be able to see that tenants data because of the tenant that my user is assigned to, right?

Well, that depends if your user has access to see tenant 2. But I believe what you're getting at is: does access control still apply? And yes, it does. 

Access control runs first using global roles and tenant scoped roles. So let's say your user has two tenant scoped roles:
- Admin role on NWAC
- Admin role on SAC

If your user logs in to nwac.{root domain}/admin their `payload-tenant` cookie is set to 1 (for NWAC), access control restricts results to just NWAC and SAC content, and then the base list filter would filter results to just NWAC content. 

If your user then changed the `payload-tenant` cookie to 2 (for SAC), they would see only SAC content because their access control allows them to see SAC content as well. 

If your user only had the Admin role on NWAC and they changed the `payload-tenant` cookie to 2 they wouldn't see any content at all because access control doesn't allow them to see SAC content so the base list filter causes no results to be returned. 

_But_ I just tested this out and because the `payload-tenant` cookie is set in the middleware, even when I change it manually it's immediately overwritten on a tenant subdomain/custom domain.

busbyk:
~~It just occurred to me that we should probably restrict users from logging in on a domain that they don't have any permission for. I.e. if an NWAC Admin tries logging in at sac.{root domain} we should probably throw a 401.~~

Updated ✅

src/middleware.ts
```diff
@@ -18,11 +18,11 @@ export const config = {
 
 const PRODUCTION_TENANTS = getProductionTenantSlugs()
 
-const TENANTS: { slug: string; customDomain: string }[] = [
-  { slug: 'dvac', customDomain: 'dvac.us' },
-  { slug: 'nwac', customDomain: 'nwac.us' },
-  { slug: 'sac', customDomain: 'sierraavalanchecenter.org' },
-  { slug: 'snfac', customDomain: 'sawtoothavalanche.com' },
+const TENANTS: { id: number; slug: string; customDomain: string }[] = [
+  { id: 1, slug: 'dvac', customDomain: 'dvac.us' },
```
stevekuznetsov: This will need to be environment-specific, e.g. different for prod and non-prod? And we will want to pre-create all of the tenants in prod, so onboarding flow doesn't need a code change and redeploy?

Have we also considered (async) a feature request to the plugin to store something else and give a hook to look up the tenant document based on that "something else"?

busbyk: > This will need to be environment-specific, e.g. different for prod and non-prod? And we will want to pre-create all of the tenants in prod, so onboarding flow doesn't need a code change and redeploy?

I am working on making the tenants dynamic for #209 but was going to accept this hardcoding for a little longer. We still want the dvac tenant in production. This is our "template tenant" which will be used for building pages that we want to duplicate to a new tenant during onboarding. 

I pre-populated all of the ACs in this list in #334 (copied the values from the nwacus/avy codebase) but because I need the id for the `payload-tenant` cookie we'd need to create the tenants in exactly the right order 😬 I don't like that. I should have a PR open for #209 by the EOD so I think that should solve our problems related to this hardcoded tenant list. 

> Have we also considered (async) a feature request to the plugin to store something else and give a hook to look up the tenant document based on that "something else"?

Sorry, I'm not following this. Can you explain differently? Are you talking about the `@payloadcms/plugin-multi-tenant` plugin?

stevekuznetsov: Yeah, I meant for the plugin provided by Payload - if the value of the cookie could be some other unique identifier that's easier to know a priori (and perhaps tied to the domain, like in our case), we would not have had to do any of that work.

PR #337, opened 2025-07-21
truncating db names in turso create

busbyk:
Turso has a 58 character db name limit


.github/workflows/cleanup.yaml
```diff
@@ -16,7 +16,13 @@ jobs:
         run: |
           ref="${{ github.event.ref }}"
           branch="${ref##refs/heads/}"
-          name="payloadcms-preview-${branch//[^a-z0-9\-]/x}"
+          clean_branch="${branch//[^a-z0-9\-]/x}"
+          # Truncate branch name to ensure total database name stays within 58 character limit
```
stevekuznetsov: You could shove the inputs through `sha256sum` and grab the first 39 to make it relatively impossible to have collisions - esp with branch names, common prefixes are not rare, might be good to handle

busbyk: Yea the collision avoidance would be nice but I have enjoyed being able to just go straight to turso and pull up my branch's db by name. Hmm what about a hybrid approach? See new commit.

PR #338, opened 2025-07-22
Users can only read and update self

busbyk:
We had been granting access for any method on a user's own document. This PR changes that to only allow read and update. 

If a user has a role or global role that grants create or delete on users they will still be able to perform those actions. This just changes what we allow on "self" if no role rule grants access prior to the self check. 

Before:
image

After: 
image


src/collections/Users/access/byGlobalRoleOrTenantRoleAssignmentOrDomain.ts
```diff
@@ -70,8 +70,8 @@ export const byGlobalRoleOrTenantRoleAssignmentOrDomain: (method: ruleMethod) =>
       }
     }
 
-    // allow users to read their own record
-    if (args?.id === args.req.user.id) {
+    // allow users to read, update their own record
+    if (args?.id === args.req.user.id && method !== 'delete' && method !== 'create') {
```
rchlfryn: Very ⛏️ I would write this as
```suggestion
    if (args?.id === args.req.user.id && (method === 'read' || method === 'update')) {
```
But what you have is fine with the comment!

rchlfryn (approved):

src/collections/Users/access/byGlobalRoleOrTenantRoleAssignmentOrDomain.ts
busbyk: I was thinking about how `method` can also be `*`.

_But_, I don't think that really applies here since the `accessBy...` function always uses CRUD actions. 

So I think we'd be safe to explicitly allow read and update and ignore the `*` case. I'm on board, I'll change this.

PR #341, opened 2025-07-23
Dynamic tenants in middleware

### LINKED ISSUE #209: Cached API route used by middleware for tenant config

busbyk:
Resolves #209 

Updates the middleware to use a dynamic tenants list. This uses a combination approach of a cached API route and build-time generation that should produce a reliable and up-to-date tenants list. 

See the decision doc and README updates for full explanation. 

The TL;DR is that we prefer the cached API route's response which is cached using `unstable_cache` but fall back to a tenants list that is generated during build time (also during local dev). This cache is revalidated via Payload hooks on the tenants collection to ensure accurate data. 

The static tenants generation was a bit tricky because the generated file needs to exist for Payload to run correctly and not get an import error. So there is a seed file that gets committed to git as a placeholder (essentially) but then both `pnpm dev` and `pnpm build` will generate the file based on the current environment's database in the `predev` and `prebuild` scripts respectively.


busbyk:
I do wonder if the build time generation is even worth it 🙄

src/scripts/generate-static-tenants.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  console.log(`Using tenant data from: ${dataSource}`)
+
+  // Ensure the generated directory exists
+  const generatedDir = join(process.cwd(), 'src/generated')
+  try {
+    mkdirSync(generatedDir, { recursive: true })
+  } catch {
+    // Directory might already exist, that's fine
```
stevekuznetsov: Any chance you can catch just that exception and let others through (no space left on device, out of inodes, whatever)

stevekuznetsov (approved):
I can appreciate that this was a monumental amount of work but I really like the approach and I think it will serve us well, thanks for tackling that.

src/scripts/generate-static-tenants.ts
busbyk: Ahh yes. This is actually a misleading comment on my part. Because I added `{ recursive: true }` we shouldn't get directory already exists errors anymore (`EEXIST`). So I updated the catch block to log the error and `process.exit(1)` because it will only throw legitimate errors with `{ recursive: true }`. 

Thanks for catching that.

busbyk:
> I can appreciate that this was a monumental amount of work but I really like the approach and I think it will serve us well, thanks for tackling that.

Thanks!

Issue #293, opened 2025-06-24
Blocks should handle missing data


busbyk:
The content block crashes the app when a background color is not set. This can happen when adding it as a new block and using live preview. 

Steps to reproduce: 
1. Open live preview on any Page document
2. Add a content block to the layout

![Image](https://github.com/user-attachments/assets/dc3a7559-cb77-4c85-aa49-40e690cc69a4)

This will likely happen for any component using the [color field](https://github.com/NWACus/web/blob/main/src/fields/color.ts) but might be a larger issue to explore. I think the scenario where this will happen is anywhere we're accessing attributes of a field on a block that doesn't have a default value for that field. 

Even though a field is required by Payload that doesn't mean it will be present because of live preview. If it's required _and_ there's a default value that's probably safer. 

We probably also want to add error boundaries so we don't get the default next.js error boundary but I'll add that as a separate issue.

PR #342, opened 2025-07-24
Fix background color bug

### LINKED ISSUE #179: Section-level blocks accept a background color from our theme colors
https://github.com/NWACus/web/pull/153 introduces a color picker. The options for this color picker should be the semantic classnames we're using for our themes: i.e. primary, secondary, card, header, footer, etc. + brand-50 - brand-950. 

An editor should be able to pick from any of these colors and have that semantic classname be used in the rendered block. The display of these colors in the admin panel should vary by theme. So if the tenant for the document being edited is SAC, the SAC theme colors should display. 

This may also be a good opportunity to ensure that all existing blocks are using semantic class names for colors.

### LINKED ISSUE #282: Improve colorPickerField
* We used to have a ColorField that validated if the color picked was valid. It would be nice to have validation for the new color picker to validate that it is a valid color 
    * See #277 for [file](https://github.com/NWACus/web/pull/277/files#diff-859d5cb2bed94b628bcf15414cfc0725bb248091036ff4658a747f5fd0797daaL13) 
* Confirm that an initial color is selected when the block is first added
* Only show theme colors

### LINKED ISSUE #302: Bug: data lost when moving between edit and preview
To recreate, add a content block and toggle between edit and live preview

rchlfryn:
### Description
This PR updates the color picker to use brand colors and fixes the seeding issue that was showing the blocks background colors as unselected.

I ended up removing the color package which evaluated the color and returned light or dark since there was no way to access custom tailwind colors outside of a client function. We will need to maintain `brand-foreground` which I think is easier. I can add to the decision doc if necessary.

### Issue
Fixes #282 & fixes #302 & fixes #179


rchlfryn:
~I will likely need to add a migration now that #334 is merged in~

Updated in [7f20f03](https://github.com/NWACus/web/pull/342/commits/7f20f03fd283d8e7c77b44ad9581b1b32aa1a908)

src/utilities/getTextColorFromBgColor.ts
busbyk, line 1: Can you add some tests for this or just a function comment explaining this a little bit?

src/components/ColorPicker/index.tsx
```diff
@@ -1,39 +1,30 @@
...
+
 import { FieldLabel, useField } from '@payloadcms/ui'
 import { TextFieldClientProps } from 'payload'
 
-// TODO - import color list from theme
-// white, slate-300, slate-600, slate-700
-const colorOptions = ['#FFFFFF', '#CBD5E1', '#475569', '#334155']
-
 const ColorPicker = (props: TextFieldClientProps) => {
+  const colorOptions = ['white', 'brand-300', 'brand-600', 'brand-700']
```
busbyk: Keeping the options here limited is nice for the UI I suppose but I think the original thought was to have all of the brand color options here. See https://github.com/NWACus/web/blob/fix-bg-color-picker/docs/decisions/006-semantic-class-names-for-themes.md#L98 

But if we add too many colors we obv need a different UI than selectable circles since there would be too many options. So maybe leaving as is for now is sufficient until someone wants more colors? What are your thoughts?

busbyk: Might want to leave a comment here explaining that if more colors are added to this list the safeList will need to be updated?

src/migrations/20250725_165619_rename_color_to_background_color.ts
busbyk, line 1: My initial reaction was: "whoa there's a lot going on to rename color to backgroundColor". I had Claude summarize what's happening here which was pretty nice: 

This migration renames a database column from color to background_color across multiple page block tables. It updates six tables
  total:

  - pages_blocks_content
  - pages_blocks_image_quote
  - pages_blocks_image_text
  - _pages_v_blocks_content (version table)
  - _pages_v_blocks_image_quote (version table)
  - _pages_v_blocks_image_text (version table)

  The migration also changes the default value from 'white' (in up) to '#fffff' (in down). It uses the SQLite pattern of creating new
  tables, copying data, dropping old tables, and renaming - required since SQLite doesn't support direct column renames.

  ---

  Good stuff.

tailwind.config.mjs
```diff
@@ -11,18 +11,12 @@ const config = {
...
-    'border-success',
-    'bg-success/30',
-    'border-warning',
-    'bg-warning/30',
+    {
+      pattern: /^bg-(brand-(300|600|700)|card|(error|success|warning)\/30)$/,
+    },
+    {
+      pattern: /^border-(border|error|success|warning)$/,
+    },
```
busbyk: Looks like a lot of these are actually complete strings in our code which means that Tailwind would pick them up and they don't need to be on the safeList explicitly: https://github.com/NWACus/web/blob/fix-bg-color-picker/src/blocks/Banner/Component.tsx#L15

So we could just include the brand colors pattern to keep this simpler. Looks like everything except for the bg-brand- colors show up in ways that Tailwind will pick up.

busbyk (approved):
Looks good! 

What was the root cause of the losing data between edit and live preview views issue though? And error because of the defaultValue or something along those lines?

Left a few comments but approving because this works as expected.

src/migrations/20250725_165619_rename_color_to_background_color.ts
rchlfryn: This was all generated with `migrate:create` 🔥 I also had a similar reaction then was also impressed

src/utilities/getTextColorFromBgColor.ts
rchlfryn: Yes I can add some tests. I will add it to my routine now that you have them up and running!

rchlfryn:
> What was the root cause of the losing data between edit and live preview views issue though? And error because of the defaultValue or something along those lines?

This was the root cause of losing data on the live preview bug, but  not #293. Sadly the root cause was actually because we were passing `#ffffff` in the seed script and looking for `#FFFFFF` in the color picker so it did not select any value. Now that we are using strings this shouldn't happen again. I tried to use tailwind colors directly but it was not playing nice with it being a server function.

tailwind.config.mjs
rchlfryn: I dont think we even need to keep the brand color patten since those are complete strings as well. I also confirmed we were not using 'bg-card' anywhere. I remove the safelist in general.

src/components/ColorPicker/index.tsx
rchlfryn: The UI isn't terrible. I was able to just do 2 rows and show the 12 colors.

PR #343, opened 2025-07-24
Password protection non prod

busbyk:
It costs $150/mo to add password protection as an [add-on to the Pro plan](https://vercel.com/docs/deployment-protection#advanced-deployment-protection) called Advanced Deployment Protection. 

If we wanted to protect our non-prod environments using basic auth we could do so using this approach. 

Do we want this? 

Based on: https://medium.com/@ratudev/simple-password-protection-for-a-next-js-app-1ff21ada93a3

We would need to enter username and password once per session for every new subdomain (i.e. avy-fx.org -> nwac.avy-fx.org)

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,5 +1,6 @@
 import { STATIC_TENANTS } from '@/generated/tenants'
 import { NextRequest, NextResponse } from 'next/server'
+import { getEnvironmentFriendlyName } from './utilities/getEnvironmentFriendlyName'
 import { getURL } from './utilities/getURL'
 import { PRODUCTION_TENANTS } from './utilities/tenancy/tenants'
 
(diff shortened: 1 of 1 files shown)


rustynwac:
$150/mo is likely out. I don't mind this approach if you feel good about it. @busbyk

busbyk:
> $150/mo is likely out. I don't mind this approach if you feel good about it. @busbyk

Yea $150/mo is def not worth it. Just wanted to confirm we actually want to implement this. 

This would mean that the actual sites themselves are password protected too, not just the admin panel. 

We could just protect the admin panel if that's what we're worried about.

TBH I'm just questioning if we need password protection for our non-prod envs at all since the admin panel already has auth.

rustynwac:
The reason we would add password protection to non-prod envs is so that nobody except whoever we let in can see them? I dont know if I'm worried about that. Or it's a nice to have later...?

busbyk:
> The reason we would add password protection to non-prod envs is so that nobody except whoever we let in can see them? I dont know if I'm worried about that. Or it's a nice to have later...?

Yea that would be why. @yuliadub had thoughts on this but I think now that we're generating a random password for dev on every sync I think we're probably good and it doesn't matter too much that our non prod non-admin pages are publicly accessible?

rchlfryn:
@busbyk do you know why the checks stopped building?

busbyk:
I'm just going to close this. I don't think we want this. We can always re-open.

Issue #146, opened 2025-04-24
Polish 'Who We Are' Blocks


stevekuznetsov:
We got a decent first pass at the page showing off staff, forecaster and board member bios in #137. This issue tracks bringing that implementation over the finish line and getting final design input on how we expect users to interact with long-form biographies.

_Originally posted by @busbyk in https://github.com/NWACus/web/pull/137#discussion_r2059162901_

We should also add thumbnails to the blocks for these.

chrisharrisonjones:
@stevekuznetsov 
I can provide a full design comp if that's useful?

stevekuznetsov:
I'll defer to @busbyk or @rchlfryn since I think this is beyond my area of expertise - I think zeroing in on how we do the full-fledged biography text would be good!

busbyk:
@chrisharrisonjones I think just some feedback would be good. You can view the current state here: https://nwac.dev.avy-fx.org/who-we-are. 

But if you want to do a full design comp that's cool too!

chrisharrisonjones:
I think we should utilize the designed shadcn blocks for these pages. Using this type of functionality for long form text isn't working for me. 

![Image](https://github.com/user-attachments/assets/98fa6d90-25d6-4330-b9c2-058ef3be92e8)

chrisharrisonjones:
The UI/UX design that ACs have reviewed (approved?) can be found here: https://www.figma.com/design/vAJkcFrHt5gddho3ILBOEx/NWAC-UX-UI-Redesign---Work?node-id=626-6750&t=sZxT3Kyv1JwQYFSJ-1

If having a full design comp is helpful, let me know. Initially there were concerns that we have uneven content between ACs to enforce a specific design on each center. Let's discuss.

![Image](https://github.com/user-attachments/assets/52c8e64a-f5cf-45e6-8f0f-84558b261424)

chrisharrisonjones:
Sorry, i wasn't understanding the problem we were solving with this task. Here's a quick comp to solve for the single picture + extended detail text functionality utilizing the modal functionality from Events. Let me know if this is a workable solve. Thanks!

![Image](https://github.com/user-attachments/assets/da3389ac-7196-449c-a371-c746ee3ba7c9)

busbyk:
Some work to polish these components was done on #360. 

We still need to:
- Handle a single Biography component (it's full width right now)
[...]

PR #360, opened 2025-07-27
Blocks: Allow for unpopulated data & add missing thumbnails

### LINKED ISSUE #293: Blocks should handle missing data

### LINKED ISSUE #301: Blocks: Add missing layout preview images
When adding a new block, some "preview" images are missing. Finish adding these.

Image

rchlfryn:
### Description
- Previously, some blocks would break when added while viewing the live preview. This update introduces graceful handling to prevent such issues when users add blocks with the live preview open.
- Adds missing thumbnails for blocks

### Issue
Fixes #293 & fixes #301


src/blocks/Biography/config.tsx
```diff
@@ -9,11 +9,19 @@ export const BiographyBlock: Block = {
       relationTo: 'biographies',
       hasMany: false,
       required: true,
+      filterOptions: ({ data }) => {
+        return {
+          tenant: {
+            equals: data.tenant,
+          },
+        }
+      },
```
busbyk: It would be great to create a separate utility function for this tenant filter for relationship field filterOptions. Apparently, we should really be using this everywhere we have a relationship field to a collection that's tenant scoped (essentially all of them).

src/blocks/Biography/Biography.tsx
busbyk, line 1: Ooo nice, does this also resolve #146? 

I am getting a bunch of nested element errors though: 
CleanShot 2025-07-27 at 12 36 08

src/components/ColorPicker/index.tsx
```diff
@@ -17,25 +17,35 @@ const ColorPicker = (props: TextFieldClientProps) => {
     'brand-800',
     'brand-900',
     'brand-950',
+    'transparent',
```
busbyk, line 20: Ahh that's what this is 
CleanShot 2025-07-27 at 12 43 40@2x

That wasn't clear to me. Hmm I wonder if we could like add a tooltip or use a grid-like image like: 
CleanShot 2025-07-27 at 12 44 44@2x

?

busbyk (changes requested):
The thumbnails look great! Looks like we're still missing one for the form block though: 
CleanShot 2025-07-27 at 12 31 05@2x

Just a couple things but looks great!

src/components/ColorPicker/index.tsx
rchlfryn: I was being lazy - updated in [969b561](https://github.com/NWACus/web/pull/360/commits/969b5619b9df6e8d3cdd2476b8c270e39352258c)

src/blocks/Biography/Biography.tsx
rchlfryn: Wow - I copied th dialog from Shadcn and it was so wrong. Fixed in [9a460cf](https://github.com/NWACus/web/pull/360/commits/9a460cf1f452f626c1949fc55d2a786dd510f435)
Screenshot 2025-07-27 at 13 33 16

src/blocks/Biography/config.tsx
rchlfryn: Agreed - done in [bc1dd35](https://github.com/NWACus/web/pull/360/commits/bc1dd35d26ac67316396f7d64daf6dc1a9b160e1)

rchlfryn:
@busbyk I think I am going to hold off on adding this since I am not confident we will even keep Forms block. From the way it is sounding, every form will be embedded.

busbyk (approved):
Looks great, thanks for making those changes!

There are a few things that will still need to be done on #146 but I'll add those to that issue.

Issue #315, opened 2025-07-03
Remove forecast archive, single forecast link for single zone center


busbyk:
- Remove the forecast archive link from the forecasts navigation dropdown
- Change the behavior for single zone centers to have a no-dropdown link with label "Avalanche Forecast" instead of a zones dropdown

PR #362, opened 2025-07-27
Fixes NAC widget/ navigation bug

### LINKED ISSUE #292: Forecast zone link doesn't navigate when on zone forecast archive
When visiting the forecast archive from a specific zone, clicking the zone link does not navigate you to the zone forecast anymore. The URL is updated but the widget does not update it's internal routing because the zone hash is not added to the url. 

I suspect this has something to do with the WidgetHashHandler component. 

Steps to reproduce: 
1. In the forecasts dropdown click on a specific zone.
2. Open the date picker and pick a previous date with a forecast.
3. Click view forecast archive in the gray callout card/banner that appears.
4. In the forecasts dropdown click on the same zone as in step 1.

Image

rchlfryn:
### Description
When visiting the forecast archive from the date picker, clicking the zone link does not navigate you to the zone forecast anymore. The URL is updated but the widget does not update it's internal routing because the zone hash is not added to the URL. This PR hides the `All Forecast Archive` button in the date picker.

### Issues
Possibly fixes #292

### DIFF
--- a/src/utilities/useHash.ts
+++ b/src/utilities/useHash.ts
@@ -3,7 +3,9 @@
 import { useEffect, useState } from 'react'
 
 export function useHash() {
-  const [hash, setHash] = useState('')
+  const [hash, setHash] = useState(() =>
+    typeof window !== 'undefined' ? window.location.hash : '',
+  )
 
   useEffect(() => {
     // Set initial hash asynchronously
--- a/src/utilities/useHash.ts
+++ b/src/utilities/useHash.ts
@@ -1,0 +1,46 @@
+'use client'
+
+import { useEffect, useState } from 'react'
+
+export function useHash() {
+  const [hash, setHash] = useState('')
+
+  useEffect(() => {
+    // Set initial hash asynchronously
+    setTimeout(() => {
+      setHash(window.location.hash)
+    }, 0)
+
+    const updateHash = () => {
+      // Defer state update to avoid useInsertionEffect error
+      setTimeout(() => {
+        setHash(window.location.hash)
+      }, 0)
+    }
+
+    window.addEventListener('hashchange', updateHash)
+    window.addEventListener('popstate', updateHash)
+
+    const originalPushState = window.history.pushState
+    const originalReplaceState = window.history.replaceState
+
+    window.history.pushState = function (...args) {
+      originalPushState.apply(window.history, args)
+      setTimeout(updateHash, 0)
+    }
+
+    window.history.replaceState = function (...args) {
+      originalReplaceState.apply(window.history, args)
+      setTimeout(updateHash, 0)
+    }
+
+    return () => {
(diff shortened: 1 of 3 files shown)


rchlfryn:
Not sure if this is the right solution to hide the archive in the date picker, or should we try to solve why the hash is not added to the URL. Thoughts?

src/utilities/useHash.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { useEffect, useState } from 'react'
+
+export function useHash() {
+  const [hash, setHash] = useState('')
+
+  useEffect(() => {
+    // Set initial hash asynchronously
+    setTimeout(() => {
+      setHash(window.location.hash)
```
rchlfryn, line 13: I think we need a check here to handle `#/archive/forecast ` by redirecting it to `/forecasts/avalanche/archive`. Without this, the bug is fixed, but  if the user copies and shares the URL (`http://nwac.localhost:3000/forecasts/avalanche/olympics#/archive/forecast`) it will take them to the zone rather than the all archived. 

Thoughts since we might be removing this is #315 ?

```suggestion
      const currentHash = window.location.hash
      // Handle '#/archive/forecast'
      if (currentHash === '#/archive/forecast' && window.location.pathname !== '/archive') {
        window.history.replaceState(null, '', `/forecasts/avalanche/archive${currentHash}`)
      }
      setHash(window.location.hash)
```

busbyk: Ah good catch on this. I think the issue is really that the hash isn't getting set to the actual hash on initial render. This causes a route with a set hash (i.e. http://nwac.localhost:3000/forecasts/avalanche/olympics#/forecast/1/165572 to get reset to the initialHash because the useState `hash` gets set to `''` as it's initial state. 

Using   
```
const [hash, setHash] = useState(() =>
    typeof window !== 'undefined' ? window.location.hash : '',
  )
```

Fixes this an allows deep linking to widget views other than the initialHash's widget view (i.e. .../olympics/#olympics). 

The fact that you can deep link to the all forecasts view on a zone page is something that I think we should accept since we're removing the navigation forecast archive link and that we'll eventually be moving away from the widgets. 

It would be a little weird to get a link like `http://nwac.localhost:3000/forecasts/avalanche/olympics#/archive/forecast` and be sort of in a zone in terms of our routing but viewing the archive for all zones. But since there's no back button the user would need to use one of our navigation links to get anywhere else anyways. 

I just don't love the idea of adding a special case here for the forecast archive. But I could be convinced that it's worth it...

rchlfryn: I'm not crazy about a special case for the forest archive either. Since most of us didn't even know this button existed, I think we can ship it as it without the commit I suggested. If we do an exception for this, we can add it later.

PR #359, opened 2025-07-26
Migrate middleware to use edge config

busbyk:
See the new decision doc for details `docs/decisions/008-edge-config-tenant-lookup.md`.

The build-time generation was complicated and flawed due to the edge runtime. Edge config feels like a really nice solution and dang it's fast.

PR #363, opened 2025-07-27
Escalation check

### LINKED ISSUE #313: Role assignments escalation check

busbyk:
Resolves #313 

Adds an escalation check:
- on beforeValidate for globalRoleAssignments
- on beforeValidate for roleAssignments
- in invite user action (checks each role and global role assignment)

I tested locally by creating some new roles manually with users:create, roleAssignments:create, etc. etc. I didn't add these to the seed since I don't think we'll want those long term. 

To test escalation check on role assignments, add the following rules to the Forecaster role: 
CleanShot 2025-07-27 at 00 44 49

Then: 
- log in as a Forecaster user
- try to self assign the Admin role (should throw error)
- try to invite a user with the Admin role (should throw error)

To test escalation check on global role assignments, add the following rules to a new global role: 
CleanShot 2025-07-27 at 00 47 03

Then: 
- assign a user that global role
- log in as that user
- try to self assign the Super Admin global role (should throw error)
- try to invite a user with the Super Admin global role (should throw error)


rchlfryn:
@busbyk looks like this has #359 code in it. Did you mean to have that as the base branch?

busbyk:
> @busbyk looks like this has #359 code in it. Did you mean to have that as the base branch?

Ah well meant to make the target branch `edge-config` (just updated) but in hindsight I should have created this branch off of `main` since I don't actually need any of the `edge-config` changes here.

src/collections/Users/components/inviteUserAction.ts
```diff
@@ -138,12 +153,31 @@ export async function inviteUserAction({
         Array.isArray(globalRoleAssignments) &&
         globalRoleAssignments.length > 0
       ) {
-        for (const assignment of globalRoleAssignments) {
+        const populatedGlobalRolesRes = await payload.find({
```
rchlfryn, line 156: ⛏️  Any reason you dont want to use `findById` here as well?

rchlfryn (approved):
LG pending one comment

src/collections/Users/components/inviteUserAction.ts
busbyk: Yea my thinking is that because globalRoleAssignments is just an array of ids I can make a single request to the db and then check canAssignGlobalRole on each vs. the roleAssignments is an array of objects (roleId and tenantId).

Technically I could do the same thing with roleAssignments where I map the roleAssignment.roleId into an array and use a similar where.id.in query but I need the tenantId for the canAssignRole function so I thought that would get too messy.

PR #365, opened 2025-07-30
Fix video rendering

### LINKED ISSUE #299: VideoMedia causes hydration error
It just needs to use `getMediaURL` with `getHostnameFromTenant` like the ImageMedia https://github.com/NWACus/web/blob/main/src/components/Media/ImageMedia/index.tsx#L59

But there are also commented out lines in there and we may want to add video controls to the video element. It just plays on a loop right now.

rchlfryn:
### Description
When adding a video through the media block, it was causing a hydration error and not loading. This PR fixes that bug.

### Issues
Fixes #299

### DIFF
--- a/src/components/Media/VideoMedia/index.tsx
+++ b/src/components/Media/VideoMedia/index.tsx
@@ -1,7 +1,7 @@
 'use client'
 
 import { cn } from '@/utilities/ui'
-import { useEffect, useRef } from 'react'
+import { useRef } from 'react'
 
 import { useTenant } from '@/providers/TenantProvider'
 import { getMediaURL } from '@/utilities/getURL'
@@ -12,17 +12,6 @@ export const VideoMedia = (props: MediaProps) => {
   const { onClick, resource, showVideoControls, videoClassName } = props
   const { tenant } = useTenant()
   const videoRef = useRef<HTMLVideoElement>(null)
-  // const [showFallback] = useState<boolean>()
-
-  useEffect(() => {
-    const { current: video } = videoRef
-    if (video) {
-      video.addEventListener('suspend', () => {
-        // setShowFallback(true);
-        // console.warn('Video was suspended, rendering fallback image.')
-      })
-    }
-  }, [])
 
   if (resource && typeof resource === 'object') {
     const { url } = resource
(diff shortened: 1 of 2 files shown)


src/components/Media/VideoMedia/index.tsx
```diff
@@ -3,12 +3,14 @@
...
+import { getHostnameFromTenant } from '@/utilities/tenancy/getHostnameFromTenant'
 import type { Props as MediaProps } from '../types'
 
 export const VideoMedia = (props: MediaProps) => {
-  const { onClick, resource, videoClassName } = props
-
+  const { onClick, resource, showVideoControls, videoClassName } = props
+  const { tenant } = useTenant()
   const videoRef = useRef<HTMLVideoElement>(null)
   // const [showFallback] = useState<boolean>()
```
busbyk: Should we just remove this and the useEffect if we're not actually rendering a fallback?

src/components/Media/VideoMedia/index.tsx
```diff
@@ -23,20 +25,23 @@ export const VideoMedia = (props: MediaProps) => {
...
+    const cacheTag = resource.updatedAt
+
+    const src = getMediaURL(url, cacheTag, getHostnameFromTenant(tenant))
 
     return (
       <video
         autoPlay
         className={cn(videoClassName)}
-        controls={false}
+        controls={showVideoControls || true}
```
busbyk: I think we should use nullish coalescing here because if `showVideoControls` is passed as `false` it would actually evaluate to `true` with the use of `||` here. 

```suggestion
        controls={showVideoControls ?? true}
```

src/components/Media/VideoMedia/index.tsx
```diff
@@ -23,20 +25,23 @@ export const VideoMedia = (props: MediaProps) => {
...
   if (resource && typeof resource === 'object') {
-    const { filename } = resource
+    const { url } = resource
+    const cacheTag = resource.updatedAt
+
+    const src = getMediaURL(url, cacheTag, getHostnameFromTenant(tenant))
 
     return (
       <video
         autoPlay
```
busbyk: What do you think about just turning autoPlay off?

busbyk (changes requested):

busbyk (approved):
🤘

PR #367, opened 2025-07-31
Add Suspense to blog

### LINKED ISSUE #354: Blog static vs dynamic rendering: 500 error in prod
We get a 500 error when visiting a center's blog page (example: nwac.avy-fx.org/blog) in production. 

Start of relevant logs: https://telemetry.betterstack.com/team/387921/tail?s=1386655&rf=now-60m&pt=overview&q=vercel.proxy.host%3D%22nwac.avy-fx.org%22&a=1753559522379000.96661396

Sentry issue: https://nwac.sentry.io/issues/6769537238/events/06ee95871fd64e9da73356ea59c0bed5/

The issue here is that Next.js is trying to statically generate this page but it's inherently dynamic because of the use of `const resolvedSearchParams = await searchParams`. 

Recommendations from Claude which I agree with: 

Solutions:
Option 1: Force Dynamic Rendering (Quick Fix)
Add this export to your page component to tell Next.js this route should always be dynamic:
typescriptexport const dynamic = 'force-dynamic'

Option 2: Use Suspense Boundary (Recommended)
Wrap the component that uses searchParams in a Suspense boundary:
typescriptimport { Suspense } from 'react'

export default function Page({ params, searchParams }: Args) {
  return (
    <Suspense fallback={<div>Loading...</div>}>
      <BlogContent params={params} searchParams={searchParams} />
    </Suspense>
  )
}

async function BlogContent({ params, searchParams }: Args) {
  // Your existing component logic here
  const { center } = await params
  const resolvedSearchParams = await searchParams
  // ... rest of your code
}
Option 3: Make it Truly Static (Best for Performance)
If you want to keep it static, you'd need to move the search/filter logic to client-side:

Pre-render the page with default content
Use client-side components for sorting/filtering
Update the URL with useRouter and useSearchParams

Option 1 does indeed seem the best for now. Options 2 and 3 would be nice to try.

rchlfryn:
### Description
This PR uses Solution 2 from #354 . It seems the [Sentry error](https://nwac.sentry.io/issues/6769537238/?project=4509675069702145&query=is:unresolved) has been resolved but this is not a bad thing to add to the blog.

### Issues
Fixes #354

### DIFF
--- a/src/app/(frontend)/[center]/blog/page.tsx
+++ b/src/app/(frontend)/[center]/blog/page.tsx
@@ -1,11 +1,12 @@
+import configPromise from '@payload-config'
 import type { Metadata } from 'next/types'
+import { getPayload } from 'payload'
+import { Suspense } from 'react'
 
 import { PageRange } from '@/components/PageRange'
 import { Pagination } from '@/components/Pagination'
 import { PostCollection } from '@/components/PostCollection'
 import { POSTS_LIMIT } from '@/utilities/constants'
-import configPromise from '@payload-config'
-import { getPayload } from 'payload'
 import { PostsSort } from './posts-sort'
 import { PostsTags } from './posts-tags'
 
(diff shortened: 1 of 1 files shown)


busbyk (approved):
🙌

PR #369, opened 2025-07-31
ignoring /assets urls from middleware to avoid rewriting those urls

### LINKED ISSUE #290: Favicon doesn't load for subdomain admin panels
Image

busbyk:
Resolves #290 

The /assets/icon.png url was being rewritten to /{tenant slug}/assets/icon.png which doesn't exist in our routing so we can ignore these from the middleware and have them served from the root domain.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -13,7 +13,7 @@ export const config = {
      * 4. all root files inside /public (e.g. /favicon.ico)
      * 5. /media and /thumbnail (inside /public)
      */
-    '/((?!api|_next|_static|_vercel|[\\w-]+\\.\\w+|media|thumbnail).*)',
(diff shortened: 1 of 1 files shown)


rchlfryn (approved):

rchlfryn:
What are the `favicon.(png|ico|svg)` used for under `public`? The client side?

src/middleware.ts
```diff
@@ -13,7 +13,7 @@ export const config = {
      * 4. all root files inside /public (e.g. /favicon.ico)
      * 5. /media and /thumbnail (inside /public)
```
rchlfryn, line 14: ⛏️  update this comment or remove it

busbyk:
> What are the `favicon.(png|ico|svg)` used for under `public`? The client side?

Yea they're used for all routes outside of Payload. We include <link> tags for them in our layouts.

PR #370, opened 2025-08-01
Change behavior of editing slug for pages

### LINKED ISSUE #340: Slug field can't be different than title
### Updated requirements

- Do away with the lock functionality completely
- Remove functionality to automatically update the slug field when the title field changes
- Add a button to generate the slug based on the title where the current lock button is (circle arrow type icon) if a user wants to do that
- If there is already a document in the collection with the same slug, keep the current functionality where an error is thrown and its up to the user to change the slug to something unique rather than appending an incrementing number suffix.

### Background
The [slug field](https://github.com/NWACus/web/blob/main/src/fields/slug/index.ts) is used in Pages, Posts, Tags. It bases the slug on the title field of those collections. There is a concept of the slug being locked or unlocked. 

When unlocked, you can set the slug to whatever value you want and publish. But if you lock the slug, it get's set back to the slug representation of the title of the document and saves a draft with this change. 

I'm not entirely sure what the intent is with the locking of the slug field. It strikes me as odd that we wouldn't be able to change the slug to something besides the direct slug representation of the document title. There might be a scenario where we want the slug to be slightly different than the title. 

This certainly isn't critical but it struck me as odd. 

My expected behavior would be that after unlocking, changing the slug value, and then re-locking, the value would stay as the changed slug value. 

What are everyone's thoughts?

rchlfryn:
### Description
The current unlock/edit/lock slug behavior for pages is confusing.

Current behavior:
slug has a value of 'contact'. I click Unlock-button, change the value of the slug to 'contact-us', click lock-button.. It show the slug value as 'contact'

Expected behavior:
Remove automatic updating of the slug and add generate button

### Issues
Fixes #340

### DIFF
--- a/src/fields/slug/SlugComponent.tsx
+++ b/src/fields/slug/SlugComponent.tsx
@@ -21,6 +21,7 @@ export const SlugComponent = ({
   const { label } = field
 
   const { value, setValue } = useField<string>({ path: path || field.name })
+  const { value: currentSlug } = useField<string>({ path: 'slug' })
 
   // Get the current value of the title field
   const targetFieldValue = useFormFields(([fields]) => {
@@ -30,13 +31,13 @@ export const SlugComponent = ({
   const handleGenerate = useCallback(
     (e: React.MouseEvent) => {
       e.preventDefault()
-      if (targetFieldValue) {
+      const newSlug = formatSlug(targetFieldValue)
+      if (targetFieldValue && newSlug !== currentSlug) {
         setValue(formatSlug(targetFieldValue))
       }
     },
-    [targetFieldValue, setValue],
+    [targetFieldValue, currentSlug, setValue],
   )
-
   const readOnly = readOnlyFromProps || false
 
   return (
@@ -44,7 +45,7 @@ export const SlugComponent = ({
       <div className="label-wrapper">
         <FieldLabel htmlFor={`field-${path}`} label={label} />
 
-        <Button className="lock-button" buttonStyle="icon-label" onClick={handleGenerate}>
+        <Button className="generate-button" buttonStyle="icon-label" onClick={handleGenerate}>
           <RefreshCw className="w-4" />
         </Button>
       </div>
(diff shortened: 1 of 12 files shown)


src/fields/slug/SlugComponent.tsx
```diff
@@ -1,79 +1,51 @@
...
+  const readOnly = readOnlyFromProps || false
 
   return (
     <div className="field-type slug-field-component">
       <div className="label-wrapper">
         <FieldLabel htmlFor={`field-${path}`} label={label} />
 
-        <Button className="lock-button" buttonStyle="none" onClick={handleLock}>
-          {checkboxValue ? 'Unlock' : 'Lock'}
+        <Button className="lock-button" buttonStyle="icon-label" onClick={handleGenerate}>
```
busbyk: Nit: `lock-button` naming could be updated

src/fields/slug/SlugComponent.tsx
```diff
@@ -1,79 +1,51 @@
...
+  const handleGenerate = useCallback(
     (e: React.MouseEvent) => {
       e.preventDefault()
-
-      dispatchFields({
-        type: 'UPDATE',
-        path: checkboxFieldPath,
-        value: !checkboxValue,
-      })
+      if (targetFieldValue) {
```
busbyk: It would be sweet to check if the new slug is going to be exactly the same as the current slug and only setValue if they're different. That way a draft wouldn't be saved if you click the generate button when the title hasn't changed. 

```
const { value: curSlug } = useField<string>({ path: 'slug' })
...
const handleGenerate = useCallback(
  (e: React.MouseEvent) => {
    e.preventDefault()
    const newSlug = formatSlug(targetFieldValue)
    if (newSlug !== curSlug) {
      setValue(formatSlug(targetFieldValue))
    }
  },
  [targetFieldValue, curSlug, setValue],
)
```

busbyk (approved):
Looks great!

Just one improvement suggested. 

I also pushed one commit to update the validation message for ensureUniqueSlug because we are using "avalanche center" as the label for tenants in user facing scenarios. Not part of your changes but figured it made sense to change in this PR.

src/fields/slug/SlugComponent.tsx
rchlfryn: Agreed! Fixed

PR #373, opened 2025-08-04
Sitemaps & Page/Post Revalidation

### LINKED ISSUE #50: Publish Updated Sitemaps

busbyk:
Resolves #50 

The intent was to have this be a focused PR on sitemaps but I realized that we needed to be revalidating pages and posts correctly for our sitemaps to be correctly generated so this evolved into a larger PR that implements correct revalidation of pages, posts, and the navigation in a tenant-aware way. 

## Key changes
- Removes static sitemap generation in `postbuild` for root domain sitemap and robots.txt in favor of [Metadata Files API files](https://nextjs.org/docs/app/api-reference/file-conventions/metadata) - sitemap.ts and robots.ts. Simpler and more visible in the filesystem IMO. They're still cached so no performance hit (they're tiny anyways). Keeps next-sitemap for it's util functions. 
- Tenant scoped robots.txt and sitemap.xml
  - Adds `robots.txt/route.ts` that disallows all `/admin` paths and links to index sitemap
  - Adds index sitemap
  - Updates pages sitemap to be tenant scoped
  - Updates posts sitemap to be tenant scoped
- Navigation & pages
  - Moves top level nav items into a cached `getCachedTopLevelNavItems` function which can be revalidated when pages/posts change (also for efficiency)
  - Enforces canonical urls for pages. Previously pages were accessible at their nav-nested url (like `/about/about-us`) and their root url (like `/about`). Now the root url will redirect to the nav-nested url and the sitemap dedupes urls included preferring the canonical, nav-nested url. 
- Revalidates pages by tenant
  - On content changes
  - On title / slug changes (revalidates old slugs correctly as well)
  - On un-publish
  - On publish/re-publish
  - On delete
- Revalidates posts by tenant
- Revalidates navigation by tanant

## Notes
- `changefreq` and `priority` are ignored by Google so I'm excluding them
- Blog index page is dynamic so does not need to be revalidated when posts change
- Page and post revalidation functions revalidate the path prior to the middleware rewrites and the rewritten paths. This is because of a bug I was running into where unpublished pages would still 404 after being re-published. This is definitely unexpected behavior to me. I wouldn't be opposed to a paired debugging sesh to see if we can figure out exactly what's going on there. I tried disabling caching in the middleware via the `x-middleware-cache` header without success. 
- I ran into a very strange bug where after getting a 404 on a page I just deleted (testing the delete revalidation) I would then get a root 404 page for {center}.localhost:3000/. I tried many debugging steps and finally tried to recreate in the preview env and could not. So this appears to be a local-only bug. Potentially due to the experimental flags we need to run a production build locally or the difference in how cached pages are served? 
- Had to use API routes for the tenant-scoped sitemaps and robots.txt because the Metadata Files weren't being recognized correctly and I wanted to keep the pages and posts sitemaps separate and use an index sitemap. There's no obvious downside to this approach IMO. 

## Testing

I had Claude beef up my ugly personal notes I used when testing this myself in case they'd be helpful for reviewing: 

Initial Setup Verification

- Check sitemap index: Visit /{center}/sitemap.xml and verify it returns XML with links to:
  - /{center}/pages-sitemap.xml
  - /{center}/posts-sitemap.xml
- Check robots.txt: Visit /{center}/robots.txt and verify it contains:
  - User-agent: *
  - Allow: /
  - Disallow: /admin
  - Sitemap: {domain}/{center}/sitemap.xml
- Check global sitemap: Visit /sitemap.xml (should contain basic site entry)
- Check global robots: Visit /robots.txt (should contain basic rules)

Page Content & Publishing Tests

- Change page content and republish
  - Update content on an existing published page
  - Republish the page
  - Expected: Page sitemap should update lastmod timestamp
  - Check: /{center}/pages-sitemap.xml shows new timestamp
- Change page name & slug and republish
  - Rename a page and change its slug
  - Republish the page
  - Expected: Navigation updates, new route accessible, old route returns 404
  - Check: Pages sitemap reflects new slug, navigation shows new name
- Unpublish then republish page
  - Unpublish an existing page
  - Expected: Page removed from sitemap, route returns 404
  - Check: /{center}/pages-sitemap.xml excludes the page
  - Republish the page
  - Expected: Page returns to sitemap and is accessible again
- Delete page
  - Delete a published page
  - Expected: Page no longer accessible, removed from sitemap
  - Check: Route returns 404, absent from pages sitemap

Post Content & Publishing Tests

- Change post content and republish
  - Update content on an existing published post
  - Republish the post
  - Expected: Post sitemap updates lastmod timestamp
  - Check: /{center}/posts-sitemap.xml shows new timestamp
- Change post name & slug and republish
  - Rename a post and change its slug
  - Republish the post
  - Expected: New route accessible at /blog/{new-slug}, old route returns 404
  - Check: Posts sitemap reflects new slug, blog index shows updated post
- Unpublish then republish post
  - Unpublish an existing post
  - Expected: Post removed from sitemap, route returns 404
  - Check: /{center}/posts-sitemap.xml excludes the post
  - Republish the post
  - Expected: Post returns to sitemap and is accessible again
- Delete post
  - Delete a published post
  - Expected: Post no longer accessible, removed from sitemap
  - Check: Route returns 404, absent from posts sitemap

Navigation & Canonical URL Tests

- Pages accessible only at canonical URLs
  - Test that pages appear in sitemap only once with their canonical navigation URL
  - Expected: No duplicate entries in pages sitemap for pages that appear in navigation
- Navigation changes
  - Modify navigation structure (add/remove/reorder items)
  - Expected: Pages sitemap reflects canonical URLs from navigation
  - Check: Navigation-linked pages use navigation URLs, not direct page slugs

Content Management Tests

- Duplicate page to tenant
  - Duplicate an existing page to the current tenant
  - Publish the duplicated page
  - Expected: New page appears in pages sitemap with correct tenant-specific URL
- Create new page
  - Create and publish a completely new page
  - Expected: Page appears in pages sitemap
  - Check: Page accessible at its slug, included in sitemap with current timestamp

Multi-tenant Verification

- Tenant isolation
  - Verify sitemaps only include content for the specific center/tenant
  - Check: URLs in sitemap match the center parameter
  - Check: Content from other tenants is excluded


src/app/(frontend)/[center]/(sitemaps)/pages-sitemap.xml/route.ts
```diff
@@ -1,60 +1,102 @@
...
+        pagesRes.docs || [],
+      )
+
+      const navigationSitemap: SitemapField[] = canonicalUrls.map((url) => {
+        const pageSlug = url.split('/').filter(Boolean).pop()
+        const matchingPage = pagesRes.docs?.find((page) => page.slug === pageSlug)
+
+        return {
+          loc: `${serverURL}${url}`,
+          lastmod: matchingPage?.updatedAt || dateFallback,
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/(sitemaps)/pages-sitemap.xml/route.ts:60]: What does it mean if we could not find the page for a slug? Is it appropriate to catastrophically error for this? Will it be caught somewhere in test?

stevekuznetsov (commented):
Let's see if we can find some time to pair on the bug you mention in your post?

src/app/(frontend)/[center]/(sitemaps)/pages-sitemap.xml/route.ts
Comment (busbyk) [reply src/app/(frontend)/[center]/(sitemaps)/pages-sitemap.xml/route.ts:60]: In this case the matchingPage is the "root" page instance (as opposed to the nav-nested page instance). The canonicalUrls returned from getCanonicalUrlsFromNavigation is just a string of urls so I'm getting the updatedAt from the root page. So if the page wasn't found it would just use the dateFallback - I don't think it's appropriate to catastrophically error for this. 

But I should probably just return the updatedAt with the canonicalUrls. I think I was planning to use this function elsewhere and just wanted the urls but I'll just see about updating to avoid this.

src/collections/Posts/hooks/revalidatePost.ts
```diff
@@ -4,41 +4,79 @@ import { revalidatePath, revalidateTag } from 'next/cache'
...
+        collection: 'tenants',
+        id: tenant,
+        depth: 0,
+      })
+    }
+
     if (doc._status === 'published') {
-      // TODO: revalidate path based on tenant
-      const path = `/blog/${doc.slug}`
+      const preRewritePath = `/blog/${doc.slug}`
```
yuliadub: this code block repeats 3 times with some param differences, but I assume there may be a reason you didnt put it into a function?

src/components/Header/utils.ts
```diff
@@ -373,4 +387,72 @@ export const getTopLevelNavItems = async ({
     ...topLevelNavItem({ tab: navigation.about, label: 'About' }),
     ...topLevelNavItem({ tab: navigation.support, label: 'Support' }),
   ]
+
+  let donateNavItem: TopLevelNavItem | undefined = undefined
+
+  if (navigation.donate?.link) {
```
yuliadub, line 393: is this way because the donate link may not be part of payload? (just wondering why donate nav is special)

src/components/Header/utils.ts
```diff
@@ -373,4 +387,72 @@ export const getTopLevelNavItems = async ({
...
+    const { topLevelNavItems } = await getCachedTopLevelNavItems(center)()
+    const navigationItem = findNavigationItemBySlug(topLevelNavItems, slug)
+
+    if (navigationItem?.link?.type === 'internal') {
+      return navigationItem.link.url
+    }
+
+    return null
+  } catch (_error) {
+    return null
```
yuliadub, line 460: lets log the error here? or leave a comment why we dont need to log it

src/collections/Pages/hooks/revalidatePage.ts
```diff
@@ -10,55 +11,124 @@ export const revalidatePage: CollectionAfterChangeHook<Page> = async ({
...
-      payload.logger.info(`Revalidating page at path: ${path}`)
+      try {
+        const { topLevelNavItems } = await getCachedTopLevelNavItems(tenant.slug)()
+        const navigationPaths = getNavigationPathForSlug(topLevelNavItems, doc.slug)
 
-      revalidatePath(path)
-      revalidateTag('pages-sitemap')
-    }
+        // Include both the direct paths and the navigation paths
+        const allPaths = [...basePaths, ...navigationPaths.map((path) => `/${tenant.slug}${path}`)]
```
yuliadub: would $path always start with a '/'?

src/components/Header/utils.ts
busbyk: I kept this separate because the donate nav link get's rendered as a button instead of like the rest of the nav links. 

This seems trivial but I would have had to add to the TopLevelNavItem type to include something to indicate what a nav link should render as (i.e. either a link or a button). But the donate button is the only element that will ever be a button so it seemed silly to add that parameter just for one item. It seemed easier and more logical to render it separately.

src/collections/Posts/hooks/revalidatePost.ts
busbyk: No particular reason really. Just simple enough to duplicate. But I can throw into a function.

src/collections/Pages/hooks/revalidatePage.ts
busbyk: Yes, getNavigationPathForSlug should return paths that start with "/" based on the getTopLevelNavItems function but hmm it might be worth writing a utility function to normalize paths. Could be used elsewhere too. That would be nice. I'll do that.

busbyk: Oh nice, just need to modify this a little: https://github.com/NWACus/web/blob/sitemaps/src/services/utils.ts#L2. I thought I'd already written this!

src/app/(frontend)/[center]/(sitemaps)/pages-sitemap.xml/route.ts
Comment (busbyk) [reply src/app/(frontend)/[center]/(sitemaps)/pages-sitemap.xml/route.ts:60]: Oh I see why I did it this way haha. I wanted to use the topLevelNavItems to make sure the sitemap accurately reflects the navigation. The topLevelNavItems only includes the url, not the reference from payload. I could update it to include that but I like it just including the bare minimum it needs to render. I think it's fine to populate the updatedAt like I'm doing here because there _should_ always be a "root" page for any page in the navigation anyways.

src/components/Header/utils.ts
yuliadub: makes sense to me! thanks for explaining

yuliadub (approved):
did not dig into the issues you mentioned in the PR description- but also didnt see anything that jumped out to me for root causes of those. aside from that this looks good to me!

src/utilities/resolveTenant.ts
rchlfryn, line 1: 👍

rchlfryn (commented):
Are we automatically updating the sitemap on when the page/post content is changed or published or both?
My biggest concern is how this works with autosave.  Will the sitemap regenerate every time a draft is saved?

busbyk:
> Are we automatically updating the sitemap on when the page/post content is changed or published or both? My biggest concern is how this works with autosave. Will the sitemap regenerate every time a draft is saved?

Updated to handle autosave scenarios in #375

PR #375, opened 2025-08-05
Revalidate everything

### LINKED ISSUE #52: Ensure Next.js cache revalidation is done as an afterChange hook on every collection

busbyk:
Resolves #52 

This triggers path revalidation based on blocks with relationship fields or relationship fields (including upload fields) changing. So for example: if a biographies document changes and that biographies document is in a post and in a page, both the post's and page's paths will be revalidated. 

When a biographies document changes the revalidation hook runs two re-usable revalidate functions: 1) to find instances of this biography in biography blocks in pages and posts and 2) to find instances of this biography in relationship fields (relationTo = biographies in this case) in pages and posts. 

The relationship fields and block fields are identified dynamically but only the pages and posts collections have paths that need to be revalidated at the moment. 

When other collections have paths we'll need to add to the `findDocumentsWithBlockReferences`, `findDocumentsWithRelationshipReferences`, and `revalidateDocument` functions. 

## Key changes
- The following collections now run `revalidateBlockReferences` and `revalidateRelationshipReferences`:
  - Biographies
  - Media
  - Settings
  - Tags
  - Teams
  - Forms
  - Posts (for related posts relationships and future post-related block types)
- The Settings collection now revalidates all paths for the given tenant and the non-tenant scoped paths when deterministic field values change (uses same logic as the incremental seed)
- The Tenants collection now revalidates all paths for the given tenant and the non-tenant scoped paths when `name` or `customDomain` change
- Several utility functions for traversing the config to find specific fields, finding documents of those fields, and revalidating have been added


rchlfryn (approved):
Thank for fixing autosave stuff. Wasn't able to completely dig in but the demo looked good

PR #386, opened 2025-08-09
Improve multi tenant selector

### LINKED ISSUE #377: Bug: Avalanche center selection in document view changes tenant
### Description
If a super admin selects a new tenant while in a document view (aka the detailed view of a collection item ie: page, post, individual user), it will move the item to that new tenant.

We should make the multi-tenant selector readonly on the document view, but make it editable for "global" collections like navigation

rchlfryn:
A collaborative PR...

### Description
We added a multi-tenant selector, but sometimes if you change the selector when you are in a specific view, it will change the avalanche center in the DB. this is obviously problematic.

### Issue
Fixes #377


src/payload.config.ts
```diff
@@ -58,6 +59,7 @@ export default buildConfig({
             useAsTitle: 'slug',
           },
         },
+        '@/components/CollectionTypeAction',
```
busbyk: The reason `useDocumentInfo` is still always returning an empty object is because this slot is apparently rendered outside of the provider associated with useDocumentInfo. That makes sense tbh. We need to inject this in a slot that's inside of the document info provider. 

`beforeDocumentControls` seems like the best candidate to me and I did confirm that provides the document info to the hook but we would have to add this for _every_ collection and that seems way too easy to forget to add to a new collection. 

I think the only way to add a client component to every edit view would be to edit the root edit view and I don't know that we want to do that.

src/components/TenantSelector/TenantSelector.tsx
```diff
@@ -1,14 +1,27 @@
...
 import TenantSelectorClient from './TenantSelector.client'
 
-const TenantSelector = async () => {
+const TenantSelector = async ({ params }: AdminViewServerProps) => {
+  const isGlobalDocument =
+    params?.segments?.includes('settings') || params?.segments?.includes('navigations')
+  const hideTenantSelect =
+    params?.segments?.includes('roles') ||
+    params?.segments?.includes('globalRoles') ||
+    params?.segments?.includes('globals')
```
busbyk: I wasn't getting any values for params here. I think we need to do this in a client component still. And we can improve this by inspecting the payload config vs. hardcoding collection slugs. I'll leave a comment in the TenantSelector.client.tsx component.

src/components/TenantSelector/TenantSelector.client.tsx
busbyk, line 1: I like where you're going with inspecting the params in the server component but I think we want to use a combination of the params, viewType, and payload config. 

If I add this in this component: 
```
  const { config } = useConfig()
  const pathname = usePathname()
  const params = useParams()

  console.log(`pathname: ${pathname}`)
  console.log(`params: ${JSON.stringify(params)}`)
  console.log(`viewType: ${viewType}`)
```

We can determine if we're working with a global or a collection (useParams), whether we're in the list view or document view (useViewType), and if the collection is a "global collection" (our own concept which is indicated by the collection having a tenant field with unique: true on it). 

CleanShot 2025-08-10 at 11 10 50@2x


So I think the approach here is to use `useParams` to determine if we're in a global or collections view by checking the first param in the array. 

If global, then show the tenant selector and stop evaluating. 

If collection, check if we're in the document view or list view. If list view, stop evaluating. If document view, check if it's a "global collection" by looking for the collection slug (second value in the param array if we're in the document view) in the payload config and determine if that field has unique:true. `config.collections.find(collectionConfig => collectionConfig.slug === slugFromParams).fields.find(fieldConfig => fieldConfig.slug === 'tenant').unique`.

If global collection then we still want to show it but if it's a collection with a tenant field without unique:true we wan the field to be readOnly.

src/payload.config.ts
rchlfryn: I will remove these files and go with the params approach above

src/components/TenantSelector/TenantSelector.tsx
rchlfryn: I was seeing params here but just moved everything to the client component as you suggested

src/components/TenantSelector/TenantSelector.client.tsx
rchlfryn: Update the check as you suggested in [f2c4ce3](https://github.com/NWACus/web/pull/386/commits/f2c4ce317254021758f81af38fb5df69e6fb7e85)

src/components/TenantSelector/TenantSelector.client.tsx
```diff
@@ -23,10 +36,21 @@ const TenantSelectorClient = ({ label }: { label: string }) => {
     [setTenant],
   )
 
+  // hide for Global RoleRoles, NAC widget & Diagnostics
+  if (
+    isGlobal ||
+    params?.segments?.includes('roles') ||
+    params?.segments?.includes('globalRoles') ||
+    params?.segments?.includes('globalRoleAssignments')
```
rchlfryn: @busbyk We still want to hide the selector for Roles, Global Roles, and Global Role assignments. Since these are collections, I did not see a better way to check for these 3 other than this. Thoughts?

busbyk: Ah yes good call. We want to hide the selector for collections without a tenant field. This should work:

```
const isGlobal = params.segments && params.segments[0] === 'globals'
const isCollection = params.segments && params.segments[0] === 'collections'
const slugFromParams = params.segments && params.segments[1]
const collectionConfig = config.collections.find(
  (collectionConfig) => collectionConfig.slug === slugFromParams,
)
const collectionTenantFieldConfig = collectionConfig?.fields.find(
  (fieldConfig) => 'name' in fieldConfig && fieldConfig.name === 'tenant',
)
const isUnique = !!collectionTenantFieldConfig?.unique

...

// hide globals and collections without a tenant field
if (isGlobal || !collectionTenantFieldConfig) return null
```

busbyk (changes requested):
Ah I found a problem without plan: if the tenant selector is cleared / no tenant is selected the create page doesn't render. 

I get this error in the terminal:
```
[16:44:24] ERROR (payload): You must select a tenant
    err: {
      "type": "APIError",
      "message": "You must select a tenant",
      "stack":
          APIError: You must select a tenant
              at hooks.beforeChange (webpack-internal:///(rsc)/./src/fields/tenantField/index.ts:39:31)
              at promise (file:///Users/kellenbusby/Code/nwac/web/node_modules/.pnpm/payload@3.43.0_graphql@16.10.0_typescript@5.7.2/node_modules/payload/dist/fields/hooks/beforeChange/promise.js:56:43)
              at file:///Users/kellenbusby/Code/nwac/web/node_modules/.pnpm/payload@3.43.0_graphql@16.10.0_typescript@5.7.2/node_modules/payload/dist/fields/hooks/beforeChange/traverseFields.js:12:23
              at traverseFields (file:///Users/kellenbusby/Code/nwac/web/node_modules/.pnpm/payload@3.43.0_graphql@16.10.0_typescript@5.7.2/node_modules/payload/dist/fields/hooks/beforeChange/traverseFields.js:11:12)
              at beforeChange (file:///Users/kellenbusby/Code/nwac/web/node_modules/.pnpm/payload@3.43.0_graphql@16.10.0_typescript@5.7.2/node_modules/payload/dist/fields/hooks/beforeChange/index.js:15:11)
              at createOperation (file:///Users/kellenbusby/Code/nwac/web/node_modules/.pnpm/payload@3.43.0_graphql@16.10.0_typescript@5.7.2/node_modules/payload/dist/collections/operations/create.js:133:41)
              at runNextTicks (node:internal/process/task_queues:65:5)
              at process.processImmediate (node:internal/timers:459:9)
              at process.callbackTrampoline (node:internal/async_hooks:130:17)
              at async renderDocument (webpack-internal:///(rsc)/./node_modules/.pnpm/@payloadcms+next@3.43.0_@types+react@19.0.1_graphql@16.10.0_monaco-editor@0.52.2_next@1_030db2ea9ee22b1902f6a2a07dd7e0f3/node_modules/@payloadcms/next/dist/views/Document/index.js:234:11)
              at async Document (webpack-internal:///(rsc)/./node_modules/.pnpm/@payloadcms+next@3.43.0_@types+react@19.0.1_graphql@16.10.0_monaco-editor@0.52.2_next@1_030db2ea9ee22b1902f6a2a07dd7e0f3/node_modules/@payloadcms/next/dist/views/Document/index.js:323:9)
      "data": null,
      "isOperational": true,
      "isPublic": true,
      "status": 400,
      "name": "APIError"
    }
```

CleanShot 2025-08-14 at 16 41 45

It looks like clicking create actually immediately creates a draft document and that request is failing because no tenant is selected. 

I'm actually not sure of the best approach here. I suppose we could add exceptions to both the tenant selector component and the tenant field validation for the create scenario when there is no tenant already selected.

Or maybe we could try defaulting the tenant selector to the default tenant (i.e. first one) if in create for a collection and no value is set. If that worked that would probably be easier? But the API request to create the draft document may have already errored in that scenario so that might not work. 

Happy to discuss further.

rchlfryn:
> Ah I found a problem without plan: if the tenant selector is cleared / no tenant is selected the create page doesn't render.

Ahhh great catch. Solution: we will hide the create button if no tenant is selected. Updated access in [0ff5962](https://github.com/NWACus/web/pull/386/commits/0ff5962b7c558e2ce58aeae53a778f7bd6d320da)

> Ah yes good call. We want to hide the selector for collections without a tenant field. This should work:

Thanks for the help on this. Updated in [6e69a2b](https://github.com/NWACus/web/pull/386/commits/6e69a2b974f5b4c6703bffc0464e9bcb2207aacc)

Ready for re-review

busbyk (commented):
Oo I was surprised that you'd only added create access tenant check for pages and posts but then I realized that it was probably because you noticed that the tenant field required error was only being thrown on collections with versions enabled because create creates a draft whereas create for collections without versions doesn't. 

But the behavior for collections without versions is that the tenant field gets set to the first tenant in the tenant options and then the user can't change it. 

Should we disable create unless a tenant is selected for all collections with a tenant field or should we just educate multi tenant admins that they need to select a tenant before creating and if they click create without a tenant selected we assume they'll just figure it out? 

I think I'd argue for consistent behavior and disable create unless a tenant is selected for _all_ collections with a tenant field. 

CleanShot 2025-08-15 at 09 53 24

rchlfryn:
> I think I'd argue for consistent behavior and disable create unless a tenant is selected for _all_ collections with a tenant field.

I think it makes sense to extend it to all collections. Because of this, I did place the tenant check in `accessByTenantRole`. I could se this not being the best place for this but it seemed silly to create a different function when it is only being used by collection we want to hide the create button from anyways.

Form submissions is the only collection that is still wonky (especially with turning on form submission creation #387 ) I checked [payload source code](https://github.com/payloadcms/payload/blob/main/packages/plugin-form-builder/src/collections/FormSubmissions/index.ts#L93) and we have the same permissions, it just breaks the multi-tenant selector.  Thoughts on how we should handle this?

busbyk:
> > I think I'd argue for consistent behavior and disable create unless a tenant is selected for _all_ collections with a tenant field.
> 
> I think it makes sense to extend it to all collections. Because of this, I did place the tenant check in `accessByTenantRole`. I could se this not being the best place for this but it seemed silly to create a different function when it is only being used by collection we want to hide the create button from anyways.
> 
> Form submissions is the only collection that is still wonky (especially with turning on form submission creation #387 ) I checked [payload source code](https://github.com/payloadcms/payload/blob/main/packages/plugin-form-builder/src/collections/FormSubmissions/index.ts#L93) and we have the same permissions, it just breaks the multi-tenant selector. Thoughts on how we should handle this?

Check if they're in the admin panel?

```
create: ({ req }) => {
  const isAdminPanel = req.pathname.startsWith('/admin')
  return !isAdminPanel
}, // world creatable outside of the admin panel
```

Could get more specific if we wanted: `req.routeParams?.collection === 'form-submissions'`

rchlfryn:
@busbyk awesome!

`req.routeParams?.collection === 'form-submissions'` blocks the front end submission if you are logged in and submit the form, but `req.pathname.startsWith('/admin')` works the way we want it to

stevekuznetsov:
What was the mechanism by which the tenant selector would touch the DB? I see the logic for `readOnly` on the input, but still am confused on how this was threading anything through to the backend?

busbyk:
> What was the mechanism by which the tenant selector would touch the DB? I see the logic for `readOnly` on the input, but still am confused on how this was threading anything through to the backend?

The tenant selector changes the value of a hidden field in the document view. Every tenant-scoped collection has a tenant field and it matches the value of the tenant selector. Global collections (i.e. tenant has unique = true) behave differently because of the multi tenant plugin's global redirect component. But on regular collections, changing the tenant selector changes the tenant for that collection. 

We just felt like this was a confusing UX and was bound to lead to mistakes where a user was just trying to switch to a different tenant, not change the tenant for the current document.

busbyk (approved):
🙌

PR #387, opened 2025-08-09
Allow everyone to submit a form

### LINKED ISSUE #274: Forms throw 500 error - invalid field Avalanche Center
This is likely due to the form submissions needing an avalanche center/tenant and the form block isn't configured to determine one from the domain right now. 

Image

We should also pull in recent changes from the website template. The `@ts-expect-error` annotations appear to have mostly been fixed. 

I believe our first use case for form submissions is events so setting the milestone to start of 25/26 season.

rchlfryn:
### Description
Previously, we did not allow a non logged in user to submit a form. This updates the access to allow anyone to submit a form. It was throwing a 500, but it was really a 403.

**How to test**: 
* Log out of admin view
* Navigate to `/contact`
* Submit the form
* Should see success 

### Issues
Fixes #274


busbyk (changes requested):
This fixes the issue with needing create permission for form-submissions but I'm still getting a 500 (400 internally) related to the tenant field missing: 
CleanShot 2025-08-10 at 09 33 39@2x

I also had a few thoughts while considering forms in general: 
- I think at least `update` should be `() => false` Since these are user-submitted entries it doesn't really make sense to ever modify them right? They're intended to be a record of the original submission. 
- We may want to restrict delete as well but maybe using byTenant() for delete makes sense since a tenant admin may want to delete an inappropriate or otherwise objectionable submission. 
- With public create we do open ourselves up to additional attack vectors where malicious actors could submit unlimited form submissions. We should consider adding a CAPTCHA (there are really good invisible CAPTCHAs these days like hCaptcha and Cloudflare Turnstile (my preferred captcha). We could also use a honeypot field (although those aren't very effective these days). Rate limiting and ensuring we have reasonable value limits on all input types would also be good to consider. I think this is probably outside of the scope of this PR but likely good to do before the start of the season.

rchlfryn:
> This fixes the issue with needing create permission for form-submissions but I'm still getting a 500 (400 internally) related to the tenant field missing:

@busbyk oh very strange. This not happening locally so I didn't see it. A good reminder to test in dev after deploying. 

I will update the permission settings and create an issue for the reCAPTCHA since I am still not convinced these forms will be used.

rchlfryn:
@busbyk As we chatted about, the 500 was being caused by from email. This was update in [3b08ac0](https://github.com/NWACus/web/pull/387/commits/3b08ac00489867d242e0b7dc47a599c20e6716de). 

Updated access in [ad4f61b](https://github.com/NWACus/web/pull/387/commits/ad4f61bc01d7cab5c1b75e00d5626a6bcd69628f) and decided to restrict delete

Ready for re-review

src/access/byTenantRole.ts
```diff
@@ -63,14 +63,14 @@ export const accessByTenantRole: (collection: ruleCollection) => CollectionConfi
...
   collection: ruleCollection,
-) => CollectionConfig['access'] = (collection: ruleCollection) => {
+) => {
   return {
     create: () => true, // world creatable
     read: byTenantRole('read', collection),
-    update: byTenantRole('update', collection),
-    delete: byTenantRole('delete', collection),
+    update: () => false,
+    delete: () => false,
```
busbyk: I wasn't very decisive in my comment but I do think we should use byTenantRole for delete for the reason I mentioned: an admin may want to delete inappropriate submissions. Thoughts?

> We may want to restrict delete as well but maybe using byTenant() for delete makes sense since a tenant admin may want to delete an inappropriate or otherwise objectionable submission.

src/access/byTenantRole.ts
```diff
@@ -63,6 +63,17 @@ export const accessByTenantRole: (collection: ruleCollection) => CollectionConfi
   }
 }
 
+export const accessForFormSubmission: (collection: ruleCollection) => CollectionConfig['access'] = (
+  collection: ruleCollection,
+) => {
+  return {
```
busbyk: Nit: Since this will only ever be used by form submissions should we just write this access directly in the formSubmissionOverrides in src/plugins/index.ts?

src/access/byTenantRole.ts
rchlfryn: I like it! I will move it with the updates above

src/access/byTenantRole.ts
rchlfryn: I like it - moved in [8a6f6c2](https://github.com/NWACus/web/pull/387/commits/8a6f6c2034ee6f8063b08d9b6c41b9b9b7dda3d7)

src/access/byTenantRole.ts
rchlfryn: I think you are right - we should keep a permission check for deletion in case an admin wants to remove submissions

busbyk (approved):

PR #393, opened 2025-08-13
Payload fixes/enhancements + workflow enhancements

busbyk:
This PR came about because of confusing errors in production and then took me down some rabbit holes of reevaluating the way we're doing a few things. This is a bit of a potluck PR but the goal is to get production and dev errors resolved by correctly using Vercel env vars (the edge config connection string wasn't set and was resulting in some errors during build for example). 

## Key changes

Environment variable handling improvements:
  - Removed env vars stored in GitHub workflows _and_ Vercel in favor of Vercel environment variables except for sensitive environment variables (i.e. marked as sensitive in Vercel). These are stored as secrets in GitHub in addition to Vercel. Vercel env vars are used properly now by using `vercel env pull` to create a `.env` file in addition to `vercel pull` to set up the vercel project for use by `vercel build`.
  - Added GitHub environments for varying PAYLOAD_SECRET values per environment.
  - Updated .gitignore to ignore all .env files

 Standalone seed script creation:
  - Minor enhancement which allows us to run the seed script without starting up a Next.js dev server and creating a bootstrap user first. 
  - Created src/scripts/standalone-seed.ts and associated package.json script `seed:standalone`
  - Updated CI workflows to use standalone seeding
  - Modified seed script to handle there not being a user during standalone seeding

Diagnostics global:
  - This doesn't show very much at the moment but we can add more here if we would like. JSON.stringify-ing the whole config is way too much so I picked only a couple attributes I want to know about in production. 
  - Added new Diagnostics global with a single UI field which displays a few relevant attributes from the config and the current serverURL
  - Restricted access to super admins only (i.e. users with `*` action on `*` collections)
CleanShot 2025-08-12 at 22 30 09@2x

Payload logging:
  - Modified Pino logger to include stack traces (figured this was a nicer solution that patching something RE https://github.com/payloadcms/payload/issues/10957)


package.json
```diff
@@ -80,7 +80,7 @@
     "import-in-the-middle": "^1.14.2",
     "json-stable-stringify": "^1.3.0",
     "libsql": "^0.5.6",
-    "lodash.merge": "^4.6.2",
+    "lodash-es": "^4.17.21",
```
stevekuznetsov, line 84: how big of a change in bundle?

.github/workflows/preview.yaml
```diff
@@ -80,32 +80,14 @@ jobs:
...
-          DATABASE_URI: '${{ steps.connection.outputs.uri }}'
-          DATABASE_AUTH_TOKEN: '${{ steps.connection.outputs.token }}'
-          PAYLOAD_SEED_PASSWORD: '${{ secrets.PAYLOAD_SEED_PASSWORD }}'
-          PAYLOAD_SECRET: '${{ secrets.PAYLOAD_SECRET }}'
-          NEXT_PUBLIC_GOOGLE_MAPS_API_KEY: '${{ secrets.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY }}'
-          VERCEL_BLOB_READ_WRITE_TOKEN: '${{ secrets.VERCEL_BLOB_READ_WRITE_TOKEN }}'
-          NEXT_PUBLIC_ROOT_DOMAIN: '${{ steps.domains.outputs.preview_alias }}'
-          NEXT_PUBLIC_POSTHOG_KEY: '${{ secrets.NEXT_PUBLIC_POSTHOG_KEY }}'
-          NEXT_PUBLIC_POSTHOG_HOST: '${{ secrets.NEXT_PUBLIC_POSTHOG_HOST }}'
+      - name: Pull Vercel Environment Information
```
stevekuznetsov, line 93: Why twice?

busbyk: They're different commands actually. `vercel pull` sets up a local Vercel environment/project settings with files like `.vercel/.env.preview.local` and `.vercel/project.json`. To load Vercel env vars into a `.env` file you need to use `vercel env pull`. 

So I'm using `vercel env pull` to get all env vars not marked as sensitive in Vercel and storing them in a `.env` file so things like the migrate and seed scripts will have access to them. But since we're using `vercel build` we still need the `vercel pull` as well.

package.json
busbyk: It should be the same. Lodash no longer suggests the [per method packages](https://lodash.com/per-method-packages) and `lodash-es` should be effectively tree-shaken.

.github/workflows/preview.yaml
stevekuznetsov: :exploding_head:

.gitignore
```diff
@@ -20,6 +20,7 @@ dev.db-wal
 .env.development
 tsconfig.tsbuildinfo
 .env*.local
+.env.*
```
stevekuznetsov, line 22: add in the two explicit opt-ins 

```
!.env.example
!.env.test
```

busbyk: Good call

PR #395, opened 2025-08-14
Dynamic OG images + Correct metadata for tenant-scoped routes

### LINKED ISSUE #191: Design OpenGraph Preview Images & Copy For Existing Pages
For the dynamic, center-driven content pages we allow users of the admin panel to configure the OpenGraph images, copy, etc when they are creating the page. However, for the common pages and navigation sections, we likely want to spend the time to carefully design the content as this is such an important part of the web presence for the site.

busbyk:
Resolves #191 

## Key changes

- Serves dynamic opengraph images for all routes that include the tenant banner, usfs logo if exists, title, and description
- Updates all metadata for each center-scoped route. There were several tags missing and we weren't merging with the parent metadata before. 
- Uses the center's favicon on center-scoped routes

## Screenshots

OG preview:
CleanShot 2025-08-13 at 17 47 25

OG image (for the NWAC blog page in this case):
CleanShot 2025-08-13 at 17 48 57

OG image (for the SAC landing page):
CleanShot 2025-08-13 at 17 50 32

Center-scoped favicons
CleanShot 2025-08-13 at 17 49 27

## Caveats

- I really wanted to get these dynamic OG images to include the page title/description and post title/description but for some reason Next.js won't pick up the `opengraph-image.tsx`/`twitter-image.tsx` files when they're nested in the `/blog/[slug]` and `[slug]`/`[...segments]` dynamic route segments. We may need to transition to a pattern where we have a single `/api/og` endpoint that accepts query params to generate the og image and we set our metadata to use that instead of the "special" files. But I've already spent too much time on this so that's a future enhancement. We can do that when we do #63 to hopefully include things like the actual forecast. 
CleanShot 2025-08-13 at 17 51 26
- We can't use css variables in `vercel/og` which is `santori` under the hood so I had to copy some of the theme colors to use for the og images. Not ideal but we'll already have to make code changes for theme colors for new tenants anyways so I think this is acceptable for now. 
- The build was failing on the parent metadata being typed as `Metadata` and not `ResolvedMetadata` but then I get tons of ts mismatch errors when spreading that into the return value of `generateMetadata` so I'm casting that to `Metadata`. Honestly, I'm just being time/budget conscious here.


busbyk:
@rustynwac this might also resolve #63. Or at least partially. Let's touch base on this to see if so.

src/utilities/convertWebpToPng.ts
```diff
@@ -0,0 +1,37 @@
...
+    const webpBuffer = await response.arrayBuffer()
+
+    // Convert WebP to PNG using Sharp
+    const pngBuffer = await sharp(Buffer.from(webpBuffer)).png().toBuffer()
+
+    // Convert to base64 data URL for use in ImageResponse
+    const base64 = pngBuffer.toString('base64')
+    return `data:image/png;base64,${base64}`
+  } catch (error) {
+    console.error('Error converting WebP to PNG:', error)
```
stevekuznetsov, line 33: should this be the pino logger?

src/utilities/generateOGImage.tsx
```diff
@@ -0,0 +1,225 @@
...
+import { resolveTenant } from '@/utilities/resolveTenant'
+import configPromise from '@payload-config'
+import * as Sentry from '@sentry/nextjs'
+import { ImageResponse } from '@vercel/og'
+import { readFile } from 'node:fs/promises'
+import { join } from 'node:path'
+import { getPayload } from 'payload'
+
+// Color mappings copied from colors.css since CSS variables don't work in OG images
+const centerColorMap = {
```
stevekuznetsov, line 12: If we have an onboarding meta-tracker, adding a link to this would be good

src/utilities/convertWebpToPng.ts
busbyk: Yanno, I haven't always been using the `payload.logger`. I don't think we really have a pattern for when to use it and when not to at the moment.

src/utilities/generateOGImage.tsx
```diff
@@ -0,0 +1,225 @@
...
+    if (
+      settings.banner &&
+      typeof settings.banner !== 'number' &&
+      settings.tenant &&
+      typeof settings.tenant !== 'number'
+    ) {
+      bannerImgProps = getImgAttrsFromMediaResource(settings.banner, settings.tenant)
+
+      // Convert WebP to PNG if needed
+      // TODO: remove this once .webp support lands in Satori: https://github.com/vercel/satori/pull/622
```
stevekuznetsov, line 106: can we file a follow up?

src/utilities/generateOGImage.tsx
```diff
@@ -0,0 +1,225 @@
...
+            flexDirection: 'column',
+            alignItems: 'center',
+            justifyContent: 'center',
+            background: colors.header,
+            padding: '40px',
+          }}
+        >
+          {settings?.banner && bannerImgProps && (
+            <div tw="flex justify-center mb-8 gap-2 w-fit">
+              {/* eslint-disable-next-line @next/next/no-img-element */}
```
stevekuznetsov, line 153: can we leave a short comment on why nextimage is inappropriate (not rendering into html)?

src/app/(frontend)/[center]/blog/page/[pageNumber]/page.tsx
```diff
@@ -97,27 +97,30 @@ export default async function Page({ params: paramsPromise, searchParams }: Args
...
-  const { center, pageNumber } = await params
-  const tenant = await payload.find({
-    collection: 'tenants',
-    select: {
-      name: true,
+export async function generateMetadata(
+  { params }: Args,
+  parent: Promise<ResolvedMetadata>,
+): Promise<Metadata> {
+  const parentMeta = (await parent) as Metadata
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/blog/page/[pageNumber]/page.tsx:104]: can we file a follow-up for this? seems like a known issue? https://github.com/vercel/next.js/issues/59950

src/utilities/convertWebpToPng.ts
stevekuznetsov: File a follow-up to use it always? nice to have central logging on a server

stevekuznetsov:
Trade-offs seem reasonable - let's try to track them at least to know what we need to fixup later if/when we get to it

src/utilities/convertWebpToPng.ts
busbyk: https://github.com/NWACus/web/issues/398

src/utilities/generateOGImage.tsx
busbyk: Great call. I'll bring this up during check-in today too because I'd like to make a decision on documentation. I think that this particular case makes sense to go in the /docs dir in the codebase but there are some things we should document that shouldn't be in a public repo and this is a good discussion starter. 

But I will get something started in /docs as part of this PR.

src/utilities/generateOGImage.tsx
busbyk: 🫡 #399

src/app/(frontend)/[center]/blog/page/[pageNumber]/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/blog/page/[pageNumber]/page.tsx:104]: https://github.com/NWACus/web/issues/400

src/utilities/generateOGImage.tsx
busbyk: added!

busbyk: Oh you merged before I added 😅

PR #402, opened 2025-08-14
Onboarding docs

busbyk:
Super basic onboarding checklists. 

Please suggest anything I missed that you can think of.


rchlfryn (approved):
A good start - we can edit and add more as we go through this process.

busbyk:
> A good start - we can edit and add more as we go through this process.

Yep, agreed. Just a quick somethin somethin. Thanks!

PR #403, opened 2025-08-14
Check if a migration is needed in CI

busbyk:
Adds a job to our CI workflow to check if a migration should have been created for this PR. Should help us avoid forgetting to create migrations when we make schema changes.


rchlfryn (approved):
LG - do we need a check to see if a migration needs to be run? I think that is automatically done but unsure..

busbyk:
> LG - do we need a check to see if a migration needs to be run? I think that is automatically done but unsure..

Yea I was able to deploy this which should have had a migration but didn't: https://github.com/NWACus/web/pull/393

The preview workflow doesn't run migrations because it's seeding. I suppose it could but there wasn't any need. I think checking separately here is nice.

PR #416, opened 2025-08-15
Footer updates - add secondary phone number slot

### LINKED ISSUE #415: Remove phone validation in footer
The phone field in the footer is a phone field rather than text which would accommodate: 

Image

rchlfryn:
### Description
Adds another phone slot to the footer since some avalanche centers use a number for the office and one for texting forecasters. It also adds an optional label for both phone numbers.
Screenshot 2025-08-15 at 13 18 56

### Issue
Fixes #415

### DIFF
--- a/src/collections/Settings/index.ts
+++ b/src/collections/Settings/index.ts
@@ -71,11 +71,45 @@ export const Settings: CollectionConfig = {
               },
             },
             {
+              name: 'phoneLabel',
+              type: 'select',
+              options: [
+                { label: 'Phone', value: 'Phone' },
+                { label: 'Office', value: 'office' },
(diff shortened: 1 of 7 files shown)


src/collections/Settings/index.ts
```diff
@@ -70,6 +70,19 @@ export const Settings: CollectionConfig = {
                 description: 'Appears in your website footer.',
               },
             },
+            {
+              name: 'phoneLabel',
+              type: 'select',
+              options: [
+                { label: 'Phone', value: 'Phone' },
```
busbyk: I think the value should be lowercase here. Looks like it shouldn't matter but just fyi.

busbyk (approved):

Issue #26, opened 2025-03-06
User can make a gift


rustynwac:
Third party giving form is either embedded on the Donate/Membership (whatever they call it) page or there's a link on that page out to a standalone form/page hosted by a donation tool (like NWAC or Sawtooth)

busbyk:
#240 should resolve this

busbyk:
@rustynwac can you get an example from Sierra of their donation form so we can test the embed block for this in the footer and elsewhere on the website?

rustynwac:
`<script src=“https://donorbox.org/widget.js” paypalExpress=“true”></script> <iframe src=“https://donorbox.org/embed/sac-donate?” name=“donorbox” allowpaymentrequest=“allowpaymentrequest” seamless=“seamless” frameborder=“0” scrolling=“no” height=“900px” width=“100%” style=“max-width: 500px; min-width: 250px; max-height:none!important” allow=“payment”></iframe>`

busbyk:
I think we should be good here but I'll leave it up to you to close this @rustynwac

PR #417, opened 2025-08-15
Generic Embed Block

### LINKED ISSUE #240: Blocks: Generic embed block/feature
A user should be able to include third-party embeds (i.e. scripts, irames, etc.) like YouTube videos, forms, social media posts, etc. 

It would be great to develop this as a block _and_ as a [feature for the Lexical editor](https://payloadcms.com/docs/rich-text/overview#creating-your-own-custom-feature). It's possible to [use a block as a feature](https://payloadcms.com/docs/rich-text/overview#features-the-building-blocks:~:text=the%20same%20time.-,BlocksFeature,-No). 

We probably want to do some basic sanitization in the block renderer component. I've used [isomorphic-dompurify](https://www.npmjs.com/package/isomorphic-dompurify) before and allowed iframe tags. 

We may also want to sandbox embeds inside an iframe for extra security. 

If we wanted to be more restrictive we could implement a domain whitelist for embeds and instruct editors on what embeds are allowed. This would mean restricting to sites like YouTube, Instagram, our centers' CRMs, etc. This may be hard to maintain though. 

Example use cases so far:
- Donation forms
- YouTube embeds for blog posts

busbyk:
Resolves #240 

This block is intended to serve as a catch-all for embedding iframes, scripts, and custom html. Custom HTML will be sandboxed in an iframe and resized according to it's rendered height.

If the GenericEmbed block contains html with iframes or scripts, we don't sandbox them as this could cause issues with their functionality. In this case we don't mess with those elements' sizing or attributes - we expect the provider of the embed script to have set those attributes correctly. 

The block has two variants: one for layout block embeds (in places like the pages collection) and one for embedding in the RichText component. They just have different defaults for `wrapInContainer`. 

I added this block to the BlocksFeature of every Lexical editor with an existing BlocksFeature as well as the defaultLexical field. I figure we may want to allow embeds in most locations so this seems reasonable but I'd love to hear from y'all if we want to restrict where embeds are allowed. 

## Screenshots

CleanShot 2025-08-15 at 16 30 04

## Demo

https://www.loom.com/share/b97bb8658b9a47cba79b39e75ba80322?sid=dab85cb6-c655-4863-af6c-c4145dbd1f6e


src/blocks/GenericEmbed/Component.tsx
```diff
@@ -0,0 +1,172 @@
...
+        'frameborder',
+        'scrolling',
+        'sandbox',
+        'style',
+        'title',
+        'name',
+        'src',
+        'height',
+        'id',
+        'allowpaymentrequest',
```
busbyk, line 49: This one is specifically for the example embed that @rustynwac provided [here](https://github.com/NWACus/web/issues/26). We may want to have something in the website settings which allows an admin to add to this list? Or maybe we can just handle one-off cases as they pop up. It's hard to write an embed component that's not too permissive but still allows for the majority of embed scenarios. 🤷

public/thumbnail/GenericEmbedThumbnail.jpg
rchlfryn, line 1: Niiiiiiiiiice 👏

src/blocks/GenericEmbed/Component.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const [iframeHeight, setIframeHeight] = useState<number>(600)
+
+  const bgColorClass = `bg-${backgroundColor}`
+  const textColor = getTextColorFromBgColor(backgroundColor)
+  // const alignClass =
+  //   alignContent === 'center'
+  //     ? 'text-center'
+  //     : alignContent === 'right'
+  //       ? 'text-right'
+  //       : 'text-left'
```
rchlfryn: Remove this

src/components/ColorPicker/index.tsx
```diff
@@ -25,9 +25,9 @@ const ColorPicker = (props: TextFieldClientProps) => {
   const { value, setValue } = useField({ path })
 
   return (
-    <div className="flex items-start justify-between mb-6">
+    <div className="flex flex-col mb-6">
       <FieldLabel htmlFor={path} label={field.label} required={field.required} />
-      <ul className="flex flex-wrap max-w-[260px] list-none">
+      <ul className="flex flex-wrap max-w-[260px] list-none pl-0">
```
rchlfryn, line 30: do we want to keep the max-w here since we changed it to flex-col? Maybe just set it to 50%

src/blocks/GenericEmbed/Component.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  useEffect(() => {
+    if (typeof window === 'undefined' || !html) return
+
+    // Normalize problematic quotes that are parsed incorrectly by DOMParser and DOMPurify
+    const normalizedHTML = html.replaceAll('“', '"').replaceAll('”', '"')
+
+    const sanitized = DOMPurify.sanitize(normalizedHTML, {
```
rchlfryn, line 35: Does isomorphic-dompurify offer any erroring for bad scripts or ill formatted code? It might be nice to display an error if something isnt right but if it's not easy to tell, probably not worth the time

src/components/AlignContentPicker/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <div className="flex items-start mb-6">
+      <FieldLabel htmlFor={path} label={field.label} required={field.required} />
+      <ToggleGroup
+        type="single"
+        value={currentValue}
+        onValueChange={(newValue) => newValue && setValue(newValue)}
+        variant="outline"
+        className="justify-start"
```
rchlfryn, line 29: maybe use `size='lg'` here instead of sizing the toggle below in line 39.

https://ui.shadcn.com/docs/components/toggle-group#large

src/components/AlignContentPicker/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+      >
+        {alignOptions.map((option) => {
+          const Icon = option.icon
+          return (
+            <ToggleGroupItem
+              key={option.value}
+              value={option.value}
+              aria-label={option.label}
+              className={cn('w-10 h-10 p-2 cursor-pointer')}
```
rchlfryn: ⛏️  do we need `cn` here?
```suggestion
              className='w-10 h-10 p-2 cursor-pointer'
```

src/components/AlignContentPicker/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    { value: 'left', label: 'Left', icon: AlignLeft },
+    { value: 'center', label: 'Center', icon: AlignCenter },
+    { value: 'right', label: 'Right', icon: AlignRight },
+  ]
+
+  const { path, field } = props
+
+  const { value, setValue } = useField({ path })
+
+  const currentValue = (value as string) || 'left'
```
rchlfryn: Do we need this type casting here? what if we set the type for value above as string?

rchlfryn (dismissed):
LG  - a couple of comments

src/components/AlignContentPicker/index.tsx
busbyk: Nope! Thanks.

src/components/AlignContentPicker/index.tsx
busbyk: 👍

src/components/AlignContentPicker/index.tsx
busbyk: Actually didn't need the sizing at all. The default size works well and I didn't need to size the Toggle or Icon after all.

src/components/ColorPicker/index.tsx
busbyk: I like the pixels max width because, with the number of colors we have right now, it ensures the two rows of colors are the same length. I think it looks nice.

src/blocks/GenericEmbed/Component.tsx
busbyk: The behavior of DOMPurify is just to remove the bad scripts from the html. So it would be a little tricky to reliably detect _why_ the input and output are different. 

DOMParser's parseFromString will return an error node if the inputted HTML is poorly formatted. It won't throw. I can add a check for that in the field config.

busbyk: Ha I'm just doing some further testing and I can't get DOMParser to return the error node because DOMPurify's sanitize does such a good job of fixing poorly formatted HTML 😂. For example if I pass in `<div>poorly formatted</` it outputs `<div>poorly formatted</div>"</"` or I pass in `<div>poorly formatted</p></div>` it outputs `<div>poorly formatted<p></p></div>`. 

So I think we're fine leaving this as is. Users can use the live preview to see how it will render.

stevekuznetsov (approved):

PR #418, opened 2025-08-15
Bugfix: Live previews

busbyk:
Live previews broke due to the canonical url logic for pages that are nested in the navigation. This fixes this by adding handling for canonical urls.

### DIFF
--- a/src/app/(frontend)/[center]/next/preview/route.ts
+++ b/src/app/(frontend)/[center]/next/preview/route.ts
@@ -72,7 +72,9 @@ export async function GET(
           select: {},
           where: {
             slug: {
-              equals: slug,
+              // Exception for pages collection which has a concept of canonical urls based on their nesting in the navigation
+              // Uses the last path part as the slug
+              equals: collection === 'pages' ? slug.split('/').filter(Boolean).pop() : slug,
             },
           },
         })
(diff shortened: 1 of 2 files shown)


src/app/(frontend)/[center]/next/preview/route.ts
```diff
@@ -72,7 +72,9 @@ export async function GET(
           select: {},
           where: {
             slug: {
-              equals: slug,
+              // Exception for pages collection which has a concept of canonical urls based on their nesting in the navigation
+              // Uses the last path part as the slug
+              equals: collection === 'pages' ? slug.split('/').filter(Boolean).pop() : slug,
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/next/preview/route.ts:77]: Will we want to add the same thing for posts?

rchlfryn (dismissed):

src/app/(frontend)/[center]/next/preview/route.ts
Comment (busbyk) [reply src/app/(frontend)/[center]/next/preview/route.ts:77]: Good thought but we don't use the same canonical url logic for posts since they are never nested under additional paths if they're included in the navigation like pages are. This is because pages have a catch all segment `src/app/(frontend)/[center]/[...segments].page.tsx` to handle that nesting. But posts should always be nested in the context of the blog so even if they're nested in the navigation the url will always be relative to `/blog`. 

So I can't think of a situation where we would need to handle parsing a slug with multiple path parts for any collections besides pages here for now.

stevekuznetsov (approved):

PR #419, opened 2025-08-15
Fixes blocks spacing issues

### LINKED ISSUE #404: Excessive spacing on production pages
**Expected behavior**

I would expect the spacing between the header and each block to be smaller. 

Some pages Rusty put together:
https://sac.avy-fx.org/about/about-us
https://sac.avy-fx.org/support/other-ways-to-give
https://sac.avy-fx.org/support/corporate-sponsorship

**To Reproduce**
Create a page. Add multiple blocks. 

**Environment details**
- Production
- Desktop
- Chrome

**Existential musings**
[My most requested recipe is this](https://smittenkitchen.com/2017/05/broccoli-rubble-farro-salad/)

### LINKED ISSUE #406: Footer logo rendering weird for SNFAC
**Describe the bug**
The footer logo feels off compared to SAC and NWAC

**Expected behavior**
I would expect the logo to be centered like it is in the header. 

**Screenshots**
Image

**Environment details**
- Desktop 
- Chrome

**Existential musings**
[My favorite cookbook](https://www.sami-tamimi.com/cookbooks/falastin)

### LINKED ISSUE #407: Margins are rending different for different blocks
**Describe the bug**
https://sac.avy-fx.org/support/corporate-sponsorship

**To Reproduce**
From the page editor, I added a full width text block after the image with text blocks and got the following:

**Expected behavior**
I would expect each block to render with the same margins 

**Screenshots**
Image

**Environment details**
- Desktop
- Chrome

**Existential musings**
For Indian cooking, I like anything by [Meera Sodha](https://meerasodha.com/books/)

### LINKED ISSUE #414: Callout block spacing seems inverse to expected
Based on the preview image in the page builder, I would expect the callout block on the right to be narrower and the text on the left to be wider:
https://sac.avy-fx.org/support/donate

Image

**To Reproduce**

1. Open a page
2. Add a callout block
3. Publish

**Environment details**

- Desktop
- Chrome

rchlfryn:
### Description
Fixes general spacing issues with blocks. See commits...

### Fixes
Fixes #414, Fixes #407, Fixes #404, Fixes #406

### DIFF
--- a/src/blocks/ContentWithCallout/Component.tsx
+++ b/src/blocks/ContentWithCallout/Component.tsx
@@ -4,7 +4,7 @@ import RichText from '@/components/RichText'
 
 export const ContentWithCalloutBlock = ({ callout, richText }: ContentWithCalloutBlockProps) => {
   return (
-    <div className="container">
+    <div className="container py-4">
       <div className="flex flex-col md:flex-row items-center gap-4">
         {richText && (
           <div className="md:basis-2/3 md:flex-shrink-0 md:flex-grow-0 w-full">
--- a/src/blocks/ContentWithCallout/Component.tsx
+++ b/src/blocks/ContentWithCallout/Component.tsx
@@ -6,9 +6,13 @@ export const ContentWithCalloutBlock = ({ callout, richText }: ContentWithCallou
   return (
     <div className="container">
       <div className="flex flex-col md:flex-row items-center gap-4">
-        {richText && <RichText className="mb-0" data={richText} enableGutter={false} />}
+        {richText && (
+          <div className="md:basis-2/3 md:flex-shrink-0 md:flex-grow-0 w-full">
+            <RichText className="mb-0" data={richText} enableGutter={false} />
+          </div>
+        )}
         {callout && (
-          <div className="p-6 bg-brand-200 rounded-sm">
+          <div className="p-6 bg-brand-200 rounded-sm md:basis-1/3 md:flex-shrink-0 md:flex-grow-0 w-full">
             <RichText className="mb-0" data={callout} enableGutter={false} />
           </div>
         )}
--- a/src/blocks/ContentWithCallout/Component.tsx
+++ b/src/blocks/ContentWithCallout/Component.tsx
@@ -4,7 +4,7 @@ import RichText from '@/components/RichText'
 
 export const ContentWithCalloutBlock = ({ callout, richText }: ContentWithCalloutBlockProps) => {
   return (
-    <div className="container py-4">
+    <div className="container py-16">
       <div className="flex flex-col md:flex-row items-center gap-4">
         {richText && (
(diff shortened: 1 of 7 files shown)


src/blocks/ImageTextList/Component.tsx
```diff
@@ -25,7 +25,7 @@ export const ImageTextList = (props: Props) => {
   const isSideLayout = layout === 'side'
 
   return (
-    <div className="container my-16">
```
rchlfryn, line 28: I know this is a "big change" but all our spacing is done in the blocks now.

src/blocks/ImageLinkGrid/Component.tsx
```diff
@@ -22,7 +22,7 @@ export const ImageLinkGrid = (props: Props) => {
   const colsSpanClass = colsClasses[numOfCols]
 
   return (
-    <div className="container">
+    <div className="container py-4">
```
busbyk: Why py-4 here and not py-16 like the other blocks?

src/blocks/ContentWithCallout/Component.tsx
```diff
@@ -4,7 +4,7 @@ import RichText from '@/components/RichText'
 
 export const ContentWithCalloutBlock = ({ callout, richText }: ContentWithCalloutBlockProps) => {
   return (
-    <div className="container">
+    <div className="container py-4">
```
busbyk: Why py-4 here and not py-16 like the other blocks?

src/blocks/ImageTextList/Component.tsx
busbyk: You mean that this block always contains nested blocks which each already have `py-16` on their container divs so we don't need this verical margin anymore?

busbyk: Oh hmm I don't think the columns have the vertical padding you were expecting? (the second red rectangle should be the same height as the top one right?)

CleanShot 2025-08-15 at 15 45 21@2x

The columns have `my-6`: 
CleanShot 2025-08-15 at 15 46 55@2x

busbyk (commented):
Looks great, except a few comments.

rchlfryn:
> You mean that this block always contains nested blocks which each already have py-16 on their container divs so we don't need this verical margin anymore?

Exactly what I mean. And I honestly don't like how much padding we have between sections, which is why I left some smaller. But for consistency sake and for them to not have it look like a random accident, I will update all of them to be `py-16`

src/blocks/ContentWithCallout/Component.tsx
rchlfryn: I don't like how much spacing we have but will change it for consistency and we can talk about it as a group whether to change it or not

busbyk (approved):
@rchlfryn I'm down to adjust our default padding down (I don't feel strongly about the current padding, fyi). Let's merge this in and then get some feedback from the rest of the team and Sierra folks? Then we can adjust the default padding based on that.

PR #422, opened 2025-08-17
Home page global collection

### LINKED ISSUE #248: AC Staff can publish news rich text box on home page
Staff can add important news directly below the forecast

Examples:
- Forecasting starts on 12/1
- Spring forecasting schedule
- Forecasting has ended 

Image

busbyk:
Resolves #248 
Contributes to #6 

## Key changes

This introduces a global collection, `homePages`, to control CMS content to be displayed on a center's home page along with the map widget. 
- Quick links (buttons that render next to the map on desktop and below it on mobile)
- Highlighted content (the intent is for this to be used for forecaster news but could be for anything)
- Layout (an array of blocks that render just like the pages collection pages)

## Things to consider

- There was some discussion about restricting access to controlling the news on the home page to forecasters. Since I took the approach of a very generic "highlighted content" section, does this still make sense? We could certainly restrict updating that field to a specific tenant role if we wanted to. 
- The height of the map is still controlled in the AFP. I think this is fine for now but eventually we may want to take control of that by passing a prop to the map widget. That would require a PR to the `afp-public-widgets` repo. 
- We can play around with the layout of quick links and map widths as well. Happy to do a live coding session to try these out. 

## Small changes included

- Seed script was using logo images for icons instead of the icon images. Fixes a corrupted icon image.
- Removes an unused argument in revalidateWidgetPages that was giving an eslint warning

## Screenshots
CleanShot 2025-08-16 at 20 22 48@2x
CleanShot 2025-08-16 at 20 23 32@2x
CleanShot 2025-08-16 at 20 23 20@2x


src/collections/HomePages/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+    afterChange: [revalidateHomePage],
+    beforeChange: [populatePublishedAt],
+    afterDelete: [revalidateHomePageDelete],
+  },
+  versions: {
+    drafts: {
+      autosave: true,
+    },
+    maxPerDoc: 10,
+  },
```
stevekuznetsov, line 192: Schema looks good!

PR #426, opened 2025-08-18
Explicit cookie domain setting & token cookie rename

### LINKED ISSUE #425: Mismatch between root domain env var and cookie domain causes permissions errors
**Describe the bug**
I cant save/publish any changes in the admin panel at the moment

**Screenshots**
I get this in a lot of places: 

Image

When trying to edit a page, I can't get a new block to load

**Environment details**

- Desktop/mobile/tablet
- Chrome

busbyk:
Resolves #425 

See decision doc.

### DIFF
--- a/docs/decisions/009-cookie-domain.md
+++ b/docs/decisions/009-cookie-domain.md
@@ -1,0 +1,27 @@
+# Explicitly Setting the Cookie Domain
+
+Date: 2025-08-18
+
+Status: accepted
+
+## Context
+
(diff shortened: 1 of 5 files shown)


busbyk:
Playing devil's advocate against myself here: setting our `NEXT_PUBLIC_ROOT_DOMAIN` to `avy-fx.org` (i.e. the apex domain) and adding a 307 redirect in our DNS for `www.avy-fx.org` -> `avy-fx.org` solves this problem without any code changes necessary. I wonder if we even need this changes / if we want the consequence of cookies being shared across subdomains (which necessitates the cookie prefix so they aren't shared across environments since we use the same root domain).

busbyk:
I'm just leaving this open in case anyone is excited about this change for some reason but I think I've effectively talked myself out of it.

busbyk:
We can re-open if necessary

PR #427, opened 2025-08-19
Minimize seed data

rchlfryn:
Minimizing seed data to only seed 4 bios per AC.

Feel free to yes and adding a flag or command to conditionally pick up a bio file. I have not had time


busbyk (approved):
I love it!

PR #428, opened 2025-08-19
Blog List Block

### LINKED ISSUE #244: Blocks: Blog listing
Block that can list blog posts:

Image

Should accept optional:
- heading/title
- tags to filter results by

Probably sort by ascending date published

busbyk:
Resolves #244 

## Backstory

This was a little different than our existing blocks because the data displayed is based on a query. Figuring out the best way to do this was a bit of a journey because of live preview, the complexity of our revalidation logic, and the timing of hooks. 

Turns out we can't use the Payload Local API in components that we embed in the our RichText component (the jsx converts don't like that). So I first implemented the use of an `afterRead` hook to execute the query and store the data in a hidden field on the block. This worked really well and played nicely with everything except for revalidation of blocks in the posts collection. 

Determining what blocks are in a `richText` field is done by inspecting the `blocksInContent` field which is kept up to date via a `beforeChange` hook. This hook runs before the `queriedPosts` field's afterRead hook (same deal for afterChange hooks) and so wouldn't recognize that a given post was actually in that document's `richText` field. 

I tried a few other approaches including `beforeChange` hooks on the fields that determine the query (`filterByTags`, `sortBy`, `maxPosts`) but you can only update the field's value in a `FieldHook` not any other fields' values. 

So I finally went with a custom field component for the `queriedPosts` field that uses the rest API to update the value of the `queriedPosts` field in real time. This has the side benefit of showing the user which posts will be displayed without needing to use the live preview. It renders a readOnly `SelectInput`. 

Ah the complexities of Payload and Next.js revalidation. 

## Features

- Allows dynamic posts via fields: tags, maxPosts, sort (same options as blog listing page)
- View all button uses tags, sort in the query params so the same options will be applied to the blog listing page upon visiting that link
- Allows setting staticPosts to manually pick specific blog posts to display in the list
- Can optionally wrap in a container in the RichText BlocksFeature
- Has a max height allowing 4 posts + a bit of the fifth with scrolling enabled
- A fallback graphic when a post does not have a featuredImage

## Revalidation

This identified some limitations with the revalidation helpers to find blocks and relationships embedded in routable collections. I considered opening a separate PR but this can't be deployed without updating the revalidation to handle deeply nested hasMany relationship fields because pages wouldn't be revalidated correctly in production. 

## Middleware and resolveTenant fns 

I originally wrote this to use the Payload Local API in the BlogListBlockComponent directly before realizing that the jsx converter wouldn't allow that. In that design I needed a way to determine the tenant to filter results by it. So I updated the middleware to always add the `payload-tenant` cookie  when not on the root domain - not just on `/admin`. 

Now that I'm querying the Local API in the afterRead hook the tenant _should_ always be available on the incoming data arg because all collections that we embed this in, as of right now, have a tenant field on them. 

So:
```
  const tenant =
    data && 'tenant' in data && typeof data?.tenant === 'number'
      ? await resolveTenantFromCookie(req.headers)
      : null
```

Became this:
```
  const tenant =
    data && 'tenant' in data && typeof data?.tenant === 'number'
      ? await resolveTenant(data.tenant)
      : null
```

But I liked the option of determining the tenant from the cookie in `(frontend)` server components so I left the new `resolveTenantFromCookie` function and middleware changes in. I also simplified the existing resolveTenant function. If it's super annoying to have those changes in here I can strip them out. 

## Screenshots

In RenderBlocks:
CleanShot 2025-08-19 at 11 08 54@2x

Mobile:
CleanShot 2025-08-19 at 11 09 20

More than 4 posts + post with a missing featuredImage using the fallback graphic:
CleanShot 2025-08-19 at 11 22 34

In RichText BlocksFeature with wrapInContainer=false (default):
CleanShot 2025-08-19 at 11 07 36

In a RichText BlocksFeature with wrapInContainer=true:
CleanShot 2025-08-19 at 11 06 09


public/thumbnail/BlogListThumbnail.jpg
rchlfryn, line 1: ⛏️  Can we add the title and buttons here since it is always shown? I added to the one you made in figma. Lmk what you think

src/blocks/BlogList/Component.tsx
```diff
@@ -0,0 +1,95 @@
...
+  if (!posts) {
+    return null
+  }
+
+  const bgColorClass = `bg-${backgroundColor}`
+
+  return (
+    <div className={cn(wrapInContainer && bgColorClass)}>
+      <div className={cn(wrapInContainer && 'container py-16', className)}>
+        <div className="bg-card text-card-foreground p-6 border shadow rounded flex flex-col gap-6">
```
rchlfryn: I think we want `rounded-lg` here

src/blocks/BlogList/config.ts
```diff
@@ -0,0 +1,137 @@
...
+    }),
+    label: 'Below Heading Content',
+    admin: {
+      description: 'Optional content to display below the heading and above the blog list.',
+    },
+  },
+  colorPickerField('Background color'),
+]
+const defaultPostRelatedFields: Field[] = [
+  {
```
rchlfryn, line 40: ⛏️  I can take or leave this... maybe we move the filter to after max post integer? For UX, I try to keep all required things together and towards the top.

src/blocks/BlogList/config.ts
```diff
@@ -0,0 +1,137 @@
...
+      { label: 'Newest to Oldest', value: '-publishedAt' },
+      { label: 'Oldest to Newest', value: 'publishedAt' },
+    ],
+    required: true,
+    admin: {
+      description: 'Select how the list of posts will be sorted.',
+    },
+  },
+  {
+    name: 'maxPosts',
```
rchlfryn, line 65: Shouldn't this be required? I am guessing it is not if we want an infinite amount of posts to show, but we have a fallback limit of 4
```ts          
limit: String(maxPosts || 4),
```

src/blocks/BlogList/Component.tsx
```diff
@@ -0,0 +1,95 @@
...
+                }
+
+                return null
+              })
+            ) : (
+              <h3>There are no posts matching these results.</h3>
+            )}
+          </div>
+          {!hasStaticPosts && (
+            <Button asChild className="w-full not-prose">
```
rchlfryn: Can we remove full width for desktop? `md:self-start` should do the trick. Also will need to remove `w-full not-prose` since they are not doing anything

src/blocks/BlogList/config.ts
```diff
@@ -0,0 +1,137 @@
...
+    hasMany: true,
+    admin: {
+      readOnly: true,
+      components: {
+        Field: '@/blocks/BlogList/fields/QueriedPostsComponent#QueriedPostsComponent',
+      },
+    },
+  },
+  {
+    name: 'staticPosts',
```
rchlfryn, line 92: This should be required because if a user does not select posts here, it does not display the queried posts. Is that the intentional behavior?

src/blocks/BlogList/Component.tsx
```diff
@@ -0,0 +1,95 @@
...
+                <h2>{heading}</h2>
+              </div>
+            )}
+            {belowHeadingContent && (
+              <div>
+                <RichText data={belowHeadingContent} enableGutter={false} />
+              </div>
+            )}
+          </div>
+          <div className="flex flex-col gap-4 not-prose max-h-[400px] overflow-y-auto">
```
rchlfryn: Remove `not-prose` since it is not doing anything

src/blocks/BlogList/Component.tsx
```diff
@@ -0,0 +1,95 @@
...
+            {posts && posts?.length > 0 ? (
+              posts?.map((post, index) => {
+                if (typeof post === 'object' && post !== null) {
+                  return <PostPreviewSmallRow doc={post} key={`${post.id}__${index}`} />
+                }
+
+                return null
```
rchlfryn: Shouldn't we return `There are no posts matching these results.` here as well? maybe we could combine this with the previous posts check or remove it completely

src/blocks/BlogList/config.ts
```diff
@@ -0,0 +1,137 @@
...
+          ...rootFeatures,
+          BlocksFeature({
+            blocks: [ButtonsBlock, MediaBlock, GenericEmbedLexical],
+          }),
+          HorizontalRuleFeature(),
+          InlineToolbarFeature(),
+        ]
+      },
+    }),
+    label: 'Below Heading Content',
```
rchlfryn: ⛏️ 
```suggestion
    label: 'Content Below Heading',
```

src/blocks/BlogList/fields/QueriedPostsComponent.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const fetchPosts = async () => {
+      setIsLoading(true)
+      setDisabled(true)
+
+      try {
+        const tenantId = typeof tenant === 'number' ? tenant : (tenant as { id?: number })?.id
+        if (!tenantId) return
+
+        const params = new URLSearchParams({
+          limit: String(maxPosts || 4),
```
rchlfryn, line 51: If this is left blank, I would assume the fallback would default to the max which is 10.

src/blocks/BlogList/config.ts
```diff
@@ -0,0 +1,137 @@
...
+export const BlogListBlockLexical = blogListBlockWithFields([
+  ...defaultStylingFields,
+  {
+    name: 'wrapInContainer',
+    admin: {
+      description:
+        'Checking this will render the block with additional padding around it and using the background color you have selected.',
+    },
+    type: 'checkbox',
+    defaultValue: false,
```
rchlfryn, line 134: Is there a reason the `defaultValue` here is false when it is true every where else in this PR (besides the rich text editor inline)?

src/blocks/BlogList/config.ts
rchlfryn: May have to alter migrations if we change this?

src/blocks/BlogList/Component.tsx
rchlfryn: Do we want to make columns for posts on desktop?

Image

rchlfryn (changes requested):
LG - a few nit picks and more style changes 

Question: Do we want to add this to the homepage as an option for content?

src/blocks/BlogList/Component.tsx
busbyk: Would you like to go with `rounded-lg` on all of our card-style components?

src/blocks/BlogList/config.ts
busbyk: We could make it required. The defaultValue here matches the fallback amount of 4 intentionally. I suppose a note in the description of the default would be good if we kept it optional. 

I often like to make things optional and handle missing values nicely to have as little friction to saving something as possible but I think that opinion is in the minority on the team.

src/blocks/BlogList/config.ts
busbyk: I figured that the tags filter is the most important field so I put it at the top of the post related fields. sortBy is required but it also can't be cleared so a user can't be held up by that being empty - it can't be empty.

src/blocks/BlogList/config.ts
busbyk: Hm if that's the case then perhaps you've found a bug. 

The behavior should be:
- If there are staticPosts selected, display only the staticPosts and do not display a view all button
- If there are no staticPosts selected, display the queriedPosts and display a view all button with query params matching filterByTags, sortBy

src/blocks/BlogList/Component.tsx
busbyk: `not-prose` is here intentionally. 

This component can be rendered as a block in the RichText component which has `prose` on it. Applying `not-prose` conditionally in the component's className is the logical solution but since belowHeadingContent is also a richText field, this would mess with the `prose` class styling there because a parent would have `not-prose` on it. 

So we need to apply `not-prose` more granularly to avoid `prose` styles being applied to this component.

busbyk: > Do we want to make columns for posts on desktop?
> 
> Image

Sure. I was following the designs:
CleanShot 2025-08-21 at 13 34 16@2x

But I think that's a nice design.

public/thumbnail/BlogListThumbnail.jpg
busbyk: Nice, looks good!

src/blocks/BlogList/Component.tsx
busbyk: Well this is for an individual post but we actually don't even need this check because I filter both queriedPosts and staticPosts to ensure they are objects and not null above already. So I'll just remove this check.

src/blocks/BlogList/config.ts
busbyk: Yea. The pattern I've used for this block, the GenericEmbed, and the SingleBlogPost has wrapInContainer always set to true for rendering the component in a blocks field / RenderBlocks component and defaults to false initially for rendering the component in a richText field / RichText component because that is typically how we render blocks in the RichText component but a user might want to set a background color with padding by wrapping in a container. 

https://github.com/NWACus/web/blob/blocks/blog-posts/src/components/RichText/index.tsx#L73

This field is only used in the lexical variant of this block field. So it makes sense for this to default to false.

src/blocks/BlogList/fields/QueriedPostsComponent.tsx
busbyk: I chose to fallback to the defaultValue of the field.

src/blocks/BlogList/config.ts
busbyk: Not for the label but I've had issues with needing to recreating migrations after merging from main where new migrations were added. So I've been just deleting the previous migrations and creating a new one which works well. So I just did that since there were two migrations for this PR anyways - now they're combined.

src/blocks/BlogList/fields/QueriedPostsComponent.tsx
busbyk: Let's decide on your other comment thread.

src/blocks/BlogList/config.ts
rchlfryn: I forgot to remove this comment. It did not load one time and then I could not recreate it. We will mark it as resolved for now.

src/blocks/BlogList/config.ts
rchlfryn: Let's leave this.

src/blocks/BlogList/Component.tsx
busbyk: I like it. Added.

src/blocks/BlogList/Component.tsx
busbyk: My laziness has been corrected 😅 : 

CleanShot 2025-08-21 at 16 13 24
CleanShot 2025-08-21 at 16 11 36

busbyk: Feel free to change the container query breakpoint

rchlfryn (approved):

PR #446, opened 2025-08-22
Add sponsors collection and block

### LINKED ISSUE #33: AC staff is able to display sponsor logos/ads
Staff is able to add sponsor logos to the website in exchange for donation. Placement weighted differently. https://docs.google.com/document/d/1UakR6R1zEpTsdBN386k1jawJ3Uh6jqRULAdu3TqHWFE/edit?tab=t.0

### LINKED ISSUE #390: Cycle Sponsors Placed in Banner
"Some manner of cycling is probably necessary this year" - David. Eventually would like to implement #389

### LINKED ISSUE #391: Grid block of sponsors
Have a block of sponsors to do things like: 

https://nwac.us/support-corporate-sponsorship/
https://www.sawtoothavalanche.com/agency-partners/

rchlfryn:
### Description
Add sponsors collection and block to give admins the ability to add sponsors to pages.

Screenshot 2025-08-22 at 09 51 18

### Issues
Resolves #33 #390 #391


src/endpoints/seed/index.ts
```diff
@@ -558,12 +562,14 @@ export const seed = async ({
...
-    getSeedImageByFilename('image-post2.webp', payload.logger),
-    getSeedImageByFilename('image-post3.webp', payload.logger),
-    getSeedImageByFilename('image-post3.webp', payload.logger),
-  ])
+  const [image1Buffer, image2Buffer, image3Buffer, imageMountainBuffer, imageAcmeBuffer] =
+    await Promise.all([
+      getSeedImageByFilename('image-post1.webp', payload.logger),
+      getSeedImageByFilename('image-post2.webp', payload.logger),
+      getSeedImageByFilename('image-post3.webp', payload.logger),
+      getSeedImageByFilename('image-mountain.webp', payload.logger),
```
busbyk, line 573: Nice! I like the graphic.

src/blocks/SponsorsBlock/config.tsx
busbyk, line 1: Do we want to:
1. Allow setting the background color of the entire section in addition to the title container?
2. Allow wrapping in a container optionally for use in richText BlocksFeatures?

src/collections/Sponsors/index.ts
```diff
@@ -0,0 +1,78 @@
+import { accessByTenantRole } from '@/access/byTenantRole'
+import { filterByTenant } from '@/access/filterByTenant'
+import { contentHashField } from '@/fields/contentHashField'
+import { tenantField } from '@/fields/tenantField'
+import { Sponsor } from '@/payload-types'
+import isAbsoluteUrl from '@/utilities/isAbsoluteUrl'
+import { CollectionConfig, DateFieldValidation, TextFieldSingleValidation } from 'payload'
+
+const validateExternalUrl: TextFieldSingleValidation = (val) =>
```
busbyk: We have an [existing function](https://github.com/NWACus/web/blob/sponsors/src/fields/navLink/index.ts#L5). Do we want to break this out into it's own utility function file?

src/blocks/SponsorsBlock/config.tsx
```diff
@@ -0,0 +1,24 @@
+import colorPickerField from '@/fields/color'
+import { getTenantFilter } from '@/utilities/collectionFilters'
+import type { Block } from 'payload'
+
+export const SponsorsBlock: Block = {
+  slug: 'sponsors',
```
busbyk: Heh, why is this showing up as singular "Sponsor" instead of "Sponsors"? 
CleanShot 2025-08-25 at 15 51 38@2x

src/blocks/SponsorsBlock/Component.tsx
```diff
@@ -0,0 +1,46 @@
...
+      )}
+      <div className="flex flex-wrap justify-evenly items-center">
+        {typeof sponsors === 'object' &&
+          sponsors.map((sponsor, index: number) => {
+            if (typeof sponsor === 'number') return
+            const now = truncateTime()
+            const showLogoStart = sponsor.startDate ? truncateTime(sponsor.startDate) <= now : true
+            const showLogoEnd = sponsor.endDate ? truncateTime(sponsor.endDate) >= now : true
+
+            if (showLogoStart && showLogoEnd)
```
busbyk: If there are no valid sponsors (i.e. they're populated, they aren't outside of startDate and endDate) should we also hide the title? Or should we render some kind of default empty state? 

We could also allow the content editor to provide a custom empty state using a richText field on the sponsors block or in the website settings.

busbyk: You could also simplify the typechecks here using a filter above:

```
  const validSponsors = sponsors
    .filter((sponsor): sponsor is Sponsor => typeof sponsor === 'object' && sponsor !== null)
```

src/blocks/SponsorsBlock/Component.tsx
```diff
@@ -0,0 +1,46 @@
...
+              return (
+                <div className="w-1/2 md:w-1/3 lg:w-1/5 p-10" key={index}>
```
busbyk: I don't think it's strictly necessary here but I've run into bugs enough times when just using the index of a .map function that I always use compound keys now like `${sponsor.id}_${index}`. See the pitfall callout in this section of the docs: https://react.dev/learn/rendering-lists#why-does-react-need-keys 

Again, I don't think we'll run into it in this particular scenario because these elements don't get reordered on the fly - even in live preview a full page refresh is happening behind the scenes. But it's just a habit of mine now and figured I'd share.

src/blocks/SponsorsBlock/Component.tsx
```diff
@@ -0,0 +1,46 @@
...
+          <h2 className={textColor}>{title}</h2>
+        </div>
+      )}
+      <div className="flex flex-wrap justify-evenly items-center">
+        {typeof sponsors === 'object' &&
+          sponsors.map((sponsor, index: number) => {
+            if (typeof sponsor === 'number') return
+            const now = truncateTime()
+            const showLogoStart = sponsor.startDate ? truncateTime(sponsor.startDate) <= now : true
+            const showLogoEnd = sponsor.endDate ? truncateTime(sponsor.endDate) >= now : true
```
busbyk: Should we just parse the ISO strings from Payload and use startOf/endOf utilities from `date-fns`? Could just determine `now` once outside of render `const now = new Date()`. I _think_ this would be less prone to timezone mismatches when the users current time is close enough to the next day.

```
  const showLogoStart = sponsor.startDate
    ? startOfDay(new Date(sponsor.startDate)) <= now
    : true
  const showLogoEnd = sponsor.endDate ? endOfDay(new Date(sponsor.endDate)) >= now : true
```

src/collections/Sponsors/index.ts
```diff
@@ -0,0 +1,78 @@
...
+      admin: {
+        position: 'sidebar',
+      },
+    },
+    contentHashField(),
+  ],
+  hooks: {
+    afterChange: [],
+    afterDelete: [],
+  },
```
busbyk, line 74: Guidance on revalidation for new reference collections: https://github.com/NWACus/web/blob/sponsors/docs/revalidation.md#L136

Example that you can basically copy and paste and change `teams` to `sponsors`: https://github.com/NWACus/web/blob/sponsors/src/collections/Teams/hooks/revalidateTeam.ts

busbyk (commented):
- I was thinking that the size of the sponsor image would be determined by the number of sponsors selected. So if a content editor wanted to have a single sponsor displayed larger than others. Something like:
	- 1 sponsor selected: display full width (or maybe more like 75% width)
	- 2 sponsors selected: each displays  ~50% 
	- 3 sponsors selected: each displays ~33%
	- 4 + we could wrap but each would still display 33% width
	This might not have the intended effect on mobile, though. 
- Do we want to allow sponsor blocks in other richText fields besides the content block and content with callout block?

src/blocks/SponsorsBlock/config.tsx
rchlfryn: Oh weird - I dont know. Maybe a payload thing?

src/blocks/SponsorsBlock/Component.tsx
rchlfryn: Good note - I will update this just in case.

src/blocks/SponsorsBlock/config.tsx
rchlfryn: For now, I would say:
1. No because then we need to make sure all the images have consistent or no backgrounds
2. Yeah we can add this

src/blocks/SponsorsBlock/Component.tsx
rchlfryn: I like using the filter. I will just add the date checks to the filter as well so we can simplify the null check you mentioned as well.

rchlfryn:
> I was thinking that the size of the sponsor image would be determined by the number of sponsors selected. So if a content editor wanted to have a single sponsor displayed larger than others. Something like:


I started with this but switched to keep it simple. Updated in [a9ce0ab](https://github.com/NWACus/web/pull/446/commits/a9ce0abea5429a3f60a5fc11239ab5510f078f2c)
Corporate-Sponsorship-Northwest-Avalanche-Center-08-28-2025_10_07_AM


> Do we want to allow sponsor blocks in other richText fields besides the content block and content with callout block?

I am going to add it in a few more places and we can add more later if needed

PR #447, opened 2025-08-23
Quick wins: block styling and admin view

### LINKED ISSUE #380: Content block is half-width on tablet
Image

### LINKED ISSUE #384: Create new blocks in open state
Right now when you add a new block to the pages layout, it is rendered in it's collapsed state. It makes more sense to me to render it in it's open/uncollapsed state. I assume this is just some default param from Payload that we can change. 

Image

Should check other collections to see if those need to be updated too.

### LINKED ISSUE #434: Media block is always left aligned
**Describe the bug**
Media from the media block is always aligned with the left margin

**To Reproduce**

1. Go to any page
2. Add Media Block
3. Publish

**Expected behavior**
I would expect to be able to choose whether it's left, center or right aligned on the page

**Screenshots**
https://sac.avy-fx.org/about/about-us

Image

**Environment details**

- Desktop
- Chrome

**Existential musings**
If you get on the plane last you can sit wherever you want... whatever seats that are open aren't going to be filled.

rchlfryn:
### Description
* Centers media block images
* Fixes admin view for initially adding a block to be open instead of collapased
* Fixes content width on tablet

### Issue
Fixes #434 
Fixes #384
Fixes #380

### DIFF
--- a/src/blocks/MediaBlock/Component.tsx
+++ b/src/blocks/MediaBlock/Component.tsx
@@ -36,13 +36,13 @@ export const MediaBlock = (props: Props) => {
       {(media || staticImage) && (
         <Media
           className="my-0"
-          imgClassName={cn(imgClassName)}
+          imgClassName={cn('mx-auto', imgClassName)}
           resource={media}
           src={staticImage}
         />
       )}
       {caption && (
-        <div className={cn({ container: !disableInnerContainer }, captionClassName)}>
+        <div className={cn('mt-4', { container: !disableInnerContainer }, captionClassName)}>
           <RichText data={caption} enableGutter={false} className="text-xs" />
         </div>
       )}
--- a/src/collections/Pages/index.ts
+++ b/src/collections/Pages/index.ts
@@ -125,7 +125,7 @@ export const Pages: CollectionConfig<'pages'> = {
               ],
               required: true,
               admin: {
-                initCollapsed: true,
+                initCollapsed: false,
               },
             },
           ],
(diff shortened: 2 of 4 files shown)


src/blocks/MediaBlock/Component.tsx
```diff
@@ -36,13 +36,13 @@ export const MediaBlock = (props: Props) => {
       {(media || staticImage) && (
         <Media
           className="my-0"
-          imgClassName={cn(imgClassName)}
+          imgClassName={cn('mx-auto', imgClassName)}
```
busbyk, line 39: So now images are always centered. It would be sweet to use the [alignContent field](https://github.com/NWACus/web/blob/main/src/fields/alignContent.ts) for this component like it's used in the GenericEmbed block.

src/collections/Pages/index.ts
```diff
@@ -125,7 +125,7 @@ export const Pages: CollectionConfig<'pages'> = {
               ],
               required: true,
               admin: {
-                initCollapsed: true,
+                initCollapsed: false,
```
busbyk, line 128: It would be sweet to have blocks initialize open when adding a new one but closed when visiting a a page. I just had the experience of opening up the blocks page and now all of the blocks are expanded by default. Since there are so many my first step was to click "collapse all" to scroll down and add a new one to the bottom. But I don't think we have separate behaviors there. 🤷

busbyk (commented):
@rchlfryn

PR #448, opened 2025-08-24
Add form option to footer

### LINKED ISSUE #200: Add subscribe form to footer

rchlfryn:
### Description
Adds a form option with optional title and subtitle to the footer with options embed or use a native Payload form 

**Options**

None
footer-none

Embedded
footer-embed

Native
footer-form

### Issue
Fixes #200


busbyk (commented):
I think the optional wrapInContainer on the form block messed with how this is rendered in a full page block:
CleanShot 2025-08-25 at 16 54 18@2x

A few other things I noticed: https://www.loom.com/share/ab53c8f4fb0946f1a7bcefdde8834412?sid=16b1949c-ba32-4150-aba6-93c8727b46c7

rchlfryn:
@busbyk  Following up from the loom (thank you)

- I fixed 500 - it was happening because of the changes in #386 . router params were not defined so I switched to checking the URL. not sure if we want a fallback for this.
- I add helper text to suggest option to display message when the form option is selected.
- Re-added `container` by default so forms will display properly again
- Created unique ids for form- this was a good edge case since we plan to show the a subscribe form on the homepage or blog page and it always in the footer

PR #449, opened 2025-08-24
Add Header block

### LINKED ISSUE #436: Header block
I think we need something like: 

Image

rchlfryn:
### Description
Adds a header block with a wrap in container option for the background color

Wrap in container
Screenshot 2025-08-24 at 16 33 29

Full width
Screenshot 2025-08-24 at 16 33 53

### Issue
Resolves #436


src/blocks/Header/config.ts
```diff
@@ -0,0 +1,45 @@
...
+  AlignFeature,
+  FixedToolbarFeature,
+  HeadingFeature,
+  lexicalEditor,
+} from '@payloadcms/richtext-lexical'
+
+export const Header: Block = {
+  slug: 'header',
+  interfaceName: 'HeaderBlock',
+  imageURL: '/thumbnail/HeaderThumbnail.jpg',
```
busbyk, line 14: I don't think this file got added to this PR?

src/blocks/Header/config.ts
```diff
@@ -0,0 +1,45 @@
...
+  fields: [
+    // Background color
```
busbyk: ⛏️ remove or move to be above the background color field?

src/blocks/Header/config.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+    colorPickerField('Background color'),
+    {
+      name: 'wrapInContainer',
+      admin: {
+        description:
+          'Checking this will render the block with additional padding around it and using the background color you have selected.',
+      },
+      type: 'checkbox',
+      defaultValue: false,
```
busbyk: I wonder if we should default this to true for this particular block since I think the intent is to have this appear like a section's heading and not a full width section itself.

src/blocks/Header/config.ts
```diff
@@ -0,0 +1,45 @@
...
+  fields: [
+    // Background color
+    {
+      name: 'richText',
+      type: 'richText',
+      required: true,
+      editor: lexicalEditor({
+        features: () => {
+          return [
+            FixedToolbarFeature(),
```
busbyk, line 23: I was surprised you use a richText field here but actually I think that's really clever. Because now users can have subtitles and add extra padding via new lines if they want to: 

CleanShot 2025-08-25 at 17 07 01@2x

I do wonder if this will lead to users having unintentional padding or using paragraph tags instead of h2/h3/h4 tags like they should.

busbyk (commented):
I wonder if we would want to allow a user to tighten up the padding/gap between a header block and another block below it:
CleanShot 2025-08-25 at 17 09 12@2x

But probably worry about that later and see how people use this.

rchlfryn:
Merging for now - we can worry about spacing later. Maybe copy figma with spacing options

PR #450, opened 2025-08-25
Attempt to update payload to v3.52.0

### LINKED ISSUE #339: Published At issues for pages and posts
### Still TODO
Pages
 
- publishedAt is set upon refreshing a page in the edit view after publishing (weird)
    -  --> Autosave is to blame https://github.com/payloadcms/payload/issues/10515 . Leaving this open to track 
- update payload to 3.51.0

~### Fixed in #352 _Updated July 29, 2025_

Pages
 
- publishedAt is not set by our seed script
- publishedAt is not set when publishing a page in the edit view

Image

Posts

- unpublishing a post does not clear the publishedAt field

These two collections use different approaches to do the same thing. Would be nice to consolidate.

rchlfryn:
> [!NOTE]
> See comment below why this is closed.
 
### Description
Updates payload to 3.52.0. Takes advantage of such features as
* Soft delete and `groupBy` in collections from ([v3.49.0](https://github.com/payloadcms/payload/releases/tag/v3.49.0))
* Migration for `user_sessions` in ([v3.44.0](https://github.com/payloadcms/payload/releases/tag/v3.44.0))

Only updates plugin-multi-tenant to 3.49.0 to avoid a breaking change in 3.50.0 
> plugin-multi-tenant: constrain results to assigned tenants when present (https://github.com/payloadcms/payload/issues/13365) ([20b4de9](https://github.com/payloadcms/payload/commit/20b4de9))

### Issue
Resolves #339 in v3.51.0


.github/workflows/dependabot.yml
busbyk, line 1: Niiiiiiice!!

src/collections/Media/index.ts
```diff
@@ -22,6 +22,7 @@ export const Media: CollectionConfig = {
     baseListFilter: filterByTenant,
     group: 'Content',
   },
+  trash: true,
```
busbyk, line 25: What happens to these files in Vercel Blob if they're soft deleted? Probably nothing?

Do we need to update any of our queries to respect the trash field? https://payloadcms.com/docs/trash/overview#:~:text=%7D)-,Return%20only%20non%2Dtrashed%20documents%3A,-1

src/payload-types.ts
```diff
@@ -364,6 +366,13 @@ export interface User {
   hash?: string | null;
   loginAttempts?: number | null;
   lockUntil?: string | null;
+  sessions?:
```
busbyk, line 373: What is the effect of this on user sessions?

src/collections/Pages/index.ts
```diff
@@ -93,6 +94,7 @@ export const Pages: CollectionConfig<'pages'> = {
       },
     },
   },
+  trash: true,
```
busbyk, line 98: Same question for this and Posts: if we need to change any of our queries to respect this.

src/collections/Pages/index.ts
```diff
@@ -48,6 +48,7 @@ export const Pages: CollectionConfig<'pages'> = {
   admin: {
     group: 'Content',
     defaultColumns: ['title', 'slug', 'updatedAt'],
+    groupBy: true,
```
busbyk: I think groupBy might be more useful for the Posts collection since users might want to group by Tags.

busbyk (changes requested):
I get this error when visiting the admin panel: 
CleanShot 2025-08-25 at 17 19 25@2x

I was actually able to visit the admin panel initially and then once I clicked into a document and tried to edit a field I got this error. But now I can't access it at all even after deleting my `.next` folder and re-running `pnpm dev`. 

I can recreate the same behavior in the preview environment by visiting any page document: https://update-payload.preview.avy-fx.org/admin/collections/pages/108

I'm assuming the behavior of the multi tenant plugin changed slightly as part of this upgrade. That function _should_ be optional though: https://payloadcms.com/docs/plugins/multi-tenant#:~:text=userHasAccessToAllTenants%3F%3A,176

We don't use that because we determine the tenants a user has access to base on tenant scoped roles and we don't have a tenants relationship on the users collection like the multi tenant plugin expects. That's why we don't just use the multi tenant plugin directly.

rchlfryn:
Closing because we need more time to figure this out.
* 3.48.0 is good but we do not get to use soft delete and groupBy in collections from ([v3.49.0](https://github.com/payloadcms/payload/releases/tag/v3.49.0))
* 3.49.0 is not stable and has a breaking bug for us fixed in 3.50.0
    * deletedAt in TypeWithID not accepting null from generated types (https://github.com/payloadcms/payload/issues/13363) ([8401b21](https://github.com/payloadcms/payload/commit/8401b21))
* 3.50.0 breaks our multi tenant plugin "requiring" `userHasAccessToAllTenants` to be defined [according to the author](https://github.com/payloadcms/payload/issues/13366#issuecomment-3201845249)
    * plugin-multi-tenant: constrain results to assigned tenants when present (https://github.com/payloadcms/payload/pull/13365) ([20b4de9](https://github.com/payloadcms/payload/commit/20b4de9))
    
We should file a payload issue

src/payload-types.ts
rchlfryn: From the release notes:

> ### 🚨 Notice 🚨
> If your project uses the local auth strategy with db-postgres or db-sqlite, a migration is required. This is due to a new security feature (enabled by default) that stores a unique auth session identifier in the database.
>
> To opt out and continue using the previous behavior, you can disable the feature by setting auth.useSessions: false in your users collection config.

https://github.com/payloadcms/payload/releases/tag/v3.45.0

PR #451, opened 2025-08-25
Add documents collection & block

### LINKED ISSUE #413: Ability to upload a PDF
I expected to be able to upload a PDF for [this functionality](https://www.sawtoothavalanche.com/annual-reports/) and either couldn't or couldnt figure it out.

rchlfryn:
### Description
Add documents collection & block. Currently we are just rendering the document as an iframe, but I want to check and see how we want to render it on the FE.

Mentorship-Death-Valley-Avalanche-Center-08-24-2025_08_11_PM

### Issues
Fixes #413


src/blocks/DocumentBlock/Component.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { getMediaURL } from '@/utilities/getURL'
+import { getHostnameFromTenant } from '@/utilities/tenancy/getHostnameFromTenant'
+import { cn } from '@/utilities/ui'
+
+type Props = DocumentBlockProps & {
+  enableGutter?: boolean
+}
+
+export const DocumentBlock = (props: Props) => {
+  const { document, enableGutter = true } = props
```
busbyk: Do we want to move to using `wrapInContainer` over `enableGutter` going forward?

src/blocks/DocumentBlock/config.ts
busbyk, line 1: [Per this discussion](https://github.com/NWACus/web/issues/413#issuecomment-3218722110) yea it might be nice to have options here for how this block gets displayed.

src/collections/Documents/index.ts
busbyk, line 1: It might be nice to use the new folders feature for Documents and Media after the payload upgrade: https://payloadcms.com/docs/folders/overview

src/plugins/index.ts
```diff
@@ -97,6 +97,9 @@ export const plugins: Plugin[] = [
   vercelBlobStorage({
     enabled: !!process.env.VERCEL_BLOB_READ_WRITE_TOKEN,
     collections: {
+      documents: {
+        prefix: getEnvironmentFriendlyName(),
+      },
```
busbyk, line 102: What if a document and a media have the same name in the same tenant? They're stored in the same blob storage so this could happen. I suppose we could avoid this by restricting the mimetypes allowed for Media. 

Right now you can still upload any file type to Media: 
image

src/collections/Documents/index.ts
rchlfryn: It's very buggy in 3.48. We should wait until we update to a newer version

src/blocks/DocumentBlock/Component.tsx
rchlfryn: Yes I think so. Enable gutter is confushing.

PR #454, opened 2025-08-26
Custom domain handling in middleware

busbyk:
## Key changes

- Adds custom domain handling to middleware for tenants with their slug in PRODUCTION_TENANTS env var and with a customDomain field configured
- Adds a redirect from the tenant's subdomain to their custom domain if their slug is in PRODUCTION_TENANTS and they have a customDomain field configured
- ~~Makes remotePatterns for next.js image optimization endpoint permissive and handles allowing specific domains in the middleware. This is necessary because the next.config.mjs is static but the domains we want to allow is dynamic based on configured customDomain fields and the PRODUCTION_TENANTS env var~~
- Includes instructions on custom domain configuration in the onboarding docs. 

## Test locally

- Update your /etc/hosts file to include some custom domain like `127.0.0.1       www.deathvalleyac.local`
- Set the customDomain for `dvac` to `www.deathvalleyac.local:3000`
- Add this domain to the `remotePatterns` in `next.config.js` without the port:
```
      {
        hostname: 'www.deathvalleyac.local',
        protocol: PROTOCOL,
      },
```
- Make sure the Preview Edge Config customDomain value for `dvac` matches this. If you don't have a VERCEL_TOKEN set locally you may need to do this manually in the Vercel console. If you see `VERCEL_EDGE_CONFIG, VERCEL_TOKEN, OR VERCEL_TEAM_ID not available, skipping Edge Config update` in the logs you'll know you need to do this manually. This shouldn't be required in production. 
- Visit http://www.deathvalleyac.local:3000

## Tested in production

I've currently got `www.avy-fx-demo.org` pointed at this PR's preview environment for `dvac`. Check it out. 

Whoever merges this: please update the domain in Vercel to point to our production environment once this is merged into `release`/deployed to prod. We might as well keep avy-fx-demo.org set up pointing at `dvac` as an example.


busbyk:
The failed Vercel build can be ignored. I tried redeploying directly on Vercel after changing the PRODUCTION_TENANTS env var but I think we need to run all of our builds in GHA now. So I just re-ran the preview workflow to redeploy.

rchlfryn (approved):

stevekuznetsov:
>This is necessary because the next.config.mjs is static but the domains we want to allow is dynamic

Can you help me understand the trade-offs here? From the perspective of onboarding a new tenant, I don't know if adding a commit and pushing a new build is sufficiently more or less difficult as compared to editing the env for production in Vercel - it strikes me that both actions would be taken by a dev, not e.g. Rusty. I'd like to understand the downsides of the approach you took for `remotePatterns` to get a better idea of the trade-off we're making here.

stevekuznetsov (approved):
Logic in `middleware.ts` looks reasonable to me.

busbyk:
> > This is necessary because the next.config.mjs is static but the domains we want to allow is dynamic
> 
> Can you help me understand the trade-offs here?  I

Yea for sure. 

> From the perspective of onboarding a new tenant, I don't know if adding a commit and pushing a new build is sufficiently more or less difficult as compared to editing the env for production in Vercel - it strikes me that both actions would be taken by a dev, not e.g. Rusty.

Good point! 

> I'd like to understand the downsides of the approach you took for `remotePatterns` to get a better idea of the trade-off we're making here.

The two options to solve for allowing next.js image requests on custom domains being:
1. Update the `remotePatterns` in `next.config.mjs` as part of the manual steps involved in onboarding a new center
2. or the change in this PR which moves the allowed domains check for next.js image requests to code in `middleware.ts` which uses the customDomain from the edge config/db

Both of these options requires a dev to add the new center's slug to the `PRODUCTION_TENANTS` env var. But option 2 does not require an additional commit. Option 1 still requires the `PRODUCTION_TENANTS` env var change but that's what determines if the middleware correctly rewrites urls for their custom domain. 

The downside that comes to mind for option 2 is that the middleware now executes for next.js image requests instead of letting next.js built-in code handle that. And after some quick testing locally there does seem to be a performance hit for this:

`/_next/image` requests hitting middleware: 
CleanShot 2025-09-22 at 12 06 20@2x

`/_next/image` requests not hitting middleware (i.e. filtered via `remotePatterns`): 
CleanShot 2025-09-22 at 12 07 31@2x

So that's probably significant enough that we'd prefer making a code change to `next.config.js` every time we onboard a new center to production/their custom domain. 

I'll update this PR to remove the handling of `/_next/image` requests in the middleware and add instructions to the onboarding docs to call out adding the custom domain to `remotePatterns`.

docs/onboarding.md
```diff
@@ -22,5 +22,12 @@ We also need to get the new tenant into our Vercel Edge Config. Adding a new ten
 
 1. Update the `PRODUCTION_TENANTS` env var to include the tenant's slug. This should be a comma separated list of slugs (see `src/utilities/tenancy/tenants.ts`).
 2. Set the `customDomain` field for the tenant. This should be the host (i.e. `www.deathvalleyavalanchecenter.com`). Confirm that the `customDomain` in the production Edge Config for the tenant matches this.
-3. Add this domain to our [list of domains](https://vercel.com/nwac/avy/settings/domains) for avy in Vercel.
-4. For domains not managed on Vercel, we'll need to coordinate with the avalanche center to set the appropriate DNS record in their DNS provider's console. Adding a non-Vercel-managed domain will display the records to add and once verified Vercel will automatically generate an SSL certificate. We should include the apex domain -> www. redirect for these custom domains.
+3. Add this domain to `remotePatterns` in `next.config.js`. Example:
```
rchlfryn, line 25: I could easily see this step getting overlooked.. is there any way we can do a check against `customDomain` and the config? do we want that?

busbyk: We could but I don't think we'd want that to run during CI. Hmm I guess maybe we could add it to the `production.yaml` workflow. 

I think we'd know that we missed a step when none of the images are loading on the center's website and hopefully look back at this list to make sure we did everything we needed and then see that we need to add the domain to `remotePatterns`.

stevekuznetsov (approved):
Looks great. Thanks for taking the time to investigate this and re-work the original approach.

PR #477, opened 2025-08-30
Debugging  'cannot read properties of null' and attempt to fix the vercel build error

yuliadub:
I was able to point local build at the deployed dev db to debug. I used the dev db url and generated a token inside of turso and populated that into DATABASE_AUTH_TOKEN env variable. That repro-ed the error right away. 

error
```bash
 ✓ Compiled successfully
   Collecting page data ...
TypeError: Cannot read properties of null (reading 'replace')
    at Array.map (<anonymous>)

> Build error occurred
[Error: Failed to collect page data for /[center]/[...segments]] {
  type: 'Error'
}
```
link to GAH with this error: https://github.com/NWACus/web/actions/runs/17338501404/job/49228723110

I inspected the db, and basically we have a null record for 1 page that is getting processed for each tenant and causing issues since our code assumed slug is never null. The problematic is in the pages table, id 112. 

image

I did not spend much time thinking how to do this the right way, just through this together quickly to test the theory that this will fix things. There is likely a better way to do this but with these fixes I was able to get the build working locally off the dev db. Going to run the development github action next just to make sure it workes in GAH world as well (it should). 

I was seeing a lot of timeout errors when trying to get the build working locally, just a note, not sure if that is normal
```bash
  params: [Array],
  [cause]: [TypeError: fetch failed] {
    [cause]: [AggregateError: ] { code: 'ETIMEDOUT' }
  }
}
```

Havent dug into how or why that empty row got into the db, would be good to do next!

### DIFF
--- a/.github/workflows/development.yaml
+++ b/.github/workflows/development.yaml
@@ -6,7 +6,6 @@ on:
   push:
     branches:
       - 'main'
-      - 'check-next-version'
 jobs:
   deploy:
     runs-on: ubuntu-latest
--- a/.github/workflows/development.yaml
+++ b/.github/workflows/development.yaml
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'check-next-version'
 jobs:
   deploy:
     runs-on: ubuntu-latest
--- a/.github/workflows/production.yaml
+++ b/.github/workflows/production.yaml
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - 'release'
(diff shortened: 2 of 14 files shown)


yuliadub:
I swear I had signed commits enabled but alas it is not working (windows always keep me on my toes) and I wont look into it more tonight. @rchlfryn please fill free to steal this code and close this PR if you look at this before I fix it 😄

rchlfryn:
Closing this PR since it is no longer needed.

PR #485, opened 2025-09-05
General maintenance

### LINKED ISSUE #310: Weather stations map should have a min height
The stations map renders as the remaining viewport height after the navbar but this is quite short on small screens so we probably want to have a minimum height.

rchlfryn:
### Description
* Adds min height to home page NAC widget
  * This is annoying however because there is a nested height calculation on `.nac-map-container` that we can not override because of inline specificity
* Improve style of Avalanche Center selection home page (including default link color - color introduced in #481 )
* Resort admin side bar categories

### Issue
Fixes #310

### DIFF
--- a/src/app/(frontend)/[center]/nac-widgets.css
+++ b/src/app/(frontend)/[center]/nac-widgets.css
@@ -16,20 +16,22 @@
 #widget-container[data-widget='observations'] {
   min-height: 100vh;
 }
-
+#widget-container[data-widget='map'],
 #widget-container[data-widget='stations'] {
   min-height: calc(100vh - 170px) !important;
 }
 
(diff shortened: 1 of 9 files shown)


busbyk:
> Adds min height to home page NAC widget
> * This is annoying however because there is a nested height calculation on `.nac-map-container` that we can not override because of inline specificity

@rchlfryn just FYI that inline height comes from the AFP settings for the avalanche center. They set a fixed height in the AFP and that's what's used by the map widget. That's why I never tried to overwrite that - figured we'd just let the center change the height in the AFP.

Although it might be nice to update the map widget to accept a height parameter in it's config -- I started messing around with that but didn't spend much time on it.

PR #489, opened 2025-09-08
Revert "Add columns to content"

rchlfryn:
Reverts NWACus/web#486 since it creates a breaking deployment


rchlfryn:
No longer needed

PR #515, opened 2025-09-17
Bugfix: filter catch-all segment to published pages only

### LINKED ISSUE #513: Null pages break build
**Describe the bug**
We had an issue in production where a draft page/post with no data added to it breaks builds/ deployments.

**To Reproduce**
1. Navigate to page or post
2. Click on 'Create page/post'
3.  DO NOT ADD ANYTHING
4.  Navigate away
5. Try to deploy/ build (if local) and see error
```bash
...
✓ Compiled successfully
   Collecting page data ...
TypeError: Cannot read properties of null (reading 'replace')
    at Array.map (<anonymous>)

> Build error occurred
[Error: Failed to collect page data for /[center]/[...segments]] {
  type: 'Error'
}
...
```

**Expected behavior**
A draft page with no data should not effect our build

**Screenshots**
A Loom or screenshots are super helpful.

**Environment details**
CI

**Existential musings**
Houses are really just cubes and coverings

busbyk:
Fixes #513 

When we added the published status filter to the page route we forgot to add this to the catch all segment as well. 

## Testing 

I was able to recreate the null page bug by creating a new page, not adding any data, and navigating away. Running a local build threw the expected error:

```
TypeError: Cannot read properties of null (reading 'replace')
    at Array.map (<anonymous>)

> Build error occurred
[Error: Failed to collect page data for /[center]/[...segments]] {
  type: 'Error'
}
```

The change in this PR resolved this error in my local build.


src/app/(frontend)/[center]/[...segments]/page.tsx
```diff
@@ -21,18 +21,24 @@ export async function generateStaticParams() {
...
     },
+    // Need where clause to ignore autosave bug (https://github.com/NWACus/web/pull/204)
+    where: {
+      _status: {
+        equals: 'published',
+      },
+    },
   })
 
   // TODO: use same logic as header to walk the tree and find all pages
```
Comment (busbyk) [inline src/app/(frontend)/[center]/[...segments]/page.tsx:32]: I'll address this TODO in a separate PR since it's unrelated to the #513 bug.

rchlfryn (changes requested):
Empty posts can break the build as well. Can we add this to `src/app/(frontend)/[center]/[slug]/page.tsx` as well?

busbyk:
> Empty posts can break the build as well. Can we add this to `src/app/(frontend)/[center]/[slug]/page.tsx` as well?

This is already there: https://github.com/NWACus/web/blob/main/src/app/(frontend)/%5Bcenter%5D/%5Bslug%5D/page.tsx#L28

And on the posts route as well: https://github.com/NWACus/web/blob/main/src/app/(frontend)/%5Bcenter%5D/blog/%5Bslug%5D/page.tsx#L28

yuliadub:
if we get a fully empty row for a page again, where there everything is null but the id (or is that not possible anymore?) does filtering by publishing handle that case? 

I guess looking back at what was broken, it was unhappy with page.slug being empty and us trying to assume it was not -- but when we filter based on status we dont care what slug is, so I guess it shouldnt matter as long as slug does get filled in when the page becomes published (is there any way slug wont get populated for some reason, say in a case it was null during a draft?)

busbyk:
> if we get a fully empty row for a page again, where there everything is null but the id (or is that not possible anymore?) does filtering by publishing handle that case?

Yea I tested both pages and posts where I created a page but didn't change any data or tried to save it and navigated away so it was completely blank. Does that match the state that y'all were seeing in prod?

> I guess looking back at what was broken, it was unhappy with page.slug being empty and us trying to assume it was not -- but when we filter based on status we dont care what slug is, so I guess it shouldnt matter as long as slug does get filled in when the page becomes published (is there any way slug wont get populated for some reason, say in a case it was null during a draft?)

Just tested on local and I can't publish pages without a slug. The [slugField is required](https://github.com/NWACus/web/blob/bugfix/null-page-build-failure/src/fields/slug/index.ts#L8) and validation runs when you attempt to publish so I think we should be good with this. 

We could add an additional check in the `Page` functions to ensure page.slug is set or add it as a filter to the `queryPageBySlug` and `queryPostBySlug` functions to ensure slug exists if we wanted to be extra careful.

rchlfryn:
LG pending Yuila's feedback.

I think leaving the slug field alone and only requiring on publish is good for now.

rchlfryn (approved):

Issue #495, opened 2025-09-11
Handle redirects at the middleware layer


rustynwac:
Redirects -> move away from the PayloadRedirect component and use our middleware to do redirects (likely use a Vercel KV (redis))
Make sure it doesn't cost us anything to use KV in this way (not worth it)

PR #516, opened 2025-09-17
Respect canonical urls in page route segments

busbyk:
Addresses [this TODO](https://github.com/NWACus/web/blob/main/src/app/(frontend)/%5Bcenter%5D/%5B...segments%5D/page.tsx#L26)

## Key changes
- [...segments]/page.tsx handles any page that is in the navigation
- [...segments/page.tsx's generateStaticParams only generates pages that are in the navigation
- [slug]/page.tsx handles any page not in the navigation
- [slug]/page.tsx's generateStaticParams only generates pages that are not in the navigation

Maintains the redirect to canonical url functionality because dynamicParams is _not_ set to false (i.e. /about-us will redirect to /about/about-us because that page is in the navigation under about). I think we want to keep dynamicParams defaulted to true (i.e. not set) for both of these route segments for this redirect functionality and for PayloadRedirects (even though we may move away from React component-based redirects under #495).


busbyk:
...why didn't CI run for this PR? 😵‍💫

rchlfryn (approved):

PR #522, opened 2025-09-18
Update colors for Sierra rebrand

### LINKED ISSUE #520: Update Sierra colors
Sierra Avalanche's color palette is attached, along with a mockup of the header. Let's plug these colors in and see how they look. 

Here are a few requested color changes SAC would like to see based on the mockup:

Header + Footer Background: Oxford (#0A2238)
Breadcrumbs Background: Alice Blue (#E2ECF2)
Button + Accent Color: Sky Blue (#1480B8)

Image
Image

rchlfryn:
### Description
Update colors for Sierra rebrand
Local-Accident-Reports-Sierra-Avalanche-Center-09-18-2025_02_38_PM

Screenshot 2025-09-18 at 14 38 24

### Issues
Fixes #520

### DIFF
--- a/src/app/(frontend)/colors.css
+++ b/src/app/(frontend)/colors.css
@@ -83,6 +83,7 @@
   --header-foreground-highlight: hsl(240 3.8% 35.3%);
 }
 
+/* Northwest Avalanche Center */
 .nwac {
   --primary: hsl(217 62% 21%);
   --primary-hover: rgba(20, 46, 87, 0.8);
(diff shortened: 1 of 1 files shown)


src/app/(frontend)/colors.css
```diff
@@ -96,46 +97,61 @@
 
   --footer: hsl(300 1% 22%);
   --footer-foreground: hsl(0 0% 100%);
-  --header-foreground-highlight: hsl(0 0% 100%);
+  --footer-foreground-highlight: hsl(0 0% 100%);
```
busbyk, line 100: Nice

src/app/(frontend)/colors.css
busbyk, line 1: The benefit of using hsl vs. hex is that it allows us to use Tailwind's opacity modifiers. If we use hex colors TW's opacity modifiers like [`bg-accent/90`](https://github.com/NWACus/web/blob/main/src/blocks/ImageLinkGrid/Component.tsx#L48) won't work. That's really the only relevant example where we're using opacity modifiers on theme classes but figured I'd call it out. 

That's why I converted hex colors to hsl originally. I think it's probably worth sticking with that so we don't try to use features like opacity modifiers and wonder why they don't work in the future. But let me know what you think.

rchlfryn: TIL - I was just being lazy/ I am more familiar with hex oddly. I will update them to hsl

rchlfryn: The robot fixed it to [modern syntax](https://css-tricks.com/almanac/functions/h/hsl/#aa-syntax)

busbyk (approved):

PR #526, opened 2025-09-19
Built in pages

### LINKED ISSUE #499: Built-in pages references
We need a way to select to built-in pages which are not defined in pages or posts

rchlfryn:
### Description
* Add Built-In Pages collection for super admins
* Removes array from ButtonsBlock (should we change the name?)
* Fixes `ImageLinkGrid` links not working for internal pages/posts
* Refactors
  * Move the page/post selector to a single file - used in `ButtonsBlock`, `LinkPreview` block, `ImageLinkGrid` block, & Quick links on the home page... did not touch NavLinks because of some differences)
  * Move the way we handle the URL from the page/post selector to a single file

### Issues
Fixes #499


src/blocks/Buttons/Component.tsx
```diff
@@ -1,12 +1,10 @@
 import { CMSLink } from '@/components/Link'
 import type { ButtonsBlock as ButtonBlockProps } from 'src/payload-types'
 
-export const ButtonsBlock = ({ buttons }: ButtonBlockProps) => {
+export const ButtonsBlock = ({ button }: ButtonBlockProps) => {
```
busbyk: Should this now be called `ButtonBlock` since it's only one button? I suppose we might need a migration if we did that though.

src/blocks/ImageLinkGrid/Component.tsx
```diff
@@ -28,10 +29,12 @@ export const ImageLinkGrid = (props: Props) => {
           columns?.length > 0 &&
           columns.map((col, index) => {
             const { caption, image, link } = col
+            const href = handleUrl({ ...link })
+            const newTabProps = link?.newTab ? { rel: 'noopener noreferrer', target: '_blank' } : {}
 
             return (
               <div className={cn('group', colsSpanClass, className)} key={index}>
-                <a href={`${link?.url}`}>
+                <a href={href || ''} {...newTabProps}>
```
busbyk: It would be ideal to use a Next `Link` for internal links to take advantage of Next.js client side routing and anchor tags for external links.

src/collections/BuiltInPages/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { populatePublishedAt } from '@/hooks/populatePublishedAt'
+
+export const BuiltInPages: CollectionConfig<'pages'> = {
+  slug: 'builtInPages',
+  labels: {
+    singular: 'Built-In Page',
+    plural: 'Built-In Pages',
+  },
+  access: {
+    ...accessByGlobalRoleWithAuthenticatedRead('builtInPages'),
```
busbyk: I think read should use tenant scoped roles so users can only see their own tenant's pages.

src/utilities/handleUrl.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const handleUrl = ({ url, type, reference }: Props) => {
+  if (!reference?.value || typeof reference.value !== 'object') {
+    return url
+  }
+  if (type === 'internal') {
+    const { relationTo, value } = reference
+
+    if (relationTo === 'builtInPages' && 'url' in value) {
+      return `/${value.url}`
```
busbyk: We're getting double slashes here so links to built-in pages aren't working for me (tested with QuickLinks and ImageLinkGrid): 
CleanShot 2025-09-22 at 13 20 33@2x

I wrote `https://github.com/NWACus/web/blob/built-in-pages/src/utilities/path.ts#L11` for this if you'd like to use that. It has an `ensureLeadingSlash` option too which might be nice here.

src/utilities/handleReferenceURL.ts
busbyk, line 1: This would be a great file to write a few basic tests for.

src/endpoints/seed/index.ts
```diff
@@ -718,6 +717,22 @@ export const seed = async ({
...
 
+  await upsert(
+    'builtInPages',
+    payload,
+    incremental,
+    tenantsById,
+    () => 'builtInPages',
+    Object.values(tenants)
+      .map((tenant): RequiredDataFromCollectionSlug<'builtInPages'>[] => [
+        builtInPage(tenant, 'All Forecasts', '/forecasts/avalanche#/all/'),
```
busbyk: I suppose it doesn't matter for seed data but more of an FYI: single zone centers won't have an all forecasts link - they'll just have an avalanche forecast link to their single zone: https://github.com/NWACus/web/blob/built-in-pages/src/components/Header/utils.ts#L261

src/endpoints/seed/index.ts
```diff
@@ -718,6 +717,22 @@ export const seed = async ({
...
+        builtInPage(tenant, 'Weather Stations', '/weather/stations/map#/all/'),
+        builtInPage(tenant, 'Recent Observations', '/observations#/view/observations'),
+        builtInPage(tenant, 'Submit Observations', '/observations/submit#/view/observations'),
```
busbyk: I think we should let the [WidgetHashHandler](https://github.com/NWACus/web/blob/main/src/components/NACWidget/WidgetHashHandler.client.tsx) component handle setting the hash. Can we remove the hashes from these urls? I know it's just seed data but I don't want future devs to get confused. Eventually we'll move away from the widgets so I think it's nice to use URLs without hashes so in the future we can just delete the WidgetHashHandler and not have to update urls.

busbyk (changes requested):
This is great. Just a few changes requested. Love the decision doc too.

src/endpoints/seed/index.ts
rchlfryn: Good note! I will try to make this dynamic

src/blocks/Buttons/Component.tsx
rchlfryn: 🥲  you are right - will update

src/collections/BuiltInPages/index.ts
rchlfryn: As we discussed - I will make a separate permission for this

src/endpoints/seed/index.ts
rchlfryn: I think this is too much effort to make it dynamic vs noting. It would be nice if NAC could return single zone centers from an api call

src/collections/BuiltInPages/index.ts
```diff
@@ -0,0 +1,45 @@
...
+export const BuiltInPages: CollectionConfig<'pages'> = {
+  slug: 'builtInPages',
+  labels: {
+    singular: 'Built-In Page',
+    plural: 'Built-In Pages',
+  },
+  access: {
+    create: accessByTenantRole('builtInPages')?.create,
+    read: byTenantRole('read', 'builtInPages'),
+    update: byGlobalRole('update', 'builtInPages'),
```
stevekuznetsov, line 23: I'd almost argue that this should be even more locked down, since editing this in the UI is never the right thing to be doing - if we change the set of pages (or the tenant adds a new zone, etc), then something should re-generate this collection automatically, since this data is truly read-only.

stevekuznetsov: (not sure it needs to be literally read-only or something, but having edit powers as a user account seems wrong)

stevekuznetsov (commented):
Are we adding the method to populate this collection for tenants and read the NAC API to know their zones, etc?

src/endpoints/seed/index.ts
busbyk: Fine by me! We could extract the logic in the Header utils into a separate function at some point.

rchlfryn:
> Are we adding the method to populate this collection for tenants and read the NAC API to know their zones, etc?

Not at this time. We thought manually entering the zones would be less costly than the dev work that would go into it.


Update - looks like @busbyk took care of the method in #531

src/collections/BuiltInPages/index.ts
rchlfryn: Ideally yes, and these pages would be connected to the navigation and we could even update the admin UI to reflect that, but similar answer to the one below. We thought this was a good in-between decision since we are only touching these pages on onboarding, it wasn't a good enough reason to spend dev time on it.

src/endpoints/seed/index.ts
busbyk: You don't need to update this but I wrote a getActiveForecastZones function for this PR: https://github.com/NWACus/web/pull/531

busbyk:
> > Are we adding the method to populate this collection for tenants and read the NAC API to know their zones, etc?
> 
> Not at this time. We thought manually entering the zones would be less costly than the dev work that would go into it.
> 
> Update - looks like @busbyk took care of the method in #531

Well we'd still need to figure out when/where to run a script to ensure that all of the built-in pages have documents in the built-in pages collection. 

You and I talked about the [`init` hook](https://payloadcms.com/docs/configuration/overview#config-options:~:text=More%20details.-,onInit,-A%20function%20that) but I think that runs every time a new serverless function is executed since Payload is re-initialized. So I don't think we'd want that performance hit. 

Running something after deployment to ensure the appropriate built-in pages collection documents exist is what we're looking for so a custom endpoint that our GHA workflows hit after deployment might be a good solution. 

It just doesn't feel worth it for the Sierra Launch milestone which is due Sept 30. But obviously we do want a better solution here in the long-term. So let's create a follow up issue for that.

rchlfryn:
> > > Are we adding the method to populate this collection for tenants and read the NAC API to know their zones, etc?
> > 
> > 
> > Not at this time. We thought manually entering the zones would be less costly than the dev work that would go into it.
> > Update - looks like @busbyk took care of the method in #531
> 
> Well we'd still need to figure out when/where to run a script to ensure that all of the built-in pages have documents in the built-in pages collection.
> 
> You and I talked about the [`init` hook](https://payloadcms.com/docs/configuration/overview#config-options:~:text=More%20details.-,onInit,-A%20function%20that) but I think that runs every time a new serverless function is executed since Payload is re-initialized. So I don't think we'd want that performance hit.
> 
> Running something after deployment to ensure the appropriate built-in pages collection documents exist is what we're looking for so a custom endpoint that our GHA workflows hit after deployment might be a good solution.
> 
> It just doesn't feel worth it for the Sierra Launch milestone which is due Sept 30. But obviously we do want a better solution here in the long-term. So let's create a follow up issue for that.

Issue created #533

busbyk:
> > > > Are we adding the method to populate this collection for tenants and read the NAC API to know their zones, etc?
> > > 
> > > 
> > > Not at this time. We thought manually entering the zones would be less costly than the dev work that would go into it.
> > > Update - looks like @busbyk took care of the method in #531
> > 
> > 
> > Well we'd still need to figure out when/where to run a script to ensure that all of the built-in pages have documents in the built-in pages collection.
> > You and I talked about the [`init` hook](https://payloadcms.com/docs/configuration/overview#config-options:~:text=More%20details.-,onInit,-A%20function%20that) but I think that runs every time a new serverless function is executed since Payload is re-initialized. So I don't think we'd want that performance hit.
> > Running something after deployment to ensure the appropriate built-in pages collection documents exist is what we're looking for so a custom endpoint that our GHA workflows hit after deployment might be a good solution.
> > It just doesn't feel worth it for the Sierra Launch milestone which is due Sept 30. But obviously we do want a better solution here in the long-term. So let's create a follow up issue for that.
> 
> Issue created #533

Thank you!

__tests__/client/handleURL.client.test.ts
busbyk, line 1: 🧡

docs/onboarding.md
```diff
@@ -16,4 +16,14 @@ We also need to get the new tenant into our Vercel Edge Config. Adding a new ten
 
 - [ ] Create the new tenant in the production admin panel
 - [ ] Fill out the new tenant's website settings
+- [ ] Create Built-In pages
+
+  | Title | URL |
+  |--------|--------|
+  | All Forecasts | /forecasts/avalanche|
+  | Zone > [ZONE NAME] | /forecasts/avalanche/[ZONE] |
```
busbyk: Can we add a caveat here: if it's a single zone center there should only be one link here which would be:
Title: Avalanche Forecast
URL: /forecasts/avalanche/[ZONE]

src/collections/BuiltInPages/index.ts
```diff
@@ -0,0 +1,45 @@
...
+import { populatePublishedAt } from '@/hooks/populatePublishedAt'
+
+export const BuiltInPages: CollectionConfig<'pages'> = {
+  slug: 'builtInPages',
+  labels: {
+    singular: 'Built-In Page',
+    plural: 'Built-In Pages',
+  },
+  access: {
+    create: accessByTenantRole('builtInPages')?.create,
```
busbyk: I think we only want global roles to be able to create built-in pages right?

src/migrations/20250922_185258_add_built_in_pages.ts
busbyk, line 1: I don't see anything in here about the ButtonsBlock rename. Did that not need a migration? I was wondering if the rename of that block would result in lost data or not.

docs/onboarding.md
rchlfryn: I will update this table to look like the one in #533

src/collections/BuiltInPages/index.ts
rchlfryn: Bless it - I knew there was a reason I didnt copy pasta... thanks

src/migrations/20250922_185258_add_built_in_pages.ts
rchlfryn: I reran the create migration script after renaming it and was also surprised it didn't pick anything up.

Upon looking further, it does not have a table. I think since it is a nested block (not in `RenderBlocks`) it doesn't get a table?

busbyk: Ohhh yes it's because we only ever use it in the Lexical BlocksFeature and not in a Blocks field like you said. OK checks out!

src/collections/BuiltInPages/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  labels: {
+    singular: 'Built-In Page',
+    plural: 'Built-In Pages',
+  },
+  access: {
+    create: ({ req }) => {
+      const tenantFromCookie = getTenantFromCookie(req.headers, 'number')
+      return tenantFromCookie ? byGlobalRole('create', 'builtInPages')({ req }) : false
+    },
+    read: byTenantRole('read', 'builtInPages'),
```
busbyk, line 22: Can you just add this permission to a tenant role (maybe all of them?) in the seed data?

busbyk (approved):
Approved pending updating the seed data. 

We should make sure to update the tenant roles in production once this is deployed to have builtInPages read permissions.

PR #530, opened 2025-09-22
Clone parentMeta in generatMeta_ functions

busbyk:
I started getting an error thrown on page and post routes: `Cannot add property rel, object is not extensible`. 

I believe this is due to Next.js attempting to add a rel property to an immutable object. I wonder if this is due to #509 and I just didn't notice this locally when I was testing for some reason. Potentially due to cached pages. 

@rchlfryn this is the issue we saw in my local when we paired the other day. I know you weren't getting this error locally but I wonder if you hadn't pulled the latest from `main`? Can you try to recreate this error on latest `main` locally?

Visit http://sac.localhost:3000/education/learn and you should get an error boundary with that error.

### DIFF
--- a/src/utilities/generateMeta.ts
+++ b/src/utilities/generateMeta.ts
@@ -8,7 +8,7 @@ import { getHostnameFromTenant } from './tenancy/getHostnameFromTenant'
 import { resolveTenant } from './tenancy/resolveTenant'
 
 function cloneParentMeta(parentMeta?: Metadata) {
-  return parentMeta ? structuredClone(parentMeta) : {}
+  return parentMeta ? JSON.parse(JSON.stringify(parentMeta)) : {}
 }
 
 export const generateMetaForPage = async (args: {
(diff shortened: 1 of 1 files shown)


rchlfryn (commented):
~I am still not seeing the error. I am using Arc (which is built off Chromium Engine Version 140.0.7339.81)~

I was able to recreate it with a fresh `pnpm i` so it is related to #509

src/utilities/generateMeta.ts
```diff
@@ -37,15 +37,18 @@ export const generateMetaForPage = async (args: {
     ? `${doc.meta?.title} | ${parentTitle ?? tenant?.name}`
     : doc.meta?.title
 
+  // Deep clone parent metadata to avoid mutation issues
+  const clonedParentMeta = parentMeta ? JSON.parse(JSON.stringify(parentMeta)) : {}
```
rchlfryn: ⛏️  Make this a single utility function in this file to be able to reuse below

rchlfryn (approved):

PR #531, opened 2025-09-22
Generate static params for centers' active zones

### LINKED ISSUE #378: Hit the NAC API for forecast zones in generateStaticParams
This TODO: https://github.com/NWACus/web/blob/revalidate-everything/src/app/(frontend)/%5Bcenter%5D/forecasts/avalanche/%5Bzone%5D/page.tsx#L25

Let's add `export const dynamicParams = false` here too for appropriate 404s.

busbyk:
Resolves #378 

Abstracts the active zones filtering logic into it's own helper function so it's shared with the top level nav logic.

### DIFF
--- a/src/components/Header/utils.ts
+++ b/src/components/Header/utils.ts
@@ -1,6 +1,10 @@
 import { Navigation, Page, Post } from '@/payload-types'
-import { getAvalancheCenterMetadata, getAvalancheCenterPlatforms } from '@/services/nac/nac'
-import { AvalancheCenter, AvalancheCenterPlatforms } from '@/services/nac/types/schemas'
+import {
+  ActiveForecastZoneWithSlug,
+  getActiveForecastZones,
+  getAvalancheCenterPlatforms,
+} from '@/services/nac/nac'
+import { AvalancheCenterPlatforms } from '@/services/nac/types/schemas'
 import configPromise from '@payload-config'
 import { unstable_cache } from 'next/cache'
 import { getPayload } from 'payload'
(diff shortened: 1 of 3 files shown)


src/components/Header/utils.ts
```diff
@@ -247,60 +251,50 @@ export const getTopLevelNavItems = async ({
...
+      forecastsNavItem = {
+        link: {
+          label: 'Avalanche Forecast',
+          type: 'internal',
+          url: `/forecasts/avalanche/${activeForecastZones[0].slug}`,
+        },
+      }
+    } else {
+      const zoneLinks: NavItem[] = activeForecastZones
+        .sort((zoneA, zoneB) => (zoneA.zone.rank ?? Infinity) - (zoneB.zone.rank ?? Infinity))
```
rchlfryn, line 265: how is `rank` defined? any rational in choosing that over alphabetical?

src/services/nac/nac.ts
```diff
@@ -130,10 +130,10 @@ export async function getAllAvalancheCenterCapabilities() {
 
 export async function getAvalancheCenterPlatforms(centerSlug: string) {
   const allAvalancheCenterCapabilities = await getAllAvalancheCenterCapabilities()
-  const actualAvyCenter = centerSlug === 'dvac' ? 'nwac' : centerSlug
+  const actualAvyCenterSlug = centerSlug === 'dvac' ? 'nwac' : centerSlug
```
rchlfryn: 👍

rchlfryn: @busbyk do you remember why we picked `actualAvyCenterSlug` instead of `avyCenterSlug`? was it only because of `dvac`?

busbyk: I think the original intent was just to have a different name than centerSlug because we're handling the special case of `dvac`. I can't think of an appropriate name here. It's not really the _actual_ avy center slug - kind of the opposite. 

`centerSlugToUse` could be descriptive enough?

src/components/Header/utils.ts
busbyk: Rank comes from the AFP. I believe the rank is set in the AFP (I don't have permission in AFP dev to see those settings unfortunately - I can request that from Chris to better understand this). But the zones aren't ordered by rank in the results (at least that's what I observed). 

So I was sorting by this to respect that rank. This is the same functionality we had before.

src/services/nac/nac.ts
rchlfryn: Up to you! I remember I choose the name but it is kind of confusing.

rchlfryn (approved):
Approved pending possible variable name change

src/services/nac/nac.ts
busbyk: I went for `centerSlugToUse`

PR #532, opened 2025-09-22
Warnings widget typo fix

### LINKED ISSUE #254: Avalanche Warning and watches widget displays above the forecast
Image

-Warnings are shown in the Warning Widget (a standalone banner). If there are multiple zones with active warnings, they're listed on the banner - but it's still one banner.
-Warnings are also shown in the Avalanche Forecast Widget as a banner on top of the zone avalanche forecasts. The widget knows which zone to show it in vs which one to not show it in

busbyk:
Resolves #254 

There was a typo in the key used for the warnings widget on the window object. The warnings widget uses singular and plural naming in the `afp-public-widget` repo. 

CleanShot 2025-09-22 at 18 56 02@2x

## To test locally
Set https://github.com/NWACus/web/blob/warnings-widget/src/components/NACWidget/index.tsx#L114 to `devMode: true`. I've created a warning and a watch for NWAC in the dev AFP.

### DIFF
--- a/src/app/(frontend)/[center]/page.tsx
+++ b/src/app/(frontend)/[center]/page.tsx
@@ -48,7 +48,7 @@ export default async function Page({ params }: Args) {
     <>
       <NACWidget
         center={center}
-        widget="warning"
+        widget="warnings"
         widgetsVersion={version}
         widgetsBaseUrl={baseUrl}
       />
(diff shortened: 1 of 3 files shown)


rchlfryn (approved):
🥲

PR #535, opened 2025-09-23
Generic embed block sizing

### LINKED ISSUE #493: Embed Blocks have issues with height and width
When creating an embed block, it seems that the height and width have to be specified in the embed code for the site to respect them. Otherwise, they default to very small. It also seems like relative widths (100%) do not work and you have to set a fixed width which is then no longer responsive. 

**To Reproduce and See**
See examples on [this page](https://sac.avy-fx.org/weather-station-table)

**Expected behavior**
Width defaults to fit the space available unless specified and is responsive.
Height respects the "Height of embed (in px)" field

**Environment details**

- Desktop - firefox and chrome 
- Android - firefox and chrome

busbyk:
Resolves #493 

## Explanation of current approach 

The GenericEmbedBlock has two fundamental rendering scenarios:
1. There are iframe/script tags in the `html` field
2. There are not iframe/script tags in the `html` field

The reason for this is related to the logic required to automatically set the height of the embed container based on the embed content. The GenericEmbedBlock component tries to determine the height of all DOM content rendered from the `html` field. There's a resize script embedded along with the content to accomplish this. But there are many things that could cause this height to be determined incorrectly (think collapsing margins or content that loads late). 

A few reasons iframes/scripts make this even harder:
- iframes load a whole separate DOM and have a much higher likelihood of their content taking a long time to load, it's likely that our resize script runs before the iframe content is fully loaded and sets the height to less than that of the DOM content.
- iframes often have their height and width set by the provider of the iframe code which we would be messing with. 
- iframes often set their own height based on loaded content
- There may be multiple iframes in the `html` field - some might have heights set, some might not

So I took the approach that if there's an iframe or script tag in the `html` field - the provider of that iframe/script has probably set a specific height for that content to be displayed at (most do) and the resize script does not run. When there are not iframes/scripts in the `html` field, the content is rendered in a sandboxed iframe and we determine the height of the content using the resize script. 

Addressed by this PR

> Height respects the "Height of embed (in px)" field
This field is for the _container_ of the embed, not the elements inside of it. Since there could be multiple iframes in the `html` field - I don't think it would make sense to use this field to set the height of those iframes since they may all need different heights. 

I actually wonder if the "Height of embed (in px)" field should be a max height property instead of height. @rchlfryn I'd love to chat with you about this further / understand why this was added. 

I think it's reasonable to require a height and width on iframes. The default size that browsers render iframes at when no height or width is specified is 300px x 150px. That's why iframes are rendering at that size if they don't have height/width specified. 

## Changes in this PR
- Fixes the container for rendering scenario 1 to be full width of it's parent so iframes can have percentage widths too
- Adds `overflow-x: hidden`
- Fixes alignment to work with the container being full width now
- Adds helper text to the html field
- Removes the hardcoded height in the embed block (also from the footer form). Keeps the 'auto' initial height prior to the resize script running. 

## Thoughts
~~There are other things we could do here, like breaking out iframes into their own block and adding custom validation (i.e. only one iframe tag allowed, must have dimensions set, etc.). We could run the resize script for `html` fields with scripts too. We could add some helper text to explain the differences in the rendering scenarios.~~


rchlfryn (approved):
LGTM - did we want to keep the height field? or remove it?

busbyk:
> LGTM - did we want to keep the height field? or remove it?

Oh yea - well what was the original need for it? I remember you mentioning something about the footer embed? 

We could have a max height to be used for the embed _container_.

busbyk:
> > LGTM - did we want to keep the height field? or remove it?
> 
> Oh yea - well what was the original need for it? I remember you mentioning something about the footer embed?
> 
> We could have a max height to be used for the embed _container_.

I ended up removing height completely. I played around with max height but I don't think it makes much sense with the resize script at play too. A user can always set a max height on their content.

rchlfryn (approved):
I think it was a good call to remove the height completely

PR #543, opened 2025-09-24
Update sponsor blocks - add carousel and individual options

### LINKED ISSUE #238: Blog - sponsors
Add sponsor block to individual blog page

### LINKED ISSUE #534: Sponsors carousel and individual view

### LINKED ISSUE #556: Sponsor padding seems huge
**Describe the bug**
Do we need this much spacing in the sponsor block? 

**To Reproduce**

1. Go to https://sac.avy-fx.org/support/sponsors
2. See error

**Expected behavior**
I would expect the logos to sit tighter together. 

**Screenshots**

Image

**Environment details**

- Desktop chrome

**Existential musings**
If you drive on snow/ice, you must have snow tires.

rchlfryn:
### Description
- Add carousel 
Kapture 2025-09-24 at 13 25 58

 
- Add banner options ( fade or static if only one sponsor is shown)
Screenshot 2025-09-26 at 11 39 01

> [!TIP]
>  use https://sponsor-blocks-2.preview.avy-fx.org/dvac/about/agency-partners for testing

### Issue
Resolves #534 
Relsoves https://github.com/NWACus/web/issues/556


src/blocks/SponsorsBlock/config.tsx
```diff
@@ -8,10 +8,32 @@ export const defaultFields: Field[] = [
     type: 'text',
   },
   colorPickerField('Title background color'),
+  {
+    type: 'radio',
+    name: 'sponsorsLayout',
+    label: 'Choose a layout for the sponsors logos',
+    defaultValue: 'dynamic',
```
busbyk: dynamic doesn't map to anything - should we default to static?

src/blocks/SponsorsBlock/config.tsx
```diff
@@ -8,10 +8,32 @@ export const defaultFields: Field[] = [
...
+      {
+        label: 'Static',
+        value: 'static',
+      },
+      {
+        label: 'Carousel (ideal for 6+ sponsors)',
+        value: 'carousel',
+      },
+      {
+        label: 'Individual',
```
busbyk: Should we add something that explains the behavior of the individual option here? It's not obvious that it will cycle through sponsors selected using a fade animation or just display one if only one is selected.

src/blocks/SponsorsBlock/components/Carousel.tsx
```diff
@@ -0,0 +1,45 @@
...
+export const SponsorsBlockCarousel = ({ sponsors }: { sponsors: Sponsor[] }) => {
+  return (
+    <div className="w-11/12">
+      <Carousel
+        opts={{
+          align: 'start',
+          loop: true,
+        }}
+        plugins={[
+          Autoplay({
```
busbyk, line 47: I think we should set [`stopOnMouseEnter`](https://www.embla-carousel.com/plugins/autoplay/#stoponmouseenter) to be true to avoid this potentially annoying behavior: 

Image

The autoplay would then start up again when the user's cursor is no longer over the element. Thoughts?

src/blocks/SponsorsBlock/config.tsx
```diff
@@ -8,10 +8,32 @@ export const defaultFields: Field[] = [
     type: 'text',
   },
   colorPickerField('Title background color'),
+  {
+    type: 'radio',
+    name: 'sponsorsLayout',
```
busbyk: A single sponsor renders as two different sizes with static vs. individual:

static:

Image

individual: 

Image

Do we want that?

How can a user make a sponsors image larger? I could see a center wanting to make a wide image like the Mammut logo a lot larger - like 75% of page width for example. We might need additional configuration for individual sponsor image sizes to account for different image ratios.

src/blocks/SponsorsBlock/components/Individual.tsx
```diff
@@ -0,0 +1,50 @@
...
+        </a>
+      ) : (
+        <Carousel
+          opts={{
+            align: 'start',
+            loop: true,
+          }}
+          plugins={[
+            Autoplay({
+              delay: 10000,
```
busbyk: This delay feels a little long to me. Can we get the rest of the team's opinion during check-in? "feels a little long" is pretty subjective 😄

src/blocks/SponsorsBlock/Component.tsx
```diff
@@ -12,35 +14,22 @@ type SponsorsBlockComponentProps = SponsorsBlockProps & {
...
-        return 'w-1/2 md:w-1/2 lg:w-1/4'
-      default:
-        return 'w-1/3 md:w-1/3 lg:w-1/5'
-    }
-  }
-  const colSpanName = getColSpanName()
-
   return (
     <div className={cn(wrapInContainer, 'py-10')}>
       {title && (
```
busbyk: Should the user be able to wrap the title in a container like the header block? 

Right now it's always full width:
Image

busbyk (commented):
My biggest hang up here is the difference between static and individual (for users choosing which layout to use) and how a user could make a single sponsor's image much larger (bc I think that's a desire from Sierra's side right?). 

Not sure of the details of the original ask here since the linked issue is blank so let me know if I'm not aware of some requirements here too. 

The carousel looks really nice! I think shadcn/ui recently added this right? Good timing.

src/blocks/SponsorsBlock/config.tsx
rchlfryn: I am going to remove this block option for now. We will likely add it back in when we address #238

src/blocks/SponsorsBlock/components/Carousel.tsx
rchlfryn: Fixed in [f3ac63a](https://github.com/NWACus/web/pull/543/commits/f3ac63aee6728f92f6ee99d6d98669ec2285f05c)

src/blocks/SponsorsBlock/config.tsx
rchlfryn: Ignoring since we removed this option

src/blocks/SponsorsBlock/components/Individual.tsx
rchlfryn: Ignoring since we removed this

src/blocks/SponsorsBlock/Component.tsx
rchlfryn: I originally has `wrapInContainer` only on the lexical editor (but it wasnt working). `wrapInContainer` is now always available.

rchlfryn:
Use https://sponsor-blocks-2.preview.avy-fx.org/dvac/about/agency-partners for testing

busbyk (approved):
LGTM

PR #544, opened 2025-09-24
Redirects updates

### LINKED ISSUE #332: Redirects don't work as expected within tenant scope
It would be good to test a few additional scenarios but I tried adding a redirect for a path `/some-path` which one would expect to redirect when visiting a tenant's site like `nwac.avy-fx.org/some-path` but the actual URL at that path is `/nwac/some-path` because we rewrite the url in our middleware. 

We also don't support redirects at the root of the project. Not sure if we actually want that but worth exploring as part of this issue. 

Make sure to handle revalidation for redirects since currently redirects happen in React and even though the redirects fetch is cached using unstable_cache the path will be fully cached still, potentially causing stale redirects. 

I wonder if we want to move redirects into the middleware and use a KV store (i.e. Vercel's redis).

busbyk:
Resolves #332 

## Key changes
- Removes `@payloadcms/plugin-redirects` in favor of our own Redirects collection. The Payload plugin was super lightweight and we have our own re-usable field for internal and external links. So we might as well just write this ourselves. 
- Renames the `PayloadRedirects` component to `Redirects`
- Updates the logic in the `Redirects` component to be center-aware
- Adds validation to the Redirects collection
- No longer supports redirects for valid paths.
  - If a path matches a page (nested in the navigation or not) or post, no redirect will occur
  - The validation in the Redirects collection will not allow creating a redirect for a path that matches a page or post path
- Re-orders sponsors in seed script's initial delete to avoid not null errors in standalone mode + adds builtInPages collection for same reason (unrelated but needed to fix these to run standalone seed).

## Questions
- ~~Should we also restrict users from creating Redirects for draft state pages and posts? Just to avoid the potentially confusing scenario where the redirect stops working after a page is published?~~ Yes, we'll restrict draft pages/posts too. 
- Should we also restrict users from creating Redirects for our built-in pages? Those pages don't currently have the `Redirects` component on them so no redirect will occur anyways.


src/utilities/collectionFilters.ts
```diff
@@ -1,14 +1,23 @@
...
-  tenant: {
-    equals: data.tenant,
-  },
-})
+export const getTenantFilter = ({ data, req }: FilterOptionsProps) => {
+  let tenantId = data.tenant
+
+  if (!tenantId) {
+    tenantId = getTenantIdFromCookie(req.headers)
+  }
```
busbyk, line 13: This was needed for the redirects collection. For some reason, the tenant isn't available in the data provided so I'm falling back to retrieving it from the cookie.

rchlfryn:
>Should we also restrict users from creating Redirects for draft state pages and posts? Just to avoid the potentially confusing scenario where the redirect stops working after a page is published?

Yes - please!

> Should we also restrict users from creating Redirects for our built-in pages? Those pages don't currently have the Redirects component on them so no redirect will occur anyways.

I think we should allow users to redirect to our built-in pages (as long as it will not effect the hash) I think the chances of this happening are slim, so I could see an argument for not doing it if it is a lot of work

busbyk:
> > Should we also restrict users from creating Redirects for draft state pages and posts? Just to avoid the potentially confusing scenario where the redirect stops working after a page is published?
> 
> Yes - please!

I like it. On it. 

> > Should we also restrict users from creating Redirects for our built-in pages? Those pages don't currently have the Redirects component on them so no redirect will occur anyways.
> 
> I think we should allow users to redirect to our built-in pages (as long as it will not effect the hash) I think the chances of this happening are slim, so I could see an argument for not doing it if it is a lot of work

Cool, we can currently redirect _to_ built-in pages but I was wondering if we should restrict a user from using a built-in page's path in the from field. I'm leaning towards yes.

src/migrations/index.ts
```diff
@@ -151,9 +153,19 @@ export const migrations = [
     down: migration_20250925_144212_remove_meta_title.down,
     name: '20250925_144212_remove_meta_title',
   },
+  {
```
rchlfryn: It looks like these files have the same migration which will cause an issue. I ran into this when working with sponsors. The best fix I found was completely dropping any commit with a migration, deleting the preview db and re running `migrate:create`

I am not sure which step made this work. I didnt check in between the rebase and db drop.

src/collections/Redirects/hooks/validateFrom.ts
```diff
@@ -0,0 +1,209 @@
...
+        equals: tenantId,
+      },
+    },
+  })
+
+  if (existingRedirects.length > 0) {
+    throw new Error('From URL must be unique.')
+  }
+
+  /** ensure from URL does not match a published page or post path */
```
rchlfryn, line 42: Since we added built-in pages, do we want to check for that here as well?

src/collections/Redirects/hooks/validateFrom.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  console.log(value.split('/').pop())
```
rchlfryn:
```suggestion
```

src/collections/Redirects/hooks/validateFrom.ts
```diff
@@ -0,0 +1,209 @@
...
+import { resolveTenant } from '@/utilities/tenancy/resolveTenant'
+import { FieldHook } from 'payload'
+import invariant from 'tiny-invariant'
+
+export const validateFrom: FieldHook = async ({ value, data, req: { payload } }) => {
+  const tenantId = data?.tenant
+
+  invariant(typeof value === 'string', 'From URL is invalid.')
+  invariant(typeof tenantId === 'number', 'Tenant not selected.')
+
```
rchlfryn, line 18: Maybe add a quick check up here that it starts with a `/`

src/collections/Redirects/hooks/validateFrom.ts
```diff
@@ -0,0 +1,209 @@
...
+  if (existingRedirects.length > 0) {
+    throw new Error('From URL must be unique.')
+  }
+
+  /** ensure from URL does not match a published page or post path */
+
+  console.log(value.split('/').pop())
+
+  if (value.split('/').slice(1).length === 1) {
```
rchlfryn: I am curious about this logic. If a user inputs `/about/about-us`, this will return false even though there is an about page that exists. What am I missing?

src/collections/Redirects/hooks/validateFrom.ts
```diff
@@ -0,0 +1,209 @@
...
+
+      if (matchingDraftPages.length > 0) {
+        throw new Error(
+          `This From URL matches a draft page nested in the navigation "${matchingDraftPages[0].title}". If you want to redirect this From URL please change that page's slug or location in the navigation.`,
+        )
+      }
+    }
+  }
+
+  if (value.startsWith('/blog/')) {
```
rchlfryn: ⛏️ There is talk about renaming blog to feed for some tenants. I wonder if we want to make this dynamic, or just know we will have to account for this in the future if the name change happens.

This is a much bigger question but just putting it here as a comment. No need to reply.

rchlfryn (changes requested):
Thanks for adding the built-in pages stuff. A couple of comments

src/collections/Redirects/hooks/validateFrom.ts
busbyk: For others reading (we already discussed): there is a check for navigation-nested urls in the else block for this if statement.

rchlfryn (approved):

PR #547, opened 2025-09-25
Bugfix: favicons weren't loading for center-scoped routes

busbyk:
Hardcoded <link> tags were unnecessary and the metadata for center-scoped routes was using the website settings' thumbnail size but that's only generated for images over a certain size which icons often are not. So falling back to just the url attribute.

### DIFF
--- a/src/app/(frontend)/[center]/layout.tsx
+++ b/src/app/(frontend)/[center]/layout.tsx
@@ -123,7 +123,7 @@ export async function generateMetadata({ params }: Args): Promise<Metadata> {
     const cacheTag = settings.icon.updatedAt
 
     iconImgSrc = getMediaURL(
-      settings.icon.sizes?.thumbnail?.url,
+      settings.icon.sizes?.thumbnail?.url || settings.icon.url,
       cacheTag,
       getHostnameFromTenant(tenant),
     )
--- a/src/app/(frontend)/layout.tsx
+++ b/src/app/(frontend)/layout.tsx
@@ -116,10 +116,6 @@ export default async function RootLayout({ children }: Args) {
 
   return (
     <html className={cn(lato.variable)} lang="en" suppressHydrationWarning>
-      <head>
-        <link href="/favicon.ico" rel="icon" sizes="96x96" />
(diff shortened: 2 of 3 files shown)


src/app/(frontend)/layout.tsx
```diff
@@ -116,10 +116,6 @@ export default async function RootLayout({ children }: Args) {
 
   return (
     <html className={cn(lato.variable)} lang="en" suppressHydrationWarning>
-      <head>
-        <link href="/favicon.ico" rel="icon" sizes="96x96" />
```
rchlfryn, line 120: We might also need to remove this in this file
https://github.com/NWACus/web/blob/main/src/app/%5B...segmentsNotFound%5D/layout.tsx#L106-L109

src/app/(frontend)/[center]/layout.tsx
```diff
@@ -123,7 +123,7 @@ export async function generateMetadata({ params }: Args): Promise<Metadata> {
     const cacheTag = settings.icon.updatedAt
 
     iconImgSrc = getMediaURL(
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/layout.tsx:125]: ⛏️ can we change this variable name to `faviconImgSrc`?

rchlfryn (changes requested):

src/app/(frontend)/layout.tsx
busbyk: Ooo good call.

src/app/(frontend)/[center]/layout.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/layout.tsx:125]: So technically... a `<link rel="icon">` could get used for more than just a favicon. We don't have a manifest right now of course but it would get used for the PWA icon and I believe if there's an absence of an `apple-touch-icon` a `rel="icon"` will be used.

Is that getting too much into the semantics? 😬 

Am I nitpicking your nitpick? 😄

Comment (rchlfryn) [reply src/app/(frontend)/[center]/layout.tsx:125]: I did see that. Makes sense. I thought it would be nice to find favicon somehow but that's fine.

rchlfryn (approved):

PR #551, opened 2025-09-25
Bugfix: decode special characters from URL in breadcrumbs

### LINKED ISSUE #517: Breadcrumbs don't handle special characters
**Describe the bug**
Breadcrumbs don't handle special characters

**To Reproduce**

1. Go to a page with a special character in the slug like: https://snfac.avy-fx.org/forecasts/avalanche/galena-summit-&-eastern-mtns#/galena-summit-%26-eastern-mtns/
4. The breadcrumb renders the url encoded version of the special character

**Expected behavior**
The actual special character is rendered. 

**Screenshots**
Image

**Existential musings**
Hamburgers seem like a very American food but they're actually from Hamburg, Germany

This is low priority since I think our slug generation logic should strip out most special characters but we might as well handle this just in case.

busbyk:
Resolves #517

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.client.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.client.tsx
@@ -52,18 +52,19 @@ const processNestedSegments = (
 
 export function Breadcrumbs() {
   const segments = useSelectedLayoutSegments()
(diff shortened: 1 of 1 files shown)


rchlfryn (approved):
TIL about `decodeURIComponent`

PR #559, opened 2025-09-26
Add formatting options to MediaBlock

### LINKED ISSUE #465: Media block changes
Need options for changing the size, dimension and scale of images within the media block.

Preferred UI:

Selectors-

Size:
Full/existing/Inset/Inline

Dimensions:
4:3/16:9/1:1

Image

busbyk:
Resolves #465 

## Key changes
- Uses similar patterns to other blocks for MediaBlock
	- Adds wrapInContainer and exports a different config for lexical vs blocks fields
	- Uses the same padding as other full page blocks (all of our other blocks use `py-10` but Media was using `my-4`)
	- Adds alignContent field
	- Adds backgroundColor field
- Adds new imageSize field
	- Uses responsive sizes instead of percentages to maintain responsive images:
		- Intrinsic (the actual size of the image)
		- Small
		- Medium
		- Large
		- Full width
- Removes caption from Media collection and adds it to the MediaBlock config since a user might want a different caption to appear next to the same image used in different locations.
- Adds variations of the media block to the /blocks seed page

## Screenshots
CleanShot 2025-09-26 at 13 29 57@2x
CleanShot 2025-09-26 at 13 37 13


src/fields/imageSize.ts
```diff
@@ -0,0 +1,22 @@
+import { Field } from 'payload'
+
+export const imageSizeField: (label: string) => Field = (label) => ({
+  name: 'imageSize',
+  type: 'select',
+  label,
+  required: false,
+  defaultValue: 'intrinsic',
+  options: [
+    { label: 'Intrinsic (Natural size)', value: 'intrinsic' },
```
rchlfryn: `Original` might be better here?

src/endpoints/seed/home-page.ts
```diff
@@ -149,6 +149,9 @@ export const homePage: (
       {
         media: heroImage.id,
         blockType: 'mediaBlock',
+        backgroundColor: 'white',
```
rchlfryn: Can we make this `transparent`?

rchlfryn (changes requested):
I think as part of this we should either leverage the [image sizes](https://github.com/NWACus/web/blob/main/src/collections/Media/index.ts#L72-L102) we are creating images or remove them. Right now there is a lot of code in [ImageMedia](https://github.com/NWACus/web/blob/main/src/components/Media/ImageMedia/index.tsx) trying to use image sizes correctly, but I am unsure if we are even passing the prop down so it might just doing a lot of work for nothing.
Take a look and let me know what you think.

src/blocks/MediaBlock/config.ts
```diff
@@ -1,15 +1,55 @@
...
+    type: 'upload',
+    relationTo: 'media',
+    required: true,
+  },
+  {
+    name: 'caption',
+    type: 'richText',
+    editor: lexicalEditor({
+      features: ({ rootFeatures }) => {
+        return [...rootFeatures, InlineToolbarFeature()]
```
rchlfryn: I dont think we want to include `rootFeatures` here. Can we limit it to just 
```
ParagraphFeature(),
UnderlineFeature(),
BoldFeature(),
ItalicFeature(),
LinkFeature(),
FixedToolbarFeature(),
AlignFeature(),
InlineToolbarFeature(),
```

src/endpoints/seed/home-page.ts
busbyk: Woops, good catch.

src/fields/imageSize.ts
busbyk: Works for me.

busbyk:
> I think as part of this we should either leverage the [image sizes](https://github.com/NWACus/web/blob/main/src/collections/Media/index.ts#L72-L102) we are creating images or remove them. Right now there is a lot of code in [ImageMedia](https://github.com/NWACus/web/blob/main/src/components/Media/ImageMedia/index.tsx) trying to use image sizes correctly, but I am unsure if we are even passing the prop down so it might just doing a lot of work for nothing. Take a look and let me know what you think.

This is a great point. I want to do this as part of a separate PR. Here's why: 

The current sizes generation is arbitrary - it's setting the suggested size to twice the size of the breakpoint. Next.js is using this sizes hint to generate the srcset used which is using Next Image's built-in image optimizations. 

But we're actually not even utilizing the different imageSizes generated by the Media collection 🤯

So there's two issues: our sizes hint is inaccurate and we're not even using the different image sizes generated by Payload in our srcsets. What we need to do is set up our sizes hint to tell Next Image to use our different image sizes which are actually different image urls instead of relying on Next.js image optimization. 

I played around with this a bit today and I want to spend more time on it to do it right but I also want to focus on the Sierra Launch this week. There's nothing functionally wrong with our sizes being off it's just inefficient and not doing what we expected it was doing -- images will display at the right size. 

I created an issue: https://github.com/NWACus/web/issues/563

rchlfryn (approved):
Thanks for making that other issue!

PR #560, opened 2025-09-29
Add progress bar on front end & increase upload file size

### LINKED ISSUE #381: Add a global loading indicator to (frontend)
Payload has a nice global loading indicator that fires on transitions (this is probably [nprogress](https://www.npmjs.com/package/nprogress)). This might be nice for the (frontend) too. 

This can actually be nice for sites with such heavy caching as we'll have because sometimes page transitions are so fast you don't realize something actually changed if the page is really similar 😅 

But would also be nice for dynamic pages like the blog index page and eventually the events index page. 

Something to discuss if we want. 

See GitHub's global loading indicator at the top (the blue bar that expands): 
Image

### LINKED ISSUE #453: Allow file uploads greater than 4.5MB
There is currently a [4.5MB file size limit for server uploads for the Vercel Blob Storage adapter](https://payloadcms.com/docs/upload/storage-adapters#:~:text=for%20each%20collection.-,When%20deploying%20to%20Vercel%2C%20server%20uploads%20are%20limited%20with%204.5MB.%20Set%20clientUploads%20to%20true%20to%20do%20uploads%20directly%20on%20the%20client.,-1).

We can enable client uploads to allow file uploads for files larger than that. 

Image

rchlfryn:
#### Add progress bar to front end.
I did some snooping on discord and found where Payload released their [`ProgressBar`](https://discord.com/channels/967097582721572934/967097582721572937/1339360118067826769). It looks like they built it fro scratch. Importing `ProgressBar` did not work, but a few message down someone mentioned https://github.com/TheSGJ/nextjs-toploader. It is a well maintained wrapper of `nprogress` (which is mentioned in the issue) for nextjs apps. I vote we use this rather than trying to write our own.

### Issue
Resolves #381 
Resolves #453

### DIFF
--- a/src/app/(frontend)/[center]/layout.tsx
+++ b/src/app/(frontend)/[center]/layout.tsx
@@ -70,7 +70,7 @@ export default async function RootLayout({ children, params }: Args) {
   return (
     <NotFoundProvider>
       <TenantProvider tenant={tenant}>
-        <NextTopLoader showSpinner={false} />
+        <NextTopLoader color="#3982e8" showSpinner={false} />
         <PostHogTenantRegister />
         <AvalancheCenterProvider platforms={platforms} metadata={metadata}>
           <div className={cn('flex flex-col min-h-screen', center)}>
--- a/src/app/(frontend)/[center]/layout.tsx
+++ b/src/app/(frontend)/[center]/layout.tsx
@@ -17,6 +17,7 @@ import { getHostnameFromTenant } from '@/utilities/tenancy/getHostnameFromTenant
 import { resolveTenant } from '@/utilities/tenancy/resolveTenant'
 import { cn } from '@/utilities/ui'
 import configPromise from '@payload-config'
+import NextTopLoader from 'nextjs-toploader'
 import { getPayload } from 'payload'
 import invariant from 'tiny-invariant'
 import './nac-widgets.css'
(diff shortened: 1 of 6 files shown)


src/app/(frontend)/[center]/layout.tsx
```diff
@@ -69,6 +70,7 @@ export default async function RootLayout({ children, params }: Args) {
   return (
     <NotFoundProvider>
       <TenantProvider tenant={tenant}>
+        <NextTopLoader showSpinner={false} />
```
Comment (busbyk) [inline src/app/(frontend)/[center]/layout.tsx]: Should we set the color of the progress bar to a center's theme color? Maybe the `header-foreground-highlight` or `secondary`?

src/app/(frontend)/layout.tsx
```diff
@@ -118,6 +119,7 @@ export default async function RootLayout({ children }: Args) {
     <html className={cn(lato.variable)} lang="en" suppressHydrationWarning>
       <body>
         <PostHogProvider>
+          <NextTopLoader showSpinner={false} />
```
busbyk: Should we set this progress bar's color to the Avy brand blue? I used `#1E1E1E` as the background blue for the opengraph images but I can't find the Avy figma at the moment.

busbyk (approved):
Looks good to me!

Just a few thoughts.

src/app/(frontend)/[center]/layout.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/layout.tsx]: Sadly I couldn't figure out how to use tailwind colors

PR #561, opened 2025-09-29
Update config for web settings and biographies

### LINKED ISSUE #309: Cannot provide name when Biography linked to User

### LINKED ISSUE #553: Separate footer info in setting
Move Footer form to its own tab in Settings

rchlfryn:
### Description
- Move `Footer Form` settings to its own tab
- Remove relation between biographies and user (we decided we do not need this)

### Issues
Resolves #553 
Resolves #309

### DIFF
--- a/src/collections/Settings/components/USFSLogoDescription.tsx
+++ b/src/collections/Settings/components/USFSLogoDescription.tsx
@@ -1,6 +1,6 @@
-import { UploadFieldDescriptionServerComponent } from 'payload'
+import { TextFieldDescriptionServerComponent } from 'payload'
 
-export const USFSLogoDescription: UploadFieldDescriptionServerComponent = () => {
+export const USFSLogoDescription: TextFieldDescriptionServerComponent = () => {
   return (
     <p className="field-description field-description-description">
       Upload your USFS logo if applicable. This will be displayed to the right of your banner logo.
--- a/src/collections/Settings/components/USFSLogoDescription.tsx
+++ b/src/collections/Settings/components/USFSLogoDescription.tsx
@@ -1,6 +1,6 @@
-import { TextFieldDescriptionServerComponent } from 'payload'
+import { UploadFieldDescriptionServerComponent } from 'payload'
 
-export const USFSLogoDescription: TextFieldDescriptionServerComponent = () => {
+export const USFSLogoDescription: UploadFieldDescriptionServerComponent = () => {
   return (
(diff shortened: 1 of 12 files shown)


src/collections/Settings/components/USFSLogoDescription.tsx
```diff
@@ -1,6 +1,6 @@
-import { UploadFieldDescriptionServerComponent } from 'payload'
+import { TextFieldDescriptionServerComponent } from 'payload'
 
-export const USFSLogoDescription: UploadFieldDescriptionServerComponent = () => {
+export const USFSLogoDescription: TextFieldDescriptionServerComponent = () => {
```
busbyk: I think the types are functionality the same but this does technically get used as the description for an upload field haha. Any reason for the change?

src/collections/Settings/index.ts
```diff
@@ -106,224 +313,28 @@ export const Settings: CollectionConfig = {
...
-              },
-              admin: {
-                components: {
-                  Description:
-                    '@/collections/Settings/components/USFSLogoDescription#USFSLogoDescription',
-                },
-              },
-            },
-          ],
+          fields: [...brandAssetsFields],
```
busbyk: ⛏️ These don't need to be spread in an array - could just be `fields: brandAssetsFields`. Same for generalFields and socialMediaFields.

src/collections/Users/index.ts
```diff
@@ -94,6 +94,7 @@ export const Users: CollectionConfig = {
       type: 'date',
       admin: {
         readOnly: true,
+        position: 'sidebar',
```
busbyk, line 97: Oo good call

busbyk (approved):
👏

src/collections/Settings/components/USFSLogoDescription.tsx
rchlfryn: Oh I misread this is `UploadFieldServerComponent`.. I will change it back

src/collections/Settings/index.ts
rchlfryn: Great call!

PR #562, opened 2025-09-30
Style updates - blogs and buttons block

### LINKED ISSUE #537: Blog post featured image sizing and effects
Two issues that I noticed:
1. We use a hover effect for blog post preview components that has the initial scale value of the image at 105% and then animates it to 100% on hover. This means that a portion of the image is cropped to allow this hover effect. 
2. If the image is wider than it is tall, part of the image is cropped. 

This wasn't an issue for our seed data images because they were more of a pattern that didn't look bad when cropped. But all of the featured images used for the [Daily Flow tag for Sierra](https://sac.avy-fx.org/admin/collections/posts?limit=10&columns=%5B%22title%22%2C%22slug%22%2C%22updatedAt%22%2C%22-id%22%2C%22-tenant%22%2C%22featuredImage%22%2C%22-description%22%2C%22-content%22%2C%22-authors%22%2C%22-publishedAt%22%2C%22-tags%22%2C%22-relatedPosts%22%2C%22-createdAt%22%2C%22-_status%22%5D&where%5Bor%5D%5B0%5D%5Band%5D%5B0%5D%5Btags%5D%5Bin%5D%5B0%5D=14&page=1) are informational graphics / icons that shouldn't be cropped. They are also all wider than they are tall. 

So for these images specifically we would want to make sure they are not cropped at all and a scaling hover animation also might not make sense - we could try an initial state of 100% and shrink to 95% or something along those lines. 

Solving for the Daily Flow tag images is simple enough:
1. Make images use `object-fit: cover;` 
2. Remove or modify the scaling hover animation

We'll just want to make sure this looks good for other images too. 

An alternative solution would be too enforce that featured images are cropped to square but I don't think we can do that with Payload at the moment. I believe you can only enforce dimensions for a whole collection (i.e. Media). This would be worth looking into.

### LINKED ISSUE #558: Two button blocks in a row render too close
**Describe the bug**
There's two buttons in a column that are stacking without enough padding

**To Reproduce**

1. Go to Sierra home page
2. See error

**Expected behavior**
I would expect a little bit of padding 

**Screenshots**

Image

Image

**Environment details**

- Desktop Chrome
- iOS DuckDuckGo

**Existential musings**
You can be anything in this world but if you aren't first a good hang, the rest is won't matter.

rchlfryn:
### Descriptions
**Blog updates**
* Fixes main blog view if blog card does not have a description
* Adds proper margin between blog cards and sidebar with filters
* Adds border around both image and blog card because it looked weird only being around the card with a white image.

Kapture 2025-09-29 at 18 59 58
Blog-Northwest-Avalanche-Center-09-29-2025_06_59_PM

**Button block updates**
* Adds y margin around button incase buttons are back to back
* Fixes label - if a label is input, it will be used, otherwise it will be taken from the title of the reference page. If the link is external, the user is forced to enter a label in `validateLabel`

Pages
Screenshot 2025-09-29 at 19 00 50

Homepage (nothing different about this rendering. It was just where it was mentioned in the ticket)
Screenshot 2025-09-29 at 19 02 03

### Issues
Resolves #537 
Resolves #558

### DIFF
--- a/src/app/(frontend)/[center]/blog/page.tsx
+++ b/src/app/(frontend)/[center]/blog/page.tsx
@@ -75,7 +75,7 @@ export default async function Page({ params, searchParams }: Args) {
 
   return (
     <div className="pt-4">
-      <div className="container md:max-lg:max-w-5xl mb-16 flex flex-col-reverse md:flex-row flex-1">
+      <div className="container md:max-lg:max-w-5xl mb-16 flex flex-col-reverse md:flex-row flex-1 gap-10 md:gap-16">
         <div className="grow">
           <Suspense fallback={<div>Loading posts...</div>}>
             <PostCollection posts={posts.docs} />
@@ -83,7 +83,7 @@ export default async function Page({ params, searchParams }: Args) {
         </div>
 
         {/* Sorting and filters */}
-        <div className="flex flex-col gap-16 shrink-0 justify-between md:justify-start md:w-[240px] lg:w-[300px]">
+        <div className="flex flex-col shrink-0 justify-between md:justify-start md:w-[240px] lg:w-[300px]">
           <Suspense fallback={<div>Loading filters...</div>}>
             <PostsSort initialSort={sort} />
             {tags.docs.length > 1 && <PostsTags tags={tags.docs} />}
--- a/src/app/(frontend)/[center]/blog/page.tsx
+++ b/src/app/(frontend)/[center]/blog/page.tsx
@@ -83,7 +83,7 @@ export default async function Page({ params, searchParams }: Args) {
         </div>
 
         {/* Sorting and filters */}
-        <div className="ms-16 flex flex-col gap-4 shrink-0 justify-between md:justify-start md:w-[240px] lg:w-[300px]">
+        <div className="flex flex-col gap-16 shrink-0 justify-between md:justify-start md:w-[240px] lg:w-[300px]">
           <Suspense fallback={<div>Loading filters...</div>}>
             <PostsSort initialSort={sort} />
             {tags.docs.length > 1 && <PostsTags tags={tags.docs} />}
--- a/src/app/(frontend)/[center]/blog/page.tsx
+++ b/src/app/(frontend)/[center]/blog/page.tsx
@@ -75,7 +75,7 @@ export default async function Page({ params, searchParams }: Args) {
 
   return (
     <div className="pt-4">
-      <div className="container md:max-lg:max-w-5xl mb-16 flex flex-col-reverse md:flex-row flex-1 gap-6">
+      <div className="container md:max-lg:max-w-5xl mb-16 flex flex-col-reverse md:flex-row flex-1">
         <div className="grow">
           <Suspense fallback={<div>Loading posts...</div>}>
             <PostCollection posts={posts.docs} />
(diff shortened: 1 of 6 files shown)


src/blocks/Button/Component.tsx
```diff
@@ -3,7 +3,7 @@ import type { ButtonBlock as ButtonBlockProps } from 'src/payload-types'
 
 export const ButtonBlockComponent = ({ button }: ButtonBlockProps) => {
   return (
-    <div>
+    <div className="my-4">
```
busbyk, line 6: I wonder if it would make more sense to add a typography plugin style that adds the same bottom padding as the `p` tag to `div`s too. That would handle this same issue for other blocks potentially too and keep this button block without extra margin if used in other places. 

Thoughts?

src/app/(frontend)/[center]/blog/page.tsx
```diff
@@ -58,15 +58,15 @@ export default async function Page({ params, searchParams }: Args) {
...
+      <div className="container md:max-lg:max-w-5xl mb-16 flex flex-col-reverse md:flex-row flex-1">
         <div className="grow">
           <Suspense fallback={<div>Loading posts...</div>}>
             <PostCollection posts={posts.docs} />
           </Suspense>
         </div>
 
         {/* Sorting and filters */}
-        <div className="flex flex-col gap-4 shrink-0 justify-between md:justify-start md:w-[240px] lg:w-[300px]">
+        <div className="ms-16 flex flex-col gap-4 shrink-0 justify-between md:justify-start md:w-[240px] lg:w-[300px]">
```
Comment (busbyk) [inline src/app/(frontend)/[center]/blog/page.tsx]: TIL about inline margin utilities.

But I'd used gap here because that applies for flex-row and flex-col and works on mobile. This adds inline-start margin on mobile too which we don't want: 

Image

src/components/Link/index.tsx
```diff
@@ -35,6 +35,8 @@ export const CMSLink = (props: CMSLinkType) => {
   } = props
 
   const href = handleReferenceURL({ url: url, type: type, reference: reference })
+  const referenceTitle = (reference && (reference?.value as BuiltInPage | Page | Post).title) || ''
```
busbyk: I think we should check that reference.value is not `string | number` here instead of casting. Based on the types above, reference could be string | number. Although I do wonder if it could actually be a string since we use number ids. It could definitely be a number if depth isn't set correctly somewhere.

src/components/PostPreview/index.tsx
```diff
@@ -46,7 +46,7 @@ export const PostPreview = (props: {
             />
           )}
         </div>
-        <div className="flex flex-col justify-between px-6 py-4 flex-grow @md:border-y border-b border-x rounded-b-lg @md:rounded-b-none @md:border-r @md:rounded-tr-lg @md:rounded-br-lg">
+        <div className="flex flex-col justify-between px-6 py-4 flex-grow border-t md:border-t-0 md:border-l">
```
busbyk: We should use container queries here since this component can be embedded in differently constrained parent elements and we can reliably resize it based on the viewport width but rather it's container width. 


```suggestion
        <div className="flex flex-col justify-between px-6 py-4 flex-grow border-t @md:border-t-0 @md:border-l">
```

busbyk (changes requested):

src/blocks/Button/Component.tsx
rchlfryn: I think this might be too much adding it to all `div`s. This is done in the `tailwind.config`. I will leave it for now and we can always adjust.

We might just want to define something like `block-container` and add it too all the blocks

busbyk (approved):

PR #566, opened 2025-10-01
Navigation top level nav items enabled toggle

### LINKED ISSUE #554: Add ability to hide blog and events
ACs want the ability to hide the blog and events landing pages. There are also 

Let's add an enabled/disabled toggle to top level nav items in the navigations collection, excluding AFP-powered tabs (forecasts, observations, weather).

busbyk:
Resolves #554 

## Key changes
- Changes `topLevelNavItem` field:
  - `isConfigurable` -> `hasConfigurableNavItems`
  - Adds `hasEnabledToggle`
  - Removes `isVirtual` based on `isConfigurable` -- this was problematic for non-configurable top level nav items that had an enabled toggle in options because the tab wouldn't be persisted to the db. `isVirtual` is still set to true for top level nav items that don't have options or configurable nav items (i.e. forecasts and observations)
- Nests the enabled toggle in an options group for visual display reasons but also in case we want to have additional options in the future
- Respects the enabled toggle in `getTopLevelNavItems` which cascades down to the canonical url logic throughout the app so no additional logic was necessary for `generateStaticParams` in `[...segments]/page.tsx`, for example. 
- Respects the enabled toggle on the /blog route


src/collections/Navigations/fields/topLevelNavTab.ts
```diff
@@ -1,27 +1,48 @@
...
       },
     }),
-  ],
-})
+  ]
+
+  if (hasEnabledToggle) {
+    const enabledToggleField: Field = {
+      type: 'group',
+      name: 'options',
```
rchlfryn, line 32:
```suggestion
      name: 'options',
      label: '',
```
Consider hiding the label since there is only one option

src/collections/Navigations/fields/topLevelNavTab.ts
```diff
@@ -1,27 +1,48 @@
...
+      fields: [
+        {
+          type: 'checkbox',
+          defaultValue: true,
+          name: 'enabled',
```
rchlfryn, line 38: Can add a label to change this to something like `Show menu/ menu item in navigation`?

src/collections/Navigations/fields/topLevelNavTab.ts
```diff
@@ -1,27 +1,48 @@
-import { Tab, toWords } from 'payload'
+import { Field, Tab, toWords } from 'payload'
 import { itemsField } from './itemsField'
 
 export const topLevelNavTab = ({
   name,
   description,
-  isConfigurable = true,
+  hasConfigurableNavItems = true,
+  hasEnabledToggle = true,
```
rchlfryn, line 8: I expected the default  for this to be `false`. Right now, if a developer adds a nav item but forgets to set `hasEnabledToggle` to false in the navigations collection config, the toggle will show up automatically. It seems "safer" to require developers to enable it intentionally rather than risk exposing it by accident.

src/components/Header/utils.ts
```diff
@@ -63,7 +63,7 @@ function topLevelNavItem({
   label: string
   enabled?: boolean
 }): TopLevelNavItem[] {
-  if (!tab || !enabled) {
+  if (!tab || !enabled || tab.options?.enabled === false) {
```
rchlfryn: is `enabled` doing anything anymore? I don't see it being passed as a prop in any of the function calls

rchlfryn (changes requested):

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: Ope - you're totally right. I'll update.

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: Ahh yes! Forgot we could just pass an empty label.

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: Oh how about a description? Because it does more than just show the menu item - for blog for example it results in /blog 404ing.

src/components/Header/utils.ts
busbyk: Ope, nope meant to remove, thanks!

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: Oh wait I take that back. There are only two tabs that shouldn't have an enabled toggle. Rusty wanted to be able to disable all top level nav items. The only ones that don't make sense to control with a toggle are forecasts and observations because those render based on whether or not they're turned on in the AFP via the capabilities API.

src/collections/Navigations/fields/topLevelNavTab.ts
busbyk: Went for a label plus description.

rchlfryn (approved):

PR #570, opened 2025-10-02
Remove @payloadcms/plugin-multi-tenant

### LINKED ISSUE #552: Migrate off of the Payload multi-tenant plugin components/functions
`@payloadcms/plugin-multi-tenant` is currently a blocker to upgrading Payload which we want to do for #339 and just to stay up to date in general. See: https://github.com/NWACus/web/pull/450#issuecomment-3235518130

It makes sense to move off the Payload multi-tenant plugin because there is a key difference in how we do multi-tenancy and how the plugin expects us to be doing multi-tenancy. 

The plugin expects there to be a `tenants` array on a user. But we determine association between a user and a tenant via rules-based, tenant-scoped roles. So it seems likely that we'll run into more issues in the future and I don't think it makes sense to request the Payload team to support our use case since they've made a clear decision on how tenancy is expected to work with their plugin. 

Right now, we're only using components and helper functions from their plugin anyways. We can easily copy and paste these into our own repo from a version that works for us and maintain our own divergent copies. We may want to make updates to these based on our approach (although they'll likely stay very similar at first). 

The work here should actually be pretty minimal - mostly just making sure that we maintain functionality. 

There might be a gotcha if there are many dependencies that we need to pull over from the plugin but I think we can manage.

rchlfryn:
### Description 
* Remove @payloadcms/plugin-multi-tenant and adds the files we need locally
* Move `TenantSelector` to plugin folder

### Issue
Resolves #552


src/plugins/multiTenant/README.md
rchlfryn, line 1: What am I missing here?

stevekuznetsov: nada

stevekuznetsov (approved):
Mechanically looks right

src/providers/TenantSelectionProvider/index.client.tsx
```diff
@@ -0,0 +1,231 @@
...
+            value: doc.id,
+          })),
+        )
+
+        if (result.totalDocs === 1) {
+          setSelectedTenantID(result.docs[0].id)
+          setCookie(String(result.docs[0].id))
+        }
+      }
+    } catch (e) {
```
busbyk, line 164: Can you remove `(e)` or use it to capture the actual error?

src/components/GlobalViewRedirect/index.ts
busbyk, line 1: Could this just go in `@/components`?

src/fields/tenantField/index.scss
```diff
@@ -0,0 +1,14 @@
...
+    width: 25%;
+  }
+  &__hr {
+    width: calc(100% + 2 * var(--gutter-h));
+    margin-left: calc(-1 * var(--gutter-h));
+    background-color: var(--theme-elevation-100);
+    height: 1px;
+    margin-bottom: var(--spacing-field);
+  }
+}
```
busbyk, line 14: We never display this field right? Could we remove these styles if they'll never be applied? Maybe we do actually want them for when the debug option is true in case we ever want that locally?

src/fields/tenantField/TenantFieldComponent.tsx
busbyk, line 1: Could this folder go in `@/fields`?

src/providers/TenantSelectionProvider/index.tsx
busbyk, line 1: Could this folder go in `@/providers`?

src/plugins/multiTenant/TenantSelector/TenantSelector.tsx
busbyk, line 1: Could this folder go in `@/components`?

src/utilities/tenancy/findTenantOptions.ts
busbyk, line 1: `@/utilities/tenancy/findTenantOptions`?

src/plugins/multiTenant/README.md
busbyk: Can we make this a decision doc instead?

src/providers/TenantSelectionProvider/index.client.tsx
```diff
@@ -0,0 +1,231 @@
...
+  setModified: () => undefined,
+  setTenant: () => null,
+  syncTenants: () => Promise.resolve(),
+  updateTenants: () => null,
+})
+
+export const TenantSelectionProviderClient = ({
+  children,
+  initialValue,
+  tenantCookie,
```
busbyk: This never gets used, can we remove?

busbyk (changes requested):
Do we need to colocate the code we're taking from the multi tenant plugin in it's own `@/plugins` directory? I'm assuming that we're going to start modifying this code as some point and was thinking that we would integrate it fully into our codebase. So using our current organizational structure with fields, components, providers, utilities, etc. 

I'm just thinking about how multi tenancy is core to our application vs. being truly a plugin. Thoughts?

I left comments on each file where I think they should go and the rest in the `@/plugins/multiTenant` folder could go in `@/utilities/tenancy`.

src/plugins/multiTenant/TenantField/index.tsx
rchlfryn: `@/fields` is intended for `PayloadField` configs but I will move it to components

src/fields/tenantField/index.scss
rchlfryn: We do use this field as the field component - it should make more sense now with the code in the proper folders

rchlfryn:
> Do we need to colocate the code we're taking from the multi tenant plugin in it's own @/plugins directory? I'm assuming that we're going to start modifying this code as some point and was thinking that we would integrate it fully into our codebase. So using our current organizational structure with fields, components, providers, utilities, etc.
>
> I'm just thinking about how multi tenancy is core to our application vs. being truly a plugin. Thoughts?
>
> I left comments on each file where I think they should go and the rest in the @/plugins/multiTenant folder could go in @/utilities/tenancy.

I thought we wanted to try to maintain it as a plugin, but it makes sense to fold it into our current structure. I moved everything into their respective folders.

busbyk (approved):
Looks good!

Are we still getting that behavior with the GlobalViewRedirect where there's an infinite redirect loop when a document doesn't exist yet for a tenant?

rchlfryn:
> Are we still getting that behavior with the GlobalViewRedirect where there's an infinite redirect loop when a document doesn't exist yet for a tenant?

I am not sure. I removed a home page locally for an AC and it did not loop. We might see it again when we add a new AC

PR #571, opened 2025-10-02
Upgrade to payload v3.58.0 & add view doc button 

### LINKED ISSUE #339: Published At issues for pages and posts

### LINKED ISSUE #382: Link to live page in document view
We have a link to the live preview, a link to the preview in a separate tab but we don't have a link to the published, live page/post/etc. 

I think it would be nice to have two buttons: 
1. A link to the published, live page
2. A button to copy the published, live page url (a convenience thing but I bet people will want to do this)

### LINKED ISSUE #565: Padding at the bottom of Post pages look a little tight
**Describe the bug**
If you look at the bottom of a published post, there isn't much padding between the end of the page and the footer

**To Reproduce**
Look at any SAC posts

**Expected behavior**

**Screenshots**

Image

Image

**Environment details**
- Desktop chrome

**Existential musings**
Was sitting here tryna think of something then ended up being late for a meeting. Sry k thanks bye

rchlfryn:
### Description
- Upgrade payload to 3.58.0 (no major changes in release notes)
- Add view page/post button that links to published version
  
Screenshot 2025-10-02 at 16 24 06

- Re-addes padding to bottom of posts before related posts

### Issues
Resolves #339 
Resolves #382 
Resolves #565


src/collections/Pages/components/ViewPageButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+    depth: 2,
+    where: {
+      id: {
+        equals: id,
+      },
+    },
+  })
+  const page = pageRes.docs[0]
+  const pageTenant = await resolveTenant(page.tenant)
+  const canonicalUrl = (await getCanonicalUrlForSlug(pageTenant.slug, page.slug)) || ''
```
stevekuznetsov: In the case where the `await` doesn't throw, why can we get an empty result and have to `|| ''`? Will a `ViewDocumentButton` with an empty URI go to some nice 404/oops page?

rchlfryn: Good call - I was just trying to please the linter. I changed it to return `null` if`canonicalUrl` is empty

src/components/ViewDocumentButton/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <button
+      aria-label="View published"
+      className="view-page-btn"
+      id="preview-button"
+      onClick={openNewTab}
+      title="View published"
+      type="button"
+    >
+      <Globe width={16} height={16} className="icon" />
```
rchlfryn, line 26: No clue which icon to use here...

busbyk: I think Globe makes sense!

src/components/ViewDocumentButton/index.scss
busbyk, line 1: No issue with using css modules here but we can use Tailwind classes in our Payload components, just fyi.

src/components/Footer/Footer.tsx
```diff
@@ -60,7 +60,7 @@ export async function Footer({ center }: { center?: string }) {
 
   const currentYear = new Date().getFullYear()
   return (
-    <footer className="mt-auto border-t border-border bg-footer text-footer-foreground">
+    <footer className="mt-auto bg-footer text-footer-foreground">
```
busbyk, line 63: 👏

busbyk (changes requested):
https://www.loom.com/share/7f384a922ef84324aeb98ce5ce4ede6f?sid=5ec5470e-7f7b-45d2-953e-3e16b54f1cbe

rchlfryn:
@busbyk Great call on the preview mode.. I added the same [code](https://github.com/payloadcms/payload/blob/main/examples/draft-preview/src/components/AdminBar/index.tsx#L53-L60) `Exit preview mode` is using. Let me know if you had something else in mind

busbyk:
Not sure when this changed but the standalone-seed script now needs an `await` (see this change I had to make in the migration safety PR: https://github.com/NWACus/web/pull/584/files#diff-ca970e298417e124a0a6cd435bf4192e84284997ea7cb6c3d6130cb6166b5954). This preview env doesn't have any seed data - the seed script doesn't run now. Maybe due to our last Payload version bump and we just never noticed? 

Want to add that here so other PRs that have to merge this change in will get seed data?

busbyk (approved):

Issue #455, opened 2025-08-26
Deeply nested block revalidation


busbyk:
Our [revalidation](https://github.com/NWACus/web/blob/main/docs/revalidation.md) has a limitation where it will not find relationships deeply nested in richText fields. 

Here's a scenario where revalidation would fail: Use a MediaBlock in the richText field (belowHeaderContent) of the BlogListBlock. Updating the MediaBlock would not cause a revalidation of pages that the BlogListBlock is embedded in. 

This is primarily because we can't query Lexical nodes. The Payload Local API doesn't support it and the query would be very complex because of the amount of nesting. To get around this for Posts we have a `blocksInContent` field that's hidden in the admin panel that has a `beforeChange` hook that finds all of the relationship fields in the Lexical content and stores them in an easily queried format. We can then query that field during revalidation to determine if a given relationship is on a given route. 

I think the solution here might be to create a reusable `richText` field that has a sibling `blocksInContent` field (potentially with unique naming for the field it's associated with? 🤔) so that every `richText` field will always have it's relationship fields stored in an associated `blocksInContent` field. We can then use this expectation to determine deep nesting of relationships.

busbyk:
Note from team: consider limiting really deep nesting.

busbyk:
Moving this to Summer 2026 and marking as tech debt.

stevekuznetsov:
It is not clear what the issue actually is - are there cases that `blocksInContent` doesn't catch? Is it not applied everywhere? The description sounds like it should fix/band-aid the issue.

busbyk:
> It is not clear what the issue actually is - are there cases that `blocksInContent` doesn't catch? Is it not applied everywhere? The description sounds like it should fix/band-aid the issue.

Does this clarify the issue? https://github.com/NWACus/web/blob/main/docs/decisions/011-incremental-static-regeneration.md#deeply-nested-relationships

PR #573, opened 2025-10-04
Update ISR settings

### LINKED ISSUE #496: ISR settings updates
Update the use of dynamic = 'force-static' and revalidate = 600 -- decide on reasonable default revalidate times and determine if we need 'force-static'

rchlfryn:
## Description

This PR addresses ISR for the application by documenting, updating settings, fixing some missing functionality, and addressing the known limitation of our revalidation system. 

## Related Issues

Resolves #496 
Addresses #455 but does not close it. Keeping it open for the future to enable on-demand ISR even for deeply nested blocks. 

## Key Changes

- Documents ISR settings decisions
- Updates ISR settings in a few places
  - Removes time-based revalidation from built-in pages
  - Adds time-based revalidation to individual posts pages
  - Adds time-based revalidation to root page pages (i.e. not nested in navigation)
  - Removes time-based revalidation from non-center-scoped landing page (on-demand should handle this since there is no deep nesting). 
- Refactors validateMaxPosts to be a client side validate function due to https://github.com/payloadcms/payload/issues/14156
- Adds blocks revalidation to the top-level `richText` type field on the home pages collection (highlightedContent)
- Updates revalidation documentation

Leaving #455 open for the future.


src/app/(frontend)/[center]/theme-preview/page.tsx
Comment (busbyk) [inline src/app/(frontend)/[center]/theme-preview/page.tsx:1]: I wonder if we should remove this route now?

busbyk (commented):
I was wanting this issue to include a decision doc on why we chose these options (i.e. the use of `dynamic = 'force-static'` and `revalidate = 600` and an explanation of whether or not these options result in an SWR (stale while revalidate) behavior after the revalidation timeout period (of 600 seconds) or if there's some other behavior at play here. 

I also wonder if we could use longer timeouts for some pages like the homePages.

busbyk:
@rchlfryn I can't request a review because this is your PR but could you please review? Thanks!

src/app/(frontend)/[center]/theme-preview/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/theme-preview/page.tsx:1]: I think we should

src/blocks/BlogList/hooks/validateMaxPosts.ts
```diff
@@ -1,15 +1,15 @@
-import { FieldHook } from 'payload'
+import { NumberFieldValidation } from 'payload'
 
-export const validateMaxPosts: FieldHook = ({ value }) => {
+export const validateMaxPosts: NumberFieldValidation = async (value) => {
```
rchlfryn: I don't see an `await`.. do we need this `async`?

docs/decisions/011-incremental-static-regeneration.md
```diff
@@ -0,0 +1,64 @@
...
+  2. Update 12 field definitions
+  3. Extend getBlocksFromConfig.ts to discover all richText blocks
+  4. Rewrite findDocumentsWithBlockReferences.ts query logic significantly
+  5. Handle draft status checking differently
+  6. Extensive testing across all affected collections/blocks
+
+## Decision
+
+- [x] Add special handling for home page richText field (highlightedContent)
+- [x] Document how to handle specific richText fields like posts and home page
```
rchlfryn: Very ⛏️ ... 
```suggestion
- [x] Add special handling for home page `richText` field (highlightedContent)
- [x] Document how to handle specific `richText` fields like posts and home page
```

docs/decisions/011-incremental-static-regeneration.md
```diff
@@ -0,0 +1,64 @@
...
+
+A few notable behaviors:
+- You can use on-demand revalidation using `revalidatePath`, `revalidateTag` and time-based revalidation using `export const revalidate = {seconds}`. These can and should be used together.
+- Time-based revalidation uses the SWR (stale-while-revalidate) pattern. The next request after the timeout will serve the cached, stale/outdated page and generate a fresh server side rendered version of the page in the background and store this in the cache. The next request will serve the cached, fresh page.
+- `revalidateTag` will only revalidate the data cache layer. If a page was statically generated based on data from the data cache that is then invalidated by `revalidateTag`, that page will be regenerated on the next visit. The granularity of this means that other data fetched by a given page might still use the cached version of it and using `revalidatePath` is easier to reason about from a high level.
```
rchlfryn, line 14: Thank you for answering my question

src/app/(frontend)/page.tsx
```diff
@@ -8,7 +8,6 @@ import configPromise from '@payload-config'
 import { getPayload } from 'payload'
 
 export const dynamic = 'force-static'
-export const revalidate = 600
```
rchlfryn, line 11: In revalidation.md you mentioned
> - **Home pages**: 900 seconds (15 minutes) - Longer interval after implementing `highlightedContent` tracking

Do we want to add this back in at 900?

rchlfryn (commented):
I learned a lot about revalidation. I am surprised payload allowed nested blocking and didn't try to solve this.

A few comments, LG after updates.

src/app/(frontend)/[center]/theme-preview/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/theme-preview/page.tsx:1]: Hid it in production in https://github.com/NWACus/web/pull/593

src/blocks/BlogList/hooks/validateMaxPosts.ts
busbyk: Good catch - nope just a leftover from when this was a FieldHook and needed to be async. Will remove.

src/app/(frontend)/page.tsx
busbyk: Ah I forgot to update the docs after reverting that change. I had set it to 900 initially but changed it back to 600 because it seemed silly to have home pages have only a slightly longer revalidation period when highlightedContent still has the same problem where there could you could have deeply nested blocks.

But highlightedContent only has: 
```
ButtonBlock,
DocumentBlock,
GenericEmbedLexical,
MediaBlockLexical,
SingleBlogPostBlockLexical,
SponsorsBlock,
```

and after looking again, none of those blocks can have deeply nested blocks. So we could actually set home pages' revalidation time much much higher. 

Maybe a few hours to a day?

rchlfryn: I think that is fine, as long as it wont prevent someone from making a quick update on the homepage

busbyk: Ope, I take it back. The layout `blocks` field has the deeply nested problem. Going to stick with 600 seconds for home pages too. I'll update the docs.

PR #575, opened 2025-10-04
Update sponsor block - remove banner option

### LINKED ISSUE #568: Sponsor Banner renders a massive logo
**Describe the bug**
When the sponsor block is using banner mode, the logo is huge

**To Reproduce**
Add a sponsor banner

**Expected behavior**
I would expect it to resize appropriately like it does with the other sponsor block types

**Screenshots**

Image

Image

**Environment details**

- Desktop Chrome

**Existential musings**
too early

rchlfryn:
### Description
Remove banner option and automatically select instead if there is a image with the correct ratio. Not sure if we want to handle this automagically so I am totally fine altering or closing this.

### Issue
Resolves #568


busbyk (commented):
I'm a little skeptical of doing this automagically. Mostly wondering how users would know to upload the correct ratio. 

I think it would be good to keep this option and add a description to that `sponsorsLayout` field that explains the best ratio to use for a banner layout.

rustynwac:
Yes can we add some built in guidance for ratio? I'll also get it into Notion.

busbyk (approved):
I like it!

busbyk:
> Yes can we add some built in guidance for ratio? I'll also get it into Notion.

@rustynwac looks like ideal size is 1280x200 from Rachel's change here.

PR #584, opened 2025-10-09
Migration safety

busbyk:
## Description 

We had a migration result in data loss in production where `_rels` tables with foreign keys to the biographies table (i.e. `posts_rels` and `teams_rels`) had rows deleted because they have CASCADE DELETE and Payload/Drizzle uses a duplicate, drop, rename strategy for renaming or if a relationship is removed from it. 

Two takeaways from this after investigating: 
1. Payload/Drizzle does not warn you if a migration will result in data loss
2. Turso (libSQL) ignores PRAGMA foreign_keys=OFF/ON inside of transactions which is what caused this particular data loss. These types of migrations generated by Payload/Drizzle depend on this to avoid cascade deletes on foreign key relations. 

## Key changes

- Adds a `/src/scripts/check-migrations.ts` script that looks for 1) this particular known issue scenario and 2) other statements that would result in lost data (either intentionally or otherwise)
	- Runs this during pre-commit hook -- can be skipped if some statements are acceptable (like DROP TABLE if truly removing a table for example)
	- Runs this during CI -- leaves a comment on the PR highlighting the need for extra attention reviewing the migration(s)
- Adds docs
	- Documents how to run migrations locally + makes a few modifications to the code to enable this
	- Documents suggested approach to creating migrations and how to skip the check if necessary
- Modifies previously run migrations - It would appear that our init migration may have been slightly out of sync. Running all of our migrations locally surfaced a few errors (see modified migration files in this PR) which needed fixing in order to run them all locally. This shouldn't affect prod since these migrations have already been marked as run. I can't see an issue with this but please let me know if anyone thinks this might cause problems. 

## Possible enhancements

- Potentially a script to assist with diffing pre and post migrations locally
- Potentially a script / guidance on how to identify the `_rels` tables associated with a given table in a migration

## Questions

- Should the pre-commit hook actually fail if patterns are identified or just warn? Or only fail on the known issue but warn on other potentially destructive patterns?
- I'm thinking that these checks should be sufficient for now rather than diffing pre and post migration but would love opinions on that.


src/scripts/standalone-seed.ts
```diff
@@ -5,4 +5,4 @@ import { seed } from '../endpoints/seed'
 const payload = await getPayload({ config: configPromise })
 const payloadReq = await createLocalReq({}, payload)
 
-seed({ payload, req: payloadReq, incremental: false })
+await seed({ payload, req: payloadReq, incremental: false })
```
rchlfryn, line 8: ⛏️ This is in `main` from [#571 ](https://github.com/NWACus/web/pull/571/files#diff-ca970e298417e124a0a6cd435bf4192e84284997ea7cb6c3d6130cb6166b5954R8)

.gitignore
```diff
@@ -16,6 +16,7 @@ public/sitemap*.xml
 dev.db
 dev.db-shm
 dev.db-wal
+*.db
```
rchlfryn, line 19: Why was this added? maybe consolidate it with line 16

src/payload.config.ts
```diff
@@ -147,6 +147,9 @@ export default buildConfig({
       url: process.env.DATABASE_URI || '',
       authToken: process.env.DATABASE_AUTH_TOKEN || '',
     },
+    // Normal dev: push=true for quick schema iteration
+    // Migration testing: push=false to force migrations to run
+    push: process.env.ENABLE_LOCAL_MIGRATIONS !== 'true',
```
rchlfryn: This logic is breaking my brain..  
- If I was migrations, I set `ENABLE_LOCAL_MIGRATIONS` to `true` making `push: true` which would not run migrations locally? help
- also will we run into javacript typing here? string vs boolean

docs/migration-safety.md
```diff
@@ -0,0 +1,127 @@
...
+- Constraint drops
+
+## Workflow
+
+### Creating a new migration
+
+1. Make schema changes in your Payload config
+2. Run `pnpm payload migrate:create`
+3. Review generated migration file
+4. **Optional:** Manually run `pnpm migrate:check <filename>` to preview warnings
```
rchlfryn, line 54: Did you know the `<filename>` is optional here? learned that from the docs today...

rchlfryn (commented):
Great documentation as always! very helpful
Definitely not my area of expertise so feel free to wait for @stevekuznetsov or @yuliadub to take a look but left a few comments..

src/payload.config.ts
yuliadub: all env variables are read in as strings I'm pretty sure 🤔 

why 'push'? I think the naming is also screwing with my brain haha

docs/migration-safety.md
```diff
@@ -0,0 +1,127 @@
...
+## Background
+
+Payload's `migrate:create` command generates migrations based on schema changes, but doesn't warn about potentially destructive commands.
+
+### Known issue: foreign key cascade delete behavior differences
+There is also a difference in behavior between Turso's libSQL servers and our local SQLite databases: Turso does not respect `PRAGMA FOREIGN_KEYS=off` statements inside of transactions. Payload/Drizzle uses this strategy to temporarily disable cascade deletes of foreign key relationships. Because this doesn't work on Turso servers (a libSQL behavior), this can result in data losss in `_rels` tables.
+
+Background reading:
+- [SQLite docs highlighting foreign_keys behavior in transactions](https://www.sqlite.org/foreignkeys.html#fk_enable:~:text=It%20is%20not%20possible%20to%20enable%20or%20disable%20foreign%20key%20constraints%20in%20the%20middle%20of%20a%20multi%2Dstatement%20transaction)
+- https://github.com/tursodatabase/libsql-client-ts/issues/173
```
yuliadub, line 14: wow wow wow 😨

docs/migration-safety.md
```diff
@@ -0,0 +1,127 @@
...
+## Fixing migrations with PRAGMA foreign_keys=OFF
+
+When Payload/Drizzle generates migrations that recreate tables (the `__new_*` pattern), they use `PRAGMA foreign_keys=OFF` to prevent cascade deletes. However, **Turso does not respect this pragma inside transactions**, which means relationship tables (like `*_rels`) will still have their data cascade-deleted when the original table is dropped.
+
+### The fix: keep old attributes + mark as disabled/hidden and make a note in the code
+
+Unfortunately, several solutions we tried still result in data loss due to the foreign key behavior in libSQL. We tried:
+- Using `defer_foreign_keys` instead of `foreign_keys` to temporarily disable the cascade delete behavior
+- Using backup tables (same foreign key cascade delete issue)
+- Temporarily disabling the cascade delete behavior (complex and still resulted in data loss when using a backup table strategy)
```
yuliadub, line 104: one more "wow wow wow 😨 " thanks for digging into this, this is kind of blowing my mind

yuliadub:
> * Should the pre-commit hook actually fail if patterns are identified or just warn? Or only fail on the known issue but warn on other potentially destructive patterns?

Warning with pre-commit hook seems slightly counterproductive - you see the warning but you already made the commit. 🤷 However, personally I think I like the flow for this to go in and be flagged in a PR/have a record that this was an issue and a change was made that others can track vs it being a bit more silent when someone deals with an issue like that and fails to mention it later. ( or maybe I just know too many people who skip pre-commit hooks on purpose 😄 ) Anyways as long as the github action on the PR will block merge if it sees destructive patterns, that's good enough for me. But if we want to block with pre-commit hook I'd fine with that too, as long as something blocks it. 

> * I'm thinking that these checks should be sufficient for now rather than diffing pre and post migration but would love opinions on that.

Below is very much me thinking out loud, but where I arrived at is -- I think it would be good to have a migration diff that runs on main before we propagate changes to release. 

Diffing pre and post migrations would certainly be full proof -- are there other weird Turso/Payload gotchas that we dont know about yet? Having that be automated sounds nice, also just to make sure no steps were missed. You instructions outline the process well, thanks for those but they do have 10 steps and room for error depending on the user. 
Running a diff like that in a gah would take time, which would be annoying to do on a PR with every push but maybe could we only run it on merge to main? Or maybe it can even be a manual GAH that someone can run on their branch when they need an extra check on migrations? 

Side question: could we ever end up in a situation if destructive patterns are detected but say we are okay with them and want to proceed?

> * Modifies previously run migrations - It would appear that our init migration may have been slightly out of sync. Running all of our migrations locally surfaced a few errors (see modified migration files in this PR) which needed fixing in order to run them all locally. This shouldn't affect prod since these migrations have already been marked as run. I can't see an issue with this but please let me know if anyone thinks this might cause problems.

Since those migrations did run as is (as far as we know, right?) and you only see these issues now I'd say maybe put them in a different folder then migrations or record these changes in a readme somewhere vs changing files for migrations that already ran. I know they wont run again, but something about changing them after they already ran is not sitting well with me.

stevekuznetsov:
A pattern I find really useful is to capture the known bugaboos in a file - run the tool on all migrations, always, and collect the instances of data loss you find. Commit those to a file:

```yaml
failures:
- migration: <path to file>
  error:
    type: <which>
    <whatever other axes you need to identify this individually - line number, etc >
  acknowledgement: <left empty by the tool, filled in by a human to ack >
```

Then, failure of your tool will only happen when _new_ failure modes are added above and beyond the existing corpus. The "further review required" is surfaced at the unit-test level and further review is encoded by humans editing in the ACK. Your GitHub comment tool can detect new entries in this file in the PR and continue posting the comment on the diff.

stevekuznetsov:
> Running a diff like that ...

@yuliadub I think the challenging part is - migrations by definition must make the database state change, that's what they're for - so you'd need intelligent diffing logic to figure out if the contents of the diff constitute data loss.

src/scripts/check-migrations.ts
```diff
@@ -0,0 +1,202 @@
...
+import path from 'path'
+
+const DANGEROUS_PATTERNS = [
+  {
+    pattern: /DROP TABLE/i,
+    message: 'DROP TABLE detected - may cause cascade deletes in relationship tables',
+  },
+  {
+    pattern: /PRAGMA foreign_keys\s*=\s*OFF/i,
+    message: 'Foreign keys disabled - cascade behavior may be unpredictable',
```
stevekuznetsov: What happens the next time Drizzle generates such a migration? What was the intent of the previous one, and can we avoid making that type of change in the future? If not, what are our other options? Is it alright to hand-write a migration instead?

src/scripts/check-migrations.ts
```diff
@@ -0,0 +1,202 @@
...
+    pattern: /ALTER TABLE.*RENAME TO/i,
+    message: 'Table rename detected - check if relationship tables reference this',
+  },
+  {
+    pattern: /DROP COLUMN/i,
+    message: 'DROP COLUMN detected - permanent data loss for this column',
+  },
+  {
+    pattern: /DELETE FROM(?!.*WHERE)/i,
+    message: 'DELETE without WHERE clause - will delete all rows in table',
```
stevekuznetsov: This seems a bit fragile. What is the downside of flagging any `DROP` `DELETE` etc, indiscriminately?

stevekuznetsov (commented):
To give this approach the best chance of working, I think it should indiscriminately flag large classes of statements - even if the flags are false positives - and require manual annotations for them all to ensure the robots are keeping us accountable.

docs/migration-safety.md
```diff
@@ -0,0 +1,127 @@
...
+git commit --no-verify
+```
+
+**⚠️ Use with caution & make sure other checks have passed**
+
+## Fixing migrations with PRAGMA foreign_keys=OFF
+
+When Payload/Drizzle generates migrations that recreate tables (the `__new_*` pattern), they use `PRAGMA foreign_keys=OFF` to prevent cascade deletes. However, **Turso does not respect this pragma inside transactions**, which means relationship tables (like `*_rels`) will still have their data cascade-deleted when the original table is dropped.
+
+### The fix: keep old attributes + mark as disabled/hidden and make a note in the code
```
stevekuznetsov, line 99: My read is that the PRAGMA is not honored inside the transaction, right? The other semi-unsafe-but-potentially-ok thing since we have client requests that happen O(once daily) is to just use the PRAGMA globally, outside the transaction. Or, do that on a copy of the DB to make sure if someone else happens to be touching it at that exact moment, you can catch that.

src/scripts/standalone-seed.ts
busbyk: Merged main ✅

.gitignore
busbyk: I was messing around with creating temporary sqlite dbs during this PR and figured I would just add this preemptively.

src/payload.config.ts
busbyk: `push` is the name of the attribute on the `sqliteAdapter` and is also the name of the mode in Drizzle. Locally, Drizzle which is the ORM Payload uses under the hood uses "push mode" to keep the local sqlite database in sync with the payload schema without needing to run migrations. 

So if you want to be able to run migrations locally, you need to turn push mode off and run your migrations. Running migrations gets the db in sync with our desired state and marks each migration as run (if run obv) in the db. 

Yes `process.env.ENABLE_LOCAL_MIGRATIONS !== 'true'` reads horribly but environment variables are always read in as strings so you have to compare to the string value not the boolean value. Locally, I want push mode to be true unless ENABLE_LOCAL_MIGRATIONS is set to `'true'`. 

And actually as I'm writing this I realize that I want to set this to undefined if we're in production so I'll add that real quick. 

But does that make sense?

docs/migration-safety.md
busbyk: hahaha yes, exactly @yuliadub

docs/migration-safety.md
busbyk: Yea I made <filename> optional for the `migrate:check` script (i.e. src/scripts/check-migrations.ts). And yea you can also omit for `payload migrate:create` but I like giving them a name so you can know what they are at a glance.

docs/migration-safety.md
busbyk: Yea it was quite the journey

docs/migration-safety.md
busbyk: The PRAGMA is fine I believe - it's just that SQLite (or libSQL actually) ignores the setting of `foreign_keys` inside of transactions. 

Payload's migration system runs `up` and `down` inside of transactions. Where would we run something (i.e. a migration) globally outside of the migration system but still keep track of it. Set up our own simple migrations table? 😬

src/scripts/check-migrations.ts
busbyk: Yea we could look for less specific patterns.

src/scripts/check-migrations.ts
busbyk: > What happens the next time Drizzle generates such a migration?

My suggestion here would be to not commit that migration and to keep the field that you removed that resulting in this type of migration that we know will fail. Mark that field as hidden in the admin panel, leave a code comment for it marking it as deprecated and move on. 

> What was the intent of the previous one, and can we avoid making that type of change in the future?

The migration that caused the prod data loss was `20251001_221006_update_bios`. The intent of that migration was to remove the relationship between `Users` and `Biographies`. In this scenario we would simply mark the user field as `admin.hidden = true` and call it a day. 

> If not, what are our other options?

If that's not a viable option, I think it would be worth trying to write out own migration using Typescript which stores relationship values in memory and then recreates them after the tables have been updated. 

> Is it alright to hand-write a migration instead?

Yea definitely.

busbyk:
> > Running a diff like that ...
> 
> @yuliadub I think the challenging part is - migrations by definition must make the database state change, that's what they're for - so you'd need intelligent diffing logic to figure out if the contents of the diff constitute data loss.

Yea it definitely feels possible to diff them and output a relatively human-readable report. But I don't think that would be a trivial amount of work (hard to know) and that feels hard to justify given the amount of other issues we have in milestones with deadlines coming up.

docs/migration-safety.md
stevekuznetsov: ... ah. Our hands are bound. Ignore me.

src/scripts/check-migrations.ts
stevekuznetsov: If you're up for it, let's. It will help keep us honest.

yuliadub:
> > > Running a diff like that ...
> > 
> > 
> > @yuliadub I think the challenging part is - migrations by definition must make the database state change, that's what they're for - so you'd need intelligent diffing logic to figure out if the contents of the diff constitute data loss.
> 
> Yea it definitely feels possible to diff them and output a relatively human-readable report. But I don't think that would be a trivial amount of work (hard to know) and that feels hard to justify given the amount of other issues we have in milestones with deadlines coming up.

I totally get that - I think I was answering more of a long term if we had time and could invest time for this, i'd say it'd be worth it. But I agree having what you have put together is 1000 times better then having nothing at all, so it going to put us in a much better spot now.

yuliadub:
> > Running a diff like that ...
> 
> @yuliadub I think the challenging part is - migrations by definition must make the database state change, that's what they're for - so you'd need intelligent diffing logic to figure out if the contents of the diff constitute data loss.

@stevekuznetsov totally , yet systems around us continue to make it harder to just trust migrations 😢 😭

src/scripts/check-migrations.ts
yuliadub: thats a good call out, I agree with Steve!

src/payload.config.ts
yuliadub: with this context, the name makes a whole lot more sense! Maybe a quick comment above the env variable with this context and that is all the info that is needed!

busbyk: Sounds good, I have comments explaining that env var in `.env.example`. Let me know if that needs any tweaking.

busbyk:
OK I'm documenting future enhancements here: https://github.com/NWACus/web/issues/610

src/scripts/check-migrations.ts
busbyk: Sounds good - updated here: https://github.com/NWACus/web/pull/584/commits/874a72ba3317d34419c4220b48a983e12108adfd 

How's that look?

rchlfryn (approved):

Issue #398, opened 2025-08-14
Prefer using payload.logger over console for all logging


busbyk:
We don't have a pattern for when to use console.log vs. when to use payload.logger (a pino instance). 

There are only a few cases where we _can't_ use payload.logger (when payload hasn't been initialized yet for example). So we could decide to always use payload.logger after payload has been initialized. 

It would be really nice to always use a pino logger but a different instance than the payload.logger (with a different name) so we could easily differentiate between where the error occurred: in our code or in Payload code.

busbyk:
A few additional thoughts on loggers:
1. We can't use Payload's logger everywhere. Specifically: middleware and client components (you can't call getPayload in a client component)
2. Some places seem a little silly to use the Payload logger like `src/services/vercel.ts` but not the end of the world to just call `getPayload()` there since it's probably already been cached anyways. 
3. Do we really want to use the Payload logger in code that's not "payload code"? Wouldn't we actually want to use a different instance or instances of `pino` loggers with the same config as the one Payload uses but with different names? i.e. one for the middleware, one for client components -- however we want to break them up. 
4. Do we even want to log warnings/errors in the browser at all or should we use a wrapper (or maybe our logger could handle this for us) that warns/errors to the console locally but captures the exception in Sentry in production.

busbyk:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement (tech debt)
**Summary:** Establish one logging pattern — prefer a structured logger over `console.*` — with a dedicated app logger instance distinct from Payload's, and migrate existing `console.*` calls.

**Current behavior:**
There are ~60 `console.*` calls across `src/`, with no rule for when to use `console` vs `payload.logger` (a pino instance). `CLAUDE.md` already states a preference for `payload.logger`, but there's no shared logger for contexts where Payload isn't available, and no way to tell our log lines from Payload's.

**Desired behavior:**
- Introduce a **shared app logger**: a pino instance with a **distinct name** (e.g. `app`) from Payload's logger, usable where `payload.logger` isn't available (before Payload init, and non-Payload code). Where Payload is in scope, `payload.logger` remains fine.
- Migrate `console.*` calls in server/utility code to the appropriate logger. Log output should make it easy to tell our code from Payload's (different logger name).

**Key interfaces:**
- A new logger utility exporting the named pino instance.
- Replace `console.*` usages across server/util code.
[...]

PR #594, opened 2025-10-12
Safely handle missing relationships + coding guide

### LINKED ISSUE #528: Relationship references breaks build
We noticed our development environment was broken because it was referencing a form that got deleted. We want a graceful way to handle this for all relationship fields. 

We noticed a required field on a published document could be null if the related document got deleted.

This should be considered for the following collections:
- Pages
- Posts
- Media
- Documents
- Sponsors
- Tags
- Biographies
- Teams
- Forms

Steps to recreate:
1. Make a form
2. Reference form in block on page
3. Delete referenced form
4. Will break build with `null` error

See https://github.com/NWACus/web/actions/runs/17874383061/job/50833413741

busbyk:
## Description

This attempts to better handle missing relationships throughout the codebase and improves our type safety. The coding guide made sense to add as part of this. We should add to this as necessary. Please feel free to suggest changes to it here for things that jump to mind too. 

Basically, we want to return null for cases when the relationship is missing. We've realized this can happen when a related document is deleted (bypassing validation on that collection that requires it) or during live preview when validation has not already run. 

## Related Issues

Resolves #528 

## Key Changes

- Adds relationship validation utilities
- Implements usage of those utilities in components that currently gracefully handle missing relationships as well as components that did not before
- Adds a coding guide


docs/coding-guide.md
```diff
@@ -0,0 +1,119 @@
...
+
+This guide documents coding patterns and conventions for `nwacus/web` (AvyWeb). It's designed to help both team members and future open source contributors write consistent, safe code.
+
+This document should be updated periodically as we agree on additional patterns/practices.
+
+Review our decision docs to understand the context of important architectural decisions for this repo: `/docs/decisions`
+
+## Logging
+
+Prefer Payload's logger over console.log.
```
stevekuznetsov, line 11: I think there's an es-lint rule we can use to enforce this

docs/coding-guide.md
```diff
@@ -0,0 +1,119 @@
...
+Combine with additional filters:
+```typescript
+filterValidRelationships(sponsors).filter(s => !s.endDate || new Date(s.endDate) >= now)
+```
+
+## Error Handling
+
+### Graceful Degradation
+
+Components should degrade gracefully when data is missing:
```
stevekuznetsov, line 109: Could you add a quick blurb about what happens to the `null` and why it's safe to return it? (what part of the stack handles it, when do we need to add more handlers, etc)

src/blocks/BlogList/Component.tsx
```diff
@@ -16,23 +20,15 @@ export const BlogListBlockComponent = async (args: BlogListComponentProps) => {
   const { filterByTags, sortBy, queriedPosts } = args.dynamicOptions || {}
   const { staticPosts } = args.staticOptions || {}
 
-  let posts = staticPosts?.filter(
-    (post): post is Post =>
-      typeof post === 'object' && post !== null && post._status === 'published',
-  )
+  let posts: Post[] = filterValidPublishedRelationships(staticPosts)
```
stevekuznetsov, line 23: Naively, I think you want a monad for the valid check so you can compose the different predicates (published, whatever else) rather than having to invent a new top-level method for every useful combination.

stevekuznetsov: Or leave it for the future when we need more than one filter

docs/coding-guide.md
rchlfryn: https://eslint.org/docs/latest/rules/no-console

docs/coding-guide.md
```diff
@@ -0,0 +1,119 @@
...
+
+## TypeScript
+
+### Avoid Casting
+
+Only use TypeScript casting when absolutely necessary. Prefer using correct types and type guards:
+
+```typescript
+// ❌ - Casting masks potential runtime errors
+const sponsors = (props.sponsors as Sponsor[]).map(...)
```
rchlfryn, line 30: ❤️

docs/coding-guide.md
```diff
@@ -0,0 +1,119 @@
...
+Components should degrade gracefully when data is missing:
+
+```typescript
+// ✅ - Returns null instead of crashing
+if (!document || !isValidRelationship(document)) {
+  return null
+}
+```
+
+
```
rchlfryn, line 132:
```suggestion
## Styling
Use Tailwind classes over adding `.(s)css` files
```

rchlfryn: Do we also want to mention something around `Block` naming convention/folder structure? We would have to decide on something first. 😂  I think we have finally decided:

`SingleButtonBlock` - block config name
`singleButton` - block slug:
`SingleButtonComponent` - components
`SingleBlogPostBlockLexical` - config with `wrapInContainer` (also maybe explain `wrapInContainer`

Simple
```md
blocks
├── SingleButton
│   ├── Component.ts
    └── config.ts
```

Complex?
```md
blocks
├── SingleButton
│   ├── access
│   │   └── ...
│   ├── components
│   │   └── ...
│   └── hooks
│   │   └── ...
    └── config.ts
```

rchlfryn (changes requested):
I think the major check missing here are references to posts and pages through `BlogList (not dynamic)` `ButtonBlock`, `Link`, `LinkPreview`, `RelatedPosts` and/or `SingleBlogPost`. They seem to all use `CMSLink`.. so it might be an easy change.

docs/coding-guide.md
busbyk: I love this. Yea this makes sense to add.

src/blocks/BlogList/Component.tsx
busbyk: Yea I like the idea of chaining filter functions here if we end up having more than these two filter (i.e. isValid, isPublished) in the future.

docs/coding-guide.md
busbyk: There's actually a lot to this. 

We already have https://github.com/NWACus/web/issues/398

1. We can't use Payload's logger everywhere. Specifically: middleware and client components (you can't call getPayload in a client component)
2. Some places seem a little silly to use the Payload logger like `src/services/vercel.ts` but not the end of the world to just call `getPayload()` there since it's probably already been cached anyways. 
3. Do we really want to use the Payload logger in code that's not "payload code"? Wouldn't we actually want to use a different instance or instances of `pino` loggers with the same config as the one Payload uses but with different names? i.e. one for the middleware, one for client components -- however we want to break them up. 
4. Do we even want to log warnings/errors in the browser at all or should we use a wrapper (or maybe our logger could handle this for us) that warns/errors to the console locally but captures the exception in Sentry in production.

docs/coding-guide.md
busbyk: Yea for sure. Added.

docs/coding-guide.md
busbyk: Checkout https://github.com/NWACus/web/pull/594/commits/f41b5bd3dab3bb546ab4030070addd9c20b564ec and let me know if you want to change anything.

rchlfryn: LGTM!

busbyk:
> I think the major check missing here are references to posts and pages through `BlogList (not dynamic)` `ButtonBlock`, `Link`, `LinkPreview`, `RelatedPosts` and/or `SingleBlogPost`. They seem to all use `CMSLink`.. so it might be an easy change.

Good call on checking these. I think we're good in CMSLink. I did update SingleBlogPost to use the new `isValidPublishedRelationship` function but everything else is looking good.

docs/coding-guide.md
busbyk: I moved these notes over to #398. I'm going to hold off on making any changes to the use of console.log for now for that issue.

rchlfryn (approved):

Issue #599, opened 2025-10-13
Bug: hitting back button on widget pages


busbyk:
**Describe the bug**
The WidgetHashHandler uses `window.location.hash = {hash}` to set the desired starting-state hash for a given widget page. This adds to the browser's history stack which means that when a user clicks the browser back button / triggers the browser back functionality they will return to the url without the hash. This then triggers the hash to be added right back, landing them in the same place. 

**To Reproduce**

There are a few variations of this issue and I'm only testing on the observation widgets. 

One variation (hardest to reproduce for me but the scenario that Andy noticed):
1. Visit https://www.sierraavalanchecenter.org/
2. Click Observations -> Submit Observation in the navigation landing you at https://www.sierraavalanchecenter.org/observations/submit#/form
3. Click the browser back button 
4. Observe that you are still at https://www.sierraavalanchecenter.org/observations/submit#/form

Another variation where you wait a second or two between back button clicks:
1. Visit https://www.sierraavalanchecenter.org/
2. Click Observations -> Submit Observation in the navigation landing you at https://www.sierraavalanchecenter.org/observations/submit#/form
3. Click the browser back button, observe that you are now at https://www.sierraavalanchecenter.org/observations/submit#/view/observations (which shouldn't not be reachable - submit page route but hash for the recent observations widget state)
4. Click the browser back button again, observe that you are now at https://www.sierraavalanchecenter.org/

Another variation where you click back button with very little pause between clicks:
1. Visit https://www.sierraavalanchecenter.org/
2. Click Observations -> Submit Observation in the navigation landing you at https://www.sierraavalanchecenter.org/observations/submit#/form
3. Click the browser back button, observe that you are still at https://www.sierraavalanchecenter.org/observations/submit#/form
4. Click the browser back button again quickly, observe that you are now at https://www.sierraavalanchecenter.org/observations/submit#/view/observations 
5. Click the browser back button again quickly, observe that you are now at https://www.sierraavalanchecenter.org/#/form

**Expected behavior**
The user would be navigated back to https://www.sierraavalanchecenter.org/ in this scenario

**Assessment**

There are two issues at play here:
1. The WidgetHashHandler is adding to the browser history stack when it's updating the url hash to be the initial state hash for the widget associated with the given built-in page. We could use `replaceState` instead to avoid this. 
[...]

PR #609, opened 2025-10-14
Make observations NAC page dynamic

### LINKED ISSUE #567: Linking externally to filtered widget page not working properly
We were trying to solve for SAC wanting to use a filtered version of the Observations widget (like they do now) for accidents. The idea was that we would just use an external link in the nav to do so. Local Accidents > https://sac.avy-fx.org/observations#/view/avalanches?impacts=[%22Humans+Caught%22]

However, upon clicking on the navigation, I'm getting different behavior—none of which takes me to the filtered obs list

**To Reproduce**

1. Go to sac.avy-fx.org
2. Click on Accidents > Local Accidents

**Expected behavior**
I would expect to get to the filtered obs view

**Screenshots**
https://www.loom.com/share/3dc8167a93c84eb0a64b6d4558a1effa

**Environment details**

- Desktop Chrome

**Existential musings**
I just paid $6.50 for a croissant

rchlfryn:
## Description
Refactor observations page to allow params to manipulate the NAC widget

## Related Issues
Resolves #567 

## Key Changes
- Allows for static params
- Updates seed script to properly seed built-in pages & allow built-in page types references in navigation

## Screenshots / Demo
Accidents-Northwest-Avalanche-Center-10-14-2025_11_25_AM

Should we just make a dynamic NAC page for the observation widget?


docs/onboarding.md
```diff
@@ -19,12 +19,13 @@ We also need to get the new tenant into our Vercel Edge Config. Adding a new ten
...
-  | Weather Stations | /weather/stations/map| both |
-  | Recent Observations | /observations | both |
-  | Submit Observations | /observations/submit | both |
+  | All Forecasts | `/forecasts/avalanche`| mutli |
+  | _ZONE NAME_ | `/forecasts/avalanche/_ZONE_` | multi |
+  | Avalanche Forecast | `/forecasts/avalanche/_ZONE_` | single |
+  | Weather Stations | `/weather/stations/map`| both |
+  | Recent Observations | `/observations` | both |
+  | Submit Observations | `/observations/submit` | both |
+  | Local Accidents | `/accidents?impacts=["Humans Caught"]` | both |
```
busbyk: Per our conversation: remove this from the docs but we'll add this to the seed data as an example of adding a one-off built-in page for a center by request.

src/app/(frontend)/[center]/observations/page.tsx
```diff
@@ -11,8 +11,7 @@ import { getNACWidgetsConfig } from '@/utilities/getNACWidgetsConfig'
 import Link from 'next/link'
 import { notFound } from 'next/navigation'
 
-export const dynamic = 'force-static'
-export const revalidate = 600
+export const dynamic = 'force-dynamic'
```
Comment (busbyk) [inline src/app/(frontend)/[center]/observations/page.tsx:14]: Is this critical to the query params being interpreted? Because we still do want this page to be statically rendered if the query params will still be included in the url and are accessible to pass down to the NACWidget client component. The query params only effect the widget which is client side only. 

So hypothetically we could still statically generate this page but maybe you observed that wasn't the case?

src/app/(frontend)/[center]/observations/page.tsx
```diff
@@ -46,9 +47,26 @@ export default async function Page({ params }: Args) {
...
+        return
+      }
+      const decodedValue = decodeURIComponent(stringValue)
+      queryParams.set(key, decodedValue)
+    }
+  })
+
+  const queryString = queryParams.toString()
+  const initialHash = `/view/${tabView}${queryString ? `?${queryString}` : ''}`
+  const cleanUrl = queryString ? `/observations` : undefined
```
Comment (busbyk) [inline src/app/(frontend)/[center]/observations/page.tsx]: Per our conversation: It would be nice if we could move this all inside of NACWidget so that query params could affect other widgets too. Would just need to pass searchParams to NACWidget. 

And wondering if instantiating URLSearchParams with the searchParams would decode them for you to avoid needing to loop through them all: `new URLSearchParams(searchParams)`

src/components/NACWidget/WidgetHashHandler.client.tsx
```diff
@@ -2,19 +2,35 @@
...
+    if (hasInitialized.current) return
+
+    const existingHash = window.location.hash
+    const hasExistingHash = existingHash && existingHash !== '#'
+
+    if (!hasExistingHash) {
+      if (cleanUrl) {
+        window.history.replaceState({}, '', `${cleanUrl}#${initialHash}`)
+      } else {
+        window.location.hash = initialHash
```
busbyk: Feel free to put this off for another PR if you want to get this merged in but using `replaceState` here too might help with resolving https://github.com/NWACus/web/issues/599

Might only address part of the problem since the way the widgets handling their internal routing may be adding to the history stack too. So that part might be out of our hands but would be nice not to add to the history stack here.

src/endpoints/seed/index.ts
```diff
@@ -723,18 +724,23 @@ export const seed = async ({
...
     tenantsById,
-    () => 'builtInPages',
+    (obj) => obj.title,
     Object.values(tenants)
       .map((tenant): RequiredDataFromCollectionSlug<'builtInPages'>[] => [
         builtInPage(tenant, 'All Forecasts', '/forecasts/avalanche'),
         builtInPage(tenant, 'Weather Stations', '/weather/stations/map'),
         builtInPage(tenant, 'Recent Observations', '/observations'),
         builtInPage(tenant, 'Submit Observations', '/observations/submit'),
+        builtInPage(
```
busbyk: Since this is representing a one-off for a single tenant making a custom request, should we only add this for one tenant? Sierra?

busbyk (commented):
Comments from our discussion + a few others.

src/app/(frontend)/[center]/observations/page.tsx
```diff
@@ -46,9 +47,26 @@ export default async function Page({ params }: Args) {
 
   const { version, baseUrl } = await getNACWidgetsConfig()
 
+  // Build query params, preserving array values
+  const queryParams = new URLSearchParams()
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/observations/page.tsx]: Decided to keep this here because it would be more work to move to `WidgetHashHandler`. We would need to keep track of `initialHash` (which differ a lot) & `cleanUrl` which we could easily do with `pathname`.. but still not worth it.

src/app/(frontend)/[center]/observations/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/observations/page.tsx:14]: According to the [next.js docs](https://nextjs.org/docs/app/guides/incremental-static-regeneration) need to use `force-dynamic` to have access to `searchParams`.

src/app/(frontend)/[center]/observations/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/observations/page.tsx]: > And wondering if instantiating URLSearchParams with the searchParams would decode them for you to avoid needing to loop through them all: new URLSearchParams(searchParams)

I tried to leave comments in the code. We unfortunately need to loop through everything to respect arrays.
```js
// This WON'T work as expected:
new URLSearchParams({ tag: ['avalanche', 'snowpack'] })
// Result: ?tag=avalanche,snowpack  (comma-separated, NOT what you want!)

// This DOES work:
const params = new URLSearchParams()
params.append('tag', 'avalanche')
params.append('tag', 'snowpack')
// Result: ?tag=avalanche&tag=snowpack  (correct!)
```

We also unfortunately need to decode the params before passing them to the NAC widget because it is not able to handle encoded params

src/components/NACWidget/WidgetHashHandler.client.tsx
rchlfryn: I updated this line to use `window.history.replaceState`, but I don't think it completely fixes the error in #599 so I am going to leave it open

rchlfryn: Update: I think the changes in [ac7ef21](https://github.com/NWACus/web/pull/609/commits/ac7ef21be9dbd68c6a8eb78e8a3d553f5d0f38f5) and [89ae8c7](https://github.com/NWACus/web/pull/609/commits/89ae8c71d107d153230be2e6a649f2c7cf37fbbd) helped fixed this. It turns out window.location.hash and window.location.href work with browser back/forward buttons naturally. I made this change because the URL was updating but not reloading the page.

src/app/(frontend)/[center]/observations/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/observations/page.tsx]: Ignore this - I updated this to move `searchParams` handling to `WidgetHashHandler`

src/components/NACWidget/WidgetHashHandler.client.tsx
```diff
@@ -1,20 +1,60 @@
...
+      } else {
+        queryParams.set(key, decodeURIComponent(value))
+      }
+    })
+
+    const tabView = queryParams.get('tabView') || 'observations'
+    queryParams.delete('tabView')
+    const queryString = queryParams.toString()
+    initialHash = `/view/${tabView}${queryString ? `?${queryString}` : ''}`
+    redirectUrl = queryString ? pathname : undefined
```
busbyk, line 36: 1. This is specific to the observations widget so I don't think it belongs in this component since this is used across the widgets. If another widget page passed in searchParams I think it would override the initialHash to the observations widgets' initialHash right?
2. Using `tabView` to determine the initialHash to redirect to means that a user couldn't just copy and paste the url after selecting filters in the observations page. They would have to know to build up the url using the right `tabView` value too.

src/components/NACWidget/WidgetHashHandler.client.tsx
```diff
@@ -1,20 +1,60 @@
...
+  }
 
   useEffect(() => {
-    if (!hasSetInitialHash.current || !hash) {
-      window.location.hash = initialHash
-      hasSetInitialHash.current = true
+    if (hasInitialized.current) return
```
busbyk, line 40: This no longer handles the scenario where a user is on the observations widget page (on the same tab as initial `/observations#/view/observations`, or a different one `/observations#/view/avalanche`) and then they click "Recent Observations" in the nav or the breadcrumbs which navigates them to `/observations` without a page reload because it's using client side navigation. 

If they were on the same tab as initial, no big deal - the url just doesn't have the hash, whatever. But if they were on the avalanche tab, they would still be on the avalanche tab. Previously, this was checking to see if the hash was empty, regardless of initialization state, and setting to the initial if so.

busbyk (commented):
The core issue here is that the observations widget only interprets query parameters as filters on initial load. So we need to do a full reload of the widget (reloading the page works) to get it to re-interpret new query params as filters and refetch obs. Except they're not actually query params - they're a query string inside of the hash. 

I tried a different approach here: https://github.com/NWACus/web/blob/busbyk/nac-widget-query-params/src/components/NACWidget/WidgetHashHandler.client.tsx with comments explaining the different scenarios. But it uses some brute logic which reloads the page (by setting location.href) if the new hash has query params in it. Since the querystring appended to the hash persists when you change tabs in the widget (by clicking the tabs) the useEffect in WidgetHashHandler does a page refresh...not ideal when moving from the avalanche tab to the observations tab. 

The better solution here would be to update the observations widget itself to watch for changes to the querystring appended to the hash and update it's filters on change, not just on initial load.

rchlfryn:
Closing this for now - We need to make an issue on the NAC widget side to accept parameters

PR #613, opened 2025-10-15
Misc PR: enable prose, add button block to posts & slug validation

### LINKED ISSUE #586: Disallow slashes and empty strings in slugs
**Describe the bug**
Currently, a user can include a slash in a slug like `attempted-grouping/actual-page-or-post-slug`. This causes issues with routing and we shouldn't allow this. 

A user can also save a document with a slug with only whitespace which we should not allow. 

**To Reproduce**
Add a slug to a page or post with a `/`.

**Expected behavior**
Only valid slug characters should be allowed. Probably only `-`. 

Let's make sure there aren't any problematic slug values in production before merging this. 

**Existential musings**
Don't forget to tip your weather forecaster

### LINKED ISSUE #592: Add button to Post editor
Right now you cant add a button in Posts

Image

### LINKED ISSUE #605: Image Text Lists do not respect formatting
**Describe the bug**
The text editor allows you to use formating for Image Text Lists, but the formating is not displayed on the preview or the published page.  

**To Reproduce**
Add formatting to the text on an Image Text List.

**Expected behavior**
The page output should match the WYSIWYG editor.

**Screenshots**

Image

**Environment details**

Desktop Chrome | Android Chrome

rchlfryn:
## Description
- Enable prose on `ImageTextLists` block
- Add `ButtonBlock` to posts editor 
- Disallow slashes and empty strings in slugs for posts and pages

## Related Issues
Fixes #605
Fixes #592
Fixes  #586

### DIFF
--- a/src/blocks/ImageTextList/Component.tsx
+++ b/src/blocks/ImageTextList/Component.tsx
@@ -59,7 +59,7 @@ export const ImageTextList = (props: Props) => {
                   <div className={`${isSideLayout ? 'col-span-3' : 'mt-4'}`}>
                     <h3 className="text-lg font-bold">{title}</h3>
                     <div className="mt-2">
-                      <RichText data={richText} enableGutter={false} enableProse={false} />
+                      <RichText data={richText} enableGutter={false} />
                     </div>
                   </div>
                 </div>
--- a/src/utilities/validateSlug.ts
+++ b/src/utilities/validateSlug.ts
@@ -7,7 +7,8 @@ export const validateSlug: TextFieldSingleValidation = (value) => {
   if (value?.includes('/')) {
     return 'Slug cannot contain /'
   }
-  const slugPattern = /^[a-z0-9]+(?:-+[a-z0-9]+)*$/
+  // checks for kebab-case strings: allows letters, numbers and -
+  const slugPattern = /^[a-zA-Z0-9]+(?:-+[a-zA-Z0-9]+)*$/
 
-  return slugPattern.test(value) || 'Invalid slug'
+  return slugPattern.test(value) || 'Invalid slug: must be letters, numbers, or -'
 }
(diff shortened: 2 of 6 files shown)


stevekuznetsov:
> Removes # check from redirects regex on from URL (need to allow Sierra to redirect old forecast page)

Does this work? URI fragments don't get sent to the server, so I am not clear on how the middleware could use them for redirects

rchlfryn:
> Does this work? URI fragments don't get sent to the server, so I am not clear on how the middleware could use them for redirects

@stevekuznetsov It does not. I removed this and Andy closed the issue.

src/blocks/ImageTextList/Component.tsx
```diff
@@ -59,7 +59,7 @@ export const ImageTextList = (props: Props) => {
                   <div className={`${isSideLayout ? 'col-span-3' : 'mt-4'}`}>
                     <h3 className="text-lg font-bold">{title}</h3>
                     <div className="mt-2">
-                      <RichText data={richText} enableGutter={false} enableProse={false} />
```
busbyk, line 62: heh, funny I didn't even know we had that prop on RichText. Do we even want that as an option?

src/utilities/validateSlug.ts
```diff
@@ -0,0 +1,13 @@
+import { TextFieldSingleValidation } from 'payload'
+
+export const validateSlug: TextFieldSingleValidation = (value) => {
+  if (value === null || value === undefined) {
+    return 'Slug must not be blank'
+  }
+  if (value?.includes('/')) {
+    return 'Slug cannot contain /'
+  }
+  const slugPattern = /^[a-z0-9]+(?:-+[a-z0-9]+)*$/
```
busbyk: Can you add a comment explaining this regex?

busbyk (changes requested):

src/blocks/ImageTextList/Component.tsx
rchlfryn: It came out of the box with payload. I think we can remove it

busbyk: Yea checks out. I think that's a good idea.

busbyk (approved):
🤘

PR #621, opened 2025-10-20
Mobile tweaks

busbyk:
## Description

I was testing out sierraavalanchecenter.org on mobile and noticed a few things I wanted to fix. 

There is extra vertical space between content columns in many places (on the home page for example). I thought this would be due to incorrect `gap-{n}` class names but it's actually just an extra `<p>` after the embedded block. This might be intentional or it might just be a common accident because Payload adds an extra paragraph after embedding a block feature. We may want to communicate this to users. 

## Key Changes

Each change has it's own commit for reference. 

#### 1 - Removed starting/ending margin on ImageQuote block components on mobile

Before: 
CleanShot 2025-10-19 at 18 33 50
After:
CleanShot 2025-10-19 at 18 34 10

#### 2 - Matched vertical gap in content columns to other block components
Really minor but figured we might as well be consistent. Most block components use gap-6.

#### 3 - Centered content in footer on mobile

Before:
CleanShot 2025-10-19 at 19 15 16
After: 
CleanShot 2025-10-19 at 19 16 05

Note that the embed still appears not centered because of the styles in the iframe itself. That iframe would need to be modified in order for it to be centered. 

#### 4 - Ensuring breadcrumbs never wrap on mobile 

Before:
CleanShot 2025-10-19 at 19 22 36
After: 
CleanShot 2025-10-19 at 19 22 54

#### 6 - Stack submit button vertically on observations page on mobile

Before:
CleanShot 2025-10-19 at 19 41 48
After: 
CleanShot 2025-10-19 at 19 41 26

### DIFF
--- a/src/blocks/ImageQuote/Component.tsx
+++ b/src/blocks/ImageQuote/Component.tsx
@@ -17,9 +17,9 @@ export const ImageQuote = (props: Props) => {
   return (
     <div className={`${bgColorClass}`}>
       <div className="container py-10">
-        <div className="grid md:grid-cols-12 gap-x-6 gap-y-6">
+        <div className="grid md:grid-cols-12 gap-6">
           <div
-            className={`items-center md:col-span-4 self-start ${imageLayout === 'right' ? 'order-last ms-6' : 'me-6 '}`}
+            className={`items-center md:col-span-4 self-start ${imageLayout === 'right' ? 'order-last md:ms-6' : 'md:me-6'}`}
           >
             {image && <ImageMedia imgClassName={cn(imgClassName)} resource={image} />}
           </div>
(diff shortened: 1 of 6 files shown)


src/blocks/ImageQuote/Component.tsx
```diff
@@ -17,9 +17,9 @@ export const ImageQuote = (props: Props) => {
   return (
     <div className={`${bgColorClass}`}>
       <div className="container py-10">
-        <div className="grid md:grid-cols-12 gap-x-6 gap-y-6">
+        <div className="grid md:grid-cols-12 gap-6">
           <div
-            className={`items-center md:col-span-4 self-start ${imageLayout === 'right' ? 'order-last ms-6' : 'me-6 '}`}
+            className={`items-center md:col-span-4 self-start ${imageLayout === 'right' ? 'order-last md:ms-6' : 'md:me-6'}`}
```
rchlfryn, line 22: should this be `sm` or `md`?

rchlfryn (approved):
LG - one minor comment but merge when you're ready

src/blocks/ImageQuote/Component.tsx
busbyk: I went for `md` here because that's when these elements change their layout. I think `md` makes sense for this component.

Issue #622, opened 2025-10-20
Home page cycling for super admin in admin panel


rustynwac:
**Describe the bug**
When I click on the home page in the admin panel it just cycles through all the tenants:
https://www.loom.com/share/5e503b2886f3410f9e906fb73d8dba4b?sid=048b0d6e-07ac-4438-b84a-b38747d70cae

**Expected behavior**
I would expect to stay on the tenant I'm already in. 

**Environment details**

-Desktop Chrome

**Existential musings**
Monday really kicking my ass in a way I havent experienced in awhile!

rustynwac:
I logged in with my SAC user and it seems I can use the homepage just fine.

busbyk:
Fixed by #624

PR #624, opened 2025-10-20
Bugfix: global collection redirect loop

busbyk:
## Description

When visiting a global collection where some documents are published and some are drafts there would be an endless redirect loop due to the interaction between the `TenantFieldComponent` and the `GlobalViewRedirect` component. The `TenantFieldComponent` would receive the old state of the field after the redirect, triggering a `setTenant` call to align the current tenant with the document's tenant field. This would trigger another redirect by the `GlobalViewRedirect`. This would continue indefinitely. 

The solution is to respect the readOnly nature of the tenant field in global collections in the `TenantFieldComponent`. This is determined by inspecting the `unique` attribute on the tenant field (if true = global collection) and not updating the set tenant based on the tenant field value. A global collection document's tenant field should be readOnly. 

## Related Issues

#622 

## Key Changes

Respect the readOnly nature of the tenant field in a global collection document

### DIFF
--- a/src/fields/tenantField/TenantFieldComponent.tsx
+++ b/src/fields/tenantField/TenantFieldComponent.tsx
@@ -27,10 +27,11 @@ export const TenantFieldComponent = (args: Props) => {
     setTenant,
   } = useTenantSelection()
 
+  const isGlobalCollection = !!unique
   const hasSetValueRef = React.useRef(false)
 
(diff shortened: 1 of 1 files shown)


busbyk:
Vercel CLI error is probably due to the AWS outage. I'll try this again later.

rchlfryn (approved):

PR #628, opened 2025-10-23
Bug: corrects link for posts in handleReferenceURL

### LINKED ISSUE #627: Bug: Button block uses incorrect link for internal posts
**Describe the bug**
When using the button block, the link we generate for internal posts is incorrect.

**To Reproduce**

1. Use the button block to add reference an internal post
2. Preview the page and click on the link
3. you are navigated to `/posts/NAME-OF-POST`

**Expected behavior**
The button should navigate to `/blog/NAME-OF-POST`

We just need to add an `if` statement in `handleReferenceUR` to handle `posts`

**Existential musings**
The tactical feeling you get from turning on and off a light switch is a small spring being bent out of alignment

rchlfryn:
## Description
Add edge case for posts to handleReferenceURL

## Related Issues
Fixes #627

### DIFF
--- a/__tests__/client/handleURL.client.test.ts
+++ b/__tests__/client/handleURL.client.test.ts
@@ -47,7 +47,7 @@ describe('handleReferenceURL', () => {
         },
       })
 
-      expect(result).toBe('/posts/my-blog-post')
+      expect(result).toBe('/blog/my-blog-post')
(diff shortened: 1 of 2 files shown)


busbyk (approved):
🤘

PR #636, opened 2025-10-25
Events: groups and tags

### LINKED ISSUE #630: Add Event groups and Event Tags
Add Event series
- ie : this would represent an event series like multiple meet your forecasters

Add Event Tags 
- i.e. another thing to group events by (an attribute of an event like it being 21+ or Women Only/Women Taught or some other generic grouping like that)

rchlfryn:
## Description
Adds event groups and tags

## Related Issues
Fixes #630


src/collections/Events/index.ts
```diff
@@ -29,7 +29,7 @@ export const Events: CollectionConfig = {
   access: accessByTenantRoleWithPermissiveRead('events'),
   admin: {
     baseListFilter: filterByTenant,
-    group: 'Content',
+    group: 'Events',
```
busbyk, line 32: 👍

src/collections/EventSubTypes/index.ts
```diff
@@ -29,28 +30,7 @@ export const EventSubTypes: CollectionConfig = {
...
-          label: 'CRM Integration',
-          type: 'select',
-          options: [
-            { label: 'AC Salesforce', value: 'ac-salesforce' },
-            { label: 'A3 CRM', value: 'a3-crm' },
-          ],
-        },
-      ],
-    },
+    crmFields(),
```
busbyk, line 33: 🧡

src/endpoints/seed/index.ts
```diff
@@ -228,7 +230,14 @@ export const seed = async ({
...
-            collections: ['navigations', 'tenants', 'eventTypes', 'eventSubTypes'],
+            collections: [
+              'navigations',
+              'tenants',
+              'eventGroups',
+              'eventTags',
+              'eventTypes',
+              'eventSubTypes',
+            ],
             actions: ['read'],
```
busbyk, line 236: Just read? I think an admin should be able to manage groups and tags.

busbyk (approved):
Looks great!

src/endpoints/seed/index.ts
rchlfryn: I was just keeping what was there, but I will update permissions

busbyk: So my comment was actually to past me? 😂

Issue #608, opened 2025-10-14
Inconsistent Blank Screen of Death When Editing


andyanderso:
**Describe the bug**
Sometimes while editing a page, the screen goes blank, and I have to refresh the page to continue editing.

**To Reproduce**
I don't have a consistent way to reproduce this bug. It seems to happen most when I add a new layout to a page or when dragging layouts around in the editor. Sometimes it happens when I am just typing along. Unfortunately, I can not seem to find a pattern. 

**Expected behavior**
No blank screen during editing. 

**Screenshots**
This is a screen shot of the blank screen with this console error message from Chrome:   
`"POST https://avy-fx.org/ingest/e/?retry_count=1&ip=0&_=1760454773317&ver=1.257.0 net::ERR_BLOCKED_BY_CLIENT"`

<img width="1907" height="994" alt="Image" src="https://github.com/user-attachments/assets/160e31ec-2473-45b5-9e30-7c3da1fc3d09" />

This is a screenshot from Firefox with its console errors:

<img width="957" height="998" alt="Image" src="https://github.com/user-attachments/assets/614744ed-b9a9-4c00-8fca-2a13056fbe14" />

Note the blank screen where the editor is supposed to be. 

**Environment details**

Desktop - Pop!_OS 22.04 LTS - Ubuntu Linux variant. 
Chrome Version 139.0.7258.154 (Official Build) (64-bit)
Firefox 142.0.1 (64-bit)

andyanderso:
I can reproduce it consistently now:
1. Edit this page: https://avy-fx.org/admin/collections/pages/72
2. Try to remove the 2nd section (the empty Image Text List)

<img width="1009" height="285" alt="Image" src="https://github.com/user-attachments/assets/7924f9e4-b8d0-4f39-a806-7b640722a3ab" />
3. Watch the screen go blank :)

rustynwac:
This happened to me once. I moved the block up or down before removing then it worked. 

I was able to replicate on Chrome. I duplicated the page and got the same behavior. I was able to delete the block by moving it down then trying.

andyanderso:
I have been able to work around it in similar ways. I will leave this page as is for now so we can hopefully debug the issue. At this point, I have definitely had it happen on multiple occasions.

busbyk:
I'm so glad you were able to reproduce this in a reliable way! I ran into a similar bug the other day once and then couldn't reproduce it. I was able to reproduce this using your steps.

I'll investigate further and report back.

[...]

PR #638, opened 2025-10-29
Bugfix: all blocks with richText fields throw uncaught errors on initialization

busbyk:
## Description

Turns out that the `undefined is not valid JSON` error was actually happening on all blocks with a `richText` type field in production environments (`HeaderBlock`, `BlogListBlock`, `ImageText`, `LinkPreview`, `MediaBlock`). This wasn't happening locally and I still can only intermittently reproduce this locally for some reason. 

This PR introduces a patch to the `@payloadcms/richtext-lexical` package. This change should really happen in the `payload` monorepo and I do intend to submit a PR but this is causing a really terrible user experience in the admin panel right now. So this is a quick fix that we should remove later once the root cause has been addressed in Payload. 

The offending line is: https://github.com/payloadcms/payload/blob/059185f88f93b4690905d3911ee27b1b6010069e/packages/richtext-lexical/src/field/Field.tsx#L150. `prevValueRef.current` can be undefined, causing `JSON.parse` to throw. I assume this wasn't caught by TS bc `JSON.stringify` is typed to return a string but when `undefined` is passed, it returns `undefined`:

```
> JSON.stringify(undefined)
undefined
> JSON.parse(JSON.stringify(undefined))
Uncaught SyntaxError: "undefined" is not valid JSON
```

## Related Issues

Originally reported under #608 

## Key Changes

- Uses a [pnpm patch](https://pnpm.io/cli/patch) to make a change to `@payloadcms/packages/richtext-lexical/src/field/Field.tsx` that falls back to a null value if `prevValueRef.current` is `undefined`. 

I'll link a future PR here.


rchlfryn (approved):
Approved, but I am curious as to why this is all of a sudden happening. It looks like they changed this line in 5.58.0. We can either do this patch or revert the package back to 5.57.

busbyk:
> Approved, but I am curious as to why this is all of a sudden happening. It looks like they changed this line in 5.58.0. We can either do this patch or revert the package back to 5.57.

The weirdest thing is that I can't reproduce this using their test repos in the payload monorepo. It doesn't exactly match our setup of course so there must be something we're doing that isn't in their lexical test repo but that's another factor for why I haven't opened a PR on their end yet. 

I like being on latest if we can be so I think using this patch for now is fine as long as we address it in Payload soon.

PR #639, opened 2025-10-29
Upgrade payload to v3.61.1

busbyk:
## Description

Upgrades Payload to latest version

## Key Changes

- Upgrade payload
- Move dependabot configuration to the correct file location
- Updates `@payloadcms/richtext-lexical` patch to match changes from https://github.com/payloadcms/payload/pull/14385 which resolve the issue and will be in the next Payload release.

We'll just need to make sure we remove the patch when payload 3.61.2 is release.


rchlfryn (approved):
Discussed in slack but foreign keys renamed because of this: https://github.com/payloadcms/payload/pull/14341

PR #640, opened 2025-10-29
Events: change types and sub types

### LINKED ISSUE #637: Events - change types and subtypes
Make event types and subtypes constants rather than a collection

rchlfryn:
## Description
Changes EventTypes and EventSubTypes from collection to constants in a file.

## Related Issues
Fixes #637 

## Key Changes
- Changes EventTypes and EventSubTypes from collection to constants in a file.
- Updates autosave time back to default
- Removes incorrect use of `Suspense` on server components

## Migration Explanation
Yes - but we will add this later

**Do we want OR logic for the filters?** I think yes.
If I choose `Awareness` and `Pro 1` as my filter options, I should see events for both rather than a awareness class with a Pro subtype which will never happen. I implemented this logic in (326b9333920dff686989bfcbbf16ff2d5218b7ea) We can leave it as is or the correction to the UI is :
- selecting a subtype automatically selects its parents type
- selecting type automatically selects all subtypes
 
This is not implemented but let me know your thoughts on if it should be.

**Do we still need all the CRM info? if so how do we want to handle it?**


src/collections/Events/constants.ts
```diff
@@ -0,0 +1,135 @@
...
+  // Sub-types for Field Class by AC
+  {
+    label: 'Snowmobile Classes',
+    description:
+      'Avalanche safety and rescue training specifically designed for snowmobile users and backcountry riders.',
+    eventType: 'field-class-by-ac',
+    value: 'snowmobile-classes',
+    crmId: 'snowmobile_classes',
+    crmIntegration: 'ac-salesforce',
+  },
```
busbyk: Looks like Rusty removed this from the Events doc. I think we chatted about this during our meeting last week - it did look weird with only one subtype here. Let's remove for now?

src/collections/Events/constants.ts
```diff
@@ -0,0 +1,135 @@
...
+  value: string
+}
+export const eventTypesData: EventType[] = [
+  {
+    label: 'Events By AC',
+    description:
+      'General events organized and hosted by the Avalanche Center, including community events, meetings, and social gatherings.',
+    value: 'events-by-ac',
+    crmId: 'events_by_ac',
+    crmIntegration: 'ac-salesforce',
```
busbyk: The difference with having these as fields on the now-removed EventTypes and EventSubTypes collections and having them here is that we have to define values for them here rather than them just being placeholders. 

These values might be confusing in the future but these are also very reasonable guesses at valid values for the future. I'd be fine with leaving these or removing them. Maybe leaning towards removing them.

src/collections/Events/index.ts
```diff
@@ -281,30 +282,51 @@ export const Events: CollectionConfig = {
...
-      relationTo: 'eventTypes',
+      type: 'select',
       required: true,
       admin: {
         position: 'sidebar',
       },
+      hooks: {
+        beforeChange: [
+          ({ value, siblingData }) => {
+            if (value !== 'field-class-by-ac' && value !== 'course-by-external-provider') {
```
busbyk: Could we just group the subtypes with their type so we could just check to see if the type has sub types here and in the condition for sub type?

src/collections/Events/index.ts
busbyk, line 1: Since event type and event sub type are only ever displayed in the context of an event now (instead of being in their own collection) should we just name or label them as type and sub type?

src/collections/Events/index.ts
```diff
@@ -281,30 +282,51 @@ export const Events: CollectionConfig = {
...
+          },
+        ],
+      },
+      options: eventTypesData.map((eventType) => ({
+        label: eventType.label,
+        value: eventType.value,
+      })),
     },
     {
       name: 'eventSubType',
```
busbyk: I'm leaning towards thinking we should make this required if there's a type with available sub types. I can't think of a scenario where we would want to allow _just_ the type being course by external provider but without a sub type. What do you think?

src/collections/HomePages/index.tsx
```diff
@@ -173,22 +173,22 @@ export const HomePages: CollectionConfig = {
       blocks: [
         BiographyBlock,
         BlogListBlock,
+        SingleBlogPostBlock,
         Content,
         DocumentBlock,
+        EventListBlock,
+        SingleEventBlock,
         FormBlock,
+        GenericEmbed,
```
busbyk, line 182: What's the ordering logic here? I don't care too much tbh just curious haha.

src/components/EventPreview/index.tsx
```diff
@@ -82,10 +86,10 @@ export const EventPreview = (props: {
       )}
     >
       {startDate && (
-        <div className="hidden @md:flex flex-col gap-1 -mt-1 items-center text-center flex-shrink-0">
+        <div className="hidden @md:flex flex-col gap-1 -mt-1 items-center text-center flex-shrink-0 p-4">
```
busbyk, line 87: I like the horizontal padding but the `-mt-1` was there to align it with the top of the other content:

Image

I think either change this to just `px-4` or remove `-mt-1` if you prefer vertical padding too.

busbyk (commented):
Looks great. Just a few comments.

src/collections/HomePages/index.tsx
rchlfryn: I just reorder single blog to be by blog list, so Blog.
Same for Events.
And GenericEmbed is not in alphabetical order.

src/collections/Events/constants.ts
rchlfryn: Yep - forgot about that.

src/collections/HomePages/index.tsx
busbyk: Ah cool, makes sense

src/collections/Events/index.ts
rchlfryn: Thats a good idea, but unfortunately there is no easy way to populate the options for the `subType` config. Rather than having two sources of truth, I will leave it as a separate object. Any thoughts on how to make it work?
I did update this check to get the unique event types from the sub type object.

src/collections/Events/constants.ts
rchlfryn: Since we dont know for sure what the ids are and they are currently not being used, I will remove them for now.

src/components/EventPreview/index.tsx
rchlfryn: I kind of liked the misaligned date. I can update the style though.

src/collections/Events/index.ts
rchlfryn: I like it - added a check

src/components/EventPreview/index.tsx
busbyk: Hey if you think it looks better I'm open to it

busbyk (approved):
🤘🤘🤘

src/collections/Events/index.ts
busbyk: Another approach would be to use a `ui` type field with a custom component that renders inputs for the type/subtype in one file and utilizes the `useField` hooks to update the document's form state. You'd then mark `admin.hidden=true` for type and subType. 

But your approach works great!

Issue #633, opened 2025-10-25
Events POCs


rchlfryn:
- [x] POC for location data that can be displayed on a map and used to filter results
- [ ] POC for allowing ACs to control which events appear on their website. Probably a global collection that allows them to specify other avalanche centers and external courses (filtered by provider and/or location) that should show up on their website.
- [ ] POC for event instances / events with multiple dates

rustynwac:
Second bullet covered by #673 
Hold on third

PR #642, opened 2025-10-30
Events location field

busbyk:
## Description

Enhanced the location specific fields for the Events collection. Those fields are now in their own re-usable location field which is a group of fields. They use Mapbox to display valid coordinates on a map in the admin panel for convenience. 

## Related Issues

#633 

## Key Changes
- Added Mapbox integration with interactive marker placement (LocationMap.tsx, Marker.tsx)
- Added placeName field (required) for venue/place name
- Changed state from text to select dropdown with all US states
- Added ZIP code validation utility
- Added map coordinates (latitude, longitude, zoom) storage
- Added CopyButton component for copying location addresses (but is a generic component)
- Updated EventPreview and EventPreviewSmallRow to display location with interactive popover
- Added location display to EventMetadata component
- Added location data to event seed files
- Created single-event block seed data
- Added event blocks to "all blocks" seed page for testing
- Added Mapbox environment variables to .env.example

We do now need to set up an NWAC Mapbox account. 

## Screenshots / Demo

CleanShot 2025-10-29 at 21 48 55@2x
CleanShot 2025-10-29 at 21 49 25@2x
CleanShot 2025-10-29 at 21 49 46@2x
CleanShot 2025-10-29 at 21 52 19@2x

## Migration Explanation

Skipping to do one migration when merging `events` branch later.

I wanted to ([and did implement](https://github.com/NWACus/web/blob/events-location-field-with-picker/src/fields/location/components/LocationPicker.tsx)) a location picker using Mapbox and their SearchBox component. Unfortunately, we can't store SearchBox API results (which is what we get back from the SearchBox component) cannot be stored permanently ([per the Mapbox terms](https://docs.mapbox.com/api/search/search-box/#search-box-api-restrictions-and-limits:~:text=The%20Mapbox%20Terms%20of%20Service%20state%20that%20all%20data%20returned%20by%20the%20Search%20Box%20API%20endpoints%20is%20only%20available%20for%20temporary%20use.%20If%20your%20use%20case%20requires%20storing%20position%20data%2C%20contact%20Mapbox%20sales.)) so I just put that in a branch for now. 

If we want to implement this we could follow up the `onRetrieve` result with a call to the [Geocoding API with a `permanent=true` option](https://docs.mapbox.com/api/search/geocoding/#storing-geocoding-results). This would just [incur cost](https://www.mapbox.com/pricing#search) - about $0.005 per call for our expected volume.


src/components/EventPreview/index.tsx
```diff
@@ -162,19 +164,21 @@ export const EventPreview = (props: {
...
-          <ImageMedia
-            imgClassName="w-full h-full object-cover transition-transform duration-300 group-hover:scale-105"
-            resource={featuredImage}
-            pictureClassName="w-full h-full"
-          />
-        )}
-      </Link>
+      <div className="flex flex-col @lg:items-end gap-1 mb-2 @lg:mb-0">
+        {location && <LocationPopover location={location} />}
+        <Link
```
rchlfryn, line 171: Since `eventUrl` is optional, we will want to add a check here (multiline suggestion didn't work)

```ts
{ eventUrl && (<Link
   ...
```

src/fields/location/index.ts
```diff
@@ -0,0 +1,171 @@
...
+        label: 'City',
+        admin: {
+          condition: (_data, siblingData) => !siblingData?.isVirtual,
+        },
+      },
+      {
+        name: 'state',
+        type: 'select',
+        label: 'State',
+        options: [
```
rchlfryn, line 38: Random thought: do we need all the states? It definitely easier to list all of them then maintain a list.

src/components/EventPreview/LocationPopover.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <div>
+            {location.city && `${location.city}, `}
+            {location.state && `${location.state} `}
+            {location.zip}
+          </div>
+        </div>
+        <div className="flex gap-2">
+          <CopyButton
+            text={[
+              location.placeName,
```
rchlfryn: I'm not sure we should include the placeName in the copy because it can refer to anything, and Google tends to prioritize that in search results over the actual street address. You can see this from the test data I clicked—although the full address isn’t real, the street address is valid, and Google still tries to find "Mountain Community Center" instead of "123 Alpine Way."

src/components/EventMetadata/index.tsx
```diff
@@ -90,12 +90,28 @@ export const EventMetadata = ({
...
                 {showLabels && <span className="font-medium">Location: </span>}
-                {location.venue && <div>{location.venue}</div>}
+                {location.placeName && <div>{location.placeName}</div>}
                 {(location.address || location.city || location.state) && (
                   <div>
-                    {location.address && <span>{location.address}, </span>}
-                    {location.city && <span>{location.city}, </span>}
-                    {location.state} {location.zip}
+                    {location.address && (
+                      <span>
```
rchlfryn: Do we need spans around each one of these? Or just around all of it?
Also remove the spaces at the end or combine them with the commas.

src/components/EventPreviewSmallRow.tsx
```diff
@@ -37,6 +37,19 @@ export const EventPreviewSmallRow = (props: { className?: string; doc?: Event })
...
 
+  // Build location display text with fallbacks
+  const getLocationText = () => {
+    if (!location) return null
+    if (location.isVirtual) return 'Virtual'
+    if (location.placeName) return location.placeName
+    if (location.city && location.state) return `${location.city}, ${location.state}`
+    if (location.city) return location.city
+    if (location.state) return location.state
+    return null
```
rchlfryn: It will likely never get to null but should the default be 'Location'? if someone only uses coordinates?

rchlfryn (changes requested):
This is a great feature! just a couple of comments

src/components/EventPreview/index.tsx
busbyk: Good call. For the clickable title and image, I still want those to show up if there's no eventUrl so I set the href to '#' instead of more complicated logic to determine whether or not to use a Link or a div as a fallback. 

[cc77163](https://github.com/NWACus/web/pull/642/commits/cc77163f76b6de466734b526d6b35b5f570ac1a3)

src/fields/location/index.ts
busbyk: What do you mean do we need all of them? 

I moved this into a shared list in #658

src/components/EventPreview/LocationPopover.tsx
busbyk: Ah good call. Removed. [04d7f68](https://github.com/NWACus/web/pull/642/commits/04d7f682473949991b5109849a75866708f13f98)

src/components/EventMetadata/index.tsx
busbyk: Yea this was ugly. Updated: [8a82b13](https://github.com/NWACus/web/pull/642/commits/8a82b13e44cc480a7029a81c422f5f6c4ffe4bee)

src/components/EventPreviewSmallRow.tsx
busbyk: I like it! [449df5a](https://github.com/NWACus/web/pull/642/commits/449df5a959c413c86b0151cfea19a195fb9f1636)

src/fields/location/index.ts
rchlfryn: I was thinking we don't need to maintain all states since we are only on the west coast but leave it since its no cost to leave it but a cost to maintain.

rchlfryn (approved):

src/fields/location/index.ts
busbyk: What about NH though!

Haha but I think a state list is totally fine to maintain - it shouldn't ever change unless there's some pretty serious political upheaval but we won't go there :)

PR #648, opened 2025-11-03
Events UI filters

### LINKED ISSUE #631: Enhance EventsCollection UI filtering
Enhance EventsCollection UI filtering
- "Quick filters" for dates: this week, next week, this month, next month, January, February, etc.
- Custom dates filtering
- Indeterminate logic for event type -> sub type filters (i.e. you select the parent event type and all sub types are checked but then deselecting a sub type makes the parent type render as indeterminate and clicking it would set all as checked again, then all unchecked if clicked again)

rchlfryn:
## Description
This PR enhances the UI filters we use for events (types and date)

When a user first lands on a page, only `Upcoming` and `Hide past events` should be selected.

## Related Issues
Fixes #631 

## Key Changes
- Adds filters for date
  - Quick filters with Upcoming, This Week, This Month, Next Week, Next Month, Past Events, Custom
- Updates type filters
  - ~Adds indeterminate state to types with subtypes~
- Add mobile version

## Screenshots / Demo
Screenshot 2025-10-31 at 12 43 29
Screenshot 2025-11-01 at 16 00 51

## Migration Explanation
Yes - but we will do this later

Do we want to add any other quick filters?

When a users interacts with the filters then navigates to events again from the menu, it still keep the previous state...


rustynwac:
Tell me to get out of here if I shouldnt be here, but I think we might be missing Events by Others for Type (there's a lot of these). 

ie Baker Splitfest. Hosted by someone else but NWAC lists it as they give all the proceeds to NWAC

src/app/(frontend)/[center]/events/events-date-filter.tsx
Comment (busbyk) [inline src/app/(frontend)/[center]/events/events-date-filter.tsx:1]: I'm not a fan of the `Hide past events` checkbox. I think it's a little confusing and the fact that it can't always be checked feels like it might confuse users. I think it would be nicer to just lean on utilizing start and end date + sort to accomplish displaying past events. 

It feels like there's a lot going on here too: many state variables, the switch case logic is significant, there's a lot of individual `set____` calls. 

I know we'll likely still need many state variables and potentially still need the filterType here but I wonder if using a list of filter definitions could simplify this component. 

So:
1. Do away with `Hide past events` completely and depend on startDate, endDate, and sort direction. We could have a past events quick filter with just an endDate, and a descending sort direction. 
2. Use the quick filters definition object to hopefully simplify the many setState calls. 

Something like:

```tsx
interface QuickFilter {
  key: string
  label: string
  startDate: () => string  // Returns ISO date string
  endDate: () => string | null  // Returns ISO date string or null for "no end"
  sortDirection: 'asc' | 'desc'  // How events should be sorted for this filter
}

const QUICK_FILTERS: QuickFilter[] = [
  {
    key: 'upcoming',
    label: 'Upcoming',
    startDate: () => new Date().toISOString().split('T')[0], // Today
    endDate: () => null, // No end date
    sortDirection: 'asc', // Soonest events first
  },
  {
    key: 'this-week',
    label: 'This Week',
    startDate: () => {
      const today = new Date()
      return today.toISOString().split('T')[0]
    },
    endDate: () => {
      const today = new Date()
      const endOfWeek = new Date(today)
      endOfWeek.setDate(today.getDate() + (6 - today.getDay())) // Next Saturday
      return endOfWeek.toISOString().split('T')[0]
    },
    sortDirection: 'asc', // Soonest events first
  },
  {
    key: 'next-week',
    label: 'Next Week',
    startDate: () => {
      const today = new Date()
      const nextSunday = new Date(today)
      nextSunday.setDate(today.getDate() + (7 - today.getDay())) // Next Sunday
      return nextSunday.toISOString().split('T')[0]
    },
    endDate: () => {
      const today = new Date()
      const nextSaturday = new Date(today)
      nextSaturday.setDate(today.getDate() + (7 - today.getDay()) + 6) // Next Saturday
      return nextSaturday.toISOString().split('T')[0]
    },
    sortDirection: 'asc', // Soonest events first
  },
  {
    key: 'this-month',
    label: 'This Month',
    startDate: () => {
      const today = new Date()
      return today.toISOString().split('T')[0]
    },
    endDate: () => {
      const today = new Date()
      const lastDayOfMonth = new Date(today.getFullYear(), today.getMonth() + 1, 0)
      return lastDayOfMonth.toISOString().split('T')[0]
    },
    sortDirection: 'asc', // Soonest events first
  },
  {
    key: 'next-month',
    label: 'Next Month',
    startDate: () => {
      const today = new Date()
      const firstDayNextMonth = new Date(today.getFullYear(), today.getMonth() + 1, 1)
      return firstDayNextMonth.toISOString().split('T')[0]
    },
    endDate: () => {
      const today = new Date()
      const lastDayNextMonth = new Date(today.getFullYear(), today.getMonth() + 2, 0)
      return lastDayNextMonth.toISOString().split('T')[0]
    },
    sortDirection: 'asc', // Soonest events first
  },
  {
    key: 'past-events',
    label: 'Past Events',
    startDate: () => {
      // Start from 1 year ago
      const today = new Date()
      const oneYearAgo = new Date(today)
      oneYearAgo.setFullYear(today.getFullYear() - 1)
      return oneYearAgo.toISOString().split('T')[0]
    },
    endDate: () => {
      // End at yesterday
      const today = new Date()
      const yesterday = new Date(today)
      yesterday.setDate(today.getDate() - 1)
      return yesterday.toISOString().split('T')[0]
    },
    sortDirection: 'desc', // Most recent events first
  },
]
```

I also wonder about just using a single custom date input instead of a month picker, year picker, start date picker, end date picker?

```tsx
import { DateRange } from 'react-day-picker'

const [customDateRange, setCustomDateRange] = useState<DateRange | undefined>()

// When user selects custom:
<DayPicker
  mode="range"
  selected={customDateRange}
  onSelect={setCustomDateRange}
  // ... other props
/>

// Then extract start/end from the range:
const startDate = customDateRange?.from?.toISOString().split('T')[0]
const endDate = customDateRange?.to?.toISOString().split('T')[0]
```

src/app/(frontend)/[center]/events/events-date-filter.tsx
```diff
@@ -0,0 +1,334 @@
...
+    const monthNum = parseInt(month)
+    const yearNum = parseInt(year)
+    const start = new Date(yearNum, monthNum, 1).toISOString().split('T')[0]
+    const end = new Date(yearNum, monthNum + 1, 0).toISOString().split('T')[0]
+    setSelectedMonth(month)
+    setSelectedYear(year)
+    updateDateSelection('custom', start, end)
+  }
+
+  const clearFilter = () => {
```
Comment (busbyk) [inline src/app/(frontend)/[center]/events/events-date-filter.tsx]: Instead of clearing the filters when the user clicks X what about setting it back to "Upcoming"? That makes more sense to me and would be needed if we do away with "hide past events".

src/app/(frontend)/[center]/events/page.tsx
```diff
@@ -51,32 +56,68 @@ export default async function Page({ params, searchParams }: Args) {
...
+        <div className="grow order-2 md:order-1">
           <EventCollection events={events.docs} />
         </div>
 
-        <div className="flex flex-col shrink-0 justify-between md:justify-start md:w-[240px] lg:w-[300px]">
-          <EventsFilters types={eventTypesData} subTypes={eventSubTypesData} />
+        {/* Desktop Sidebar - Hidden on Mobile */}
+        <div className="hidden md:flex flex-col shrink-0 justify-between md:justify-start md:w-[240px] lg:w-[300px] order-1 md:order-2">
+          <EventsTypeFilter types={eventTypesData} subTypes={eventSubTypesData} />
+          <EventsDatePicker startDate={selectedStartDate} endDate={selectedEndDate} />
```
Comment (busbyk) [inline src/app/(frontend)/[center]/events/page.tsx:106]: I think if we have the date filters in the sidebar on desktop we should not render them as button groups, but something with spacing like this:

Image

Originally, I was thinking that we'd display these horizontally across the top of the page like this:

Image

What do you think about that?

src/app/(frontend)/[center]/events/events-mobile-filters.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export const EventsMobileFilters = ({ eventCount, types, subTypes, hasActiveFilters }: Props) => {
+  const [isOpen, setIsOpen] = useState(false)
+  const [typeOpen, setTypeOpen] = useState(true)
+  const [dateOpen, setDateOpen] = useState(true)
+
+  return (
+    <>
+      <Button variant="outline" onClick={() => setIsOpen(true)} aria-label="Open filters">
```
Comment (busbyk) [inline src/app/(frontend)/[center]/events/events-mobile-filters.tsx]: I think it would be nice if this was full width. 

Image

src/app/(frontend)/[center]/events/events-mobile-filters.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+      {isOpen && (
+        <div
+          className="fixed inset-0 bg-black/50 z-40 md:hidden"
+          onClick={() => setIsOpen(false)}
+          aria-hidden="true"
+        />
+      )}
+
+      {/* Drawer */}
```
Comment (busbyk) [inline src/app/(frontend)/[center]/events/events-mobile-filters.tsx:43]: I like the drawer!

src/app/(frontend)/[center]/events/events-type-filter.tsx
Comment (busbyk) [inline src/app/(frontend)/[center]/events/events-type-filter.tsx:1]: I love the indeterminate logic, works really well. Sadly we're removing subType now since `courses-by-external-providers` is going away in favor of the Courses collection in #658 😅

busbyk (commented):
> When a users interacts with the filters then navigates to events again from the menu, it still keep the previous state...

I think that's fine!

src/app/(frontend)/[center]/events/events-type-filter.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/events/events-type-filter.tsx:1]: I went ahead and removed all subtype references in [75ea9b4](https://github.com/NWACus/web/pull/648/commits/75ea9b432241c1ac59d36d75f9b6115781932a36). We still have the constant but that will be removed in another PR

src/app/(frontend)/[center]/events/events-date-filter.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/events/events-date-filter.tsx]: Love it - I think with the past dates I wanted to give the user the ability to completely clear the date selection to show all dates, but now with the new filters, it makes the most sense to set it as Upcoming when clearing the filter.

src/app/(frontend)/[center]/events/events-date-filter.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/events/events-date-filter.tsx:1]: I like the suggested update. I went ahead and changed how we render the quick filter buttons in [040ec13](https://github.com/NWACus/web/pull/648/commits/040ec13d2c1a537b8389918c0efca15e9c44a44e)

I am HAPPY to get rid of the month/year picker, so I did that in [049182d](https://github.com/NWACus/web/pull/648/commits/049182d35c99197193b1248002149860d999239d)

I am torn on this type of date picker. I enjoy these for week trips or something within the same month but find them annoying to use otherwise. Maybe we get feedback on this.

src/app/(frontend)/[center]/events/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/events/page.tsx:106]: I tried to combine the two which is why I landed where I did. I updated the buttons in [6c84245](https://github.com/NWACus/web/pull/648/commits/6c84245c4b5402756257dd0fa10fedd84889f179) to look more like the first screen shot. Still not stoked about the sidebar look but we can iterate.

src/app/(frontend)/[center]/events/events-mobile-filters.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/events/events-mobile-filters.tsx]: I don't like altering the width of buttons but agree it's a better UX for now.. reluctantly updated in [6c84245](https://github.com/NWACus/web/pull/648/commits/6c84245c4b5402756257dd0fa10fedd84889f179)

Comment (busbyk) [reply src/app/(frontend)/[center]/events/events-mobile-filters.tsx]: I'm still seeing the button as content width? 

Image

I think you could just give it the `w-full` class.

src/app/(frontend)/[center]/events/events-date-filter.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/events/events-date-filter.tsx:1]: > I am torn on this type of date picker. I enjoy these for week trips or something within the same month but find them annoying to use otherwise. Maybe we get feedback on this.

Fair enough. I'm happy with there being a start and end date input here. Makes it easier to leave one empty if you only want to set one or the other. 

Why not use the shadcn [DatePicker with Input](https://ui.shadcn.com/docs/components/date-picker#picker-with-input): 
Image

1. I think it looks nicer and 2. it has some nice behaviors like when you click the icon in the input when the date picker is open, it closes the date picker. 

I also think we should have a way of clearing an individual date picker to leave endDate empty for example.

busbyk (changes requested):
- I got a 500 when clicking to the next page of events results: 
Image

- Tablet needs some love. The custom date pickers overflow their container (should wrap). And the EventPreview component needs some tweaking of the container query breakpoints. The EventPreview component is unrelated to this PR though. 
Image

- There is no way to clear filters on mobile

- Type labels: on mobile you cannot click the label to select the checkbox. On desktop you can but the cursor is default not pointer on hover. 

- Image

rchlfryn:
Updated everything we talked about

Issue #625, opened 2025-10-20
Duplicate page to... enhancements


busbyk:
Right now the duplicate page to button copies _all_ content from a page document except for top level ids. 

See https://github.com/NWACus/web/blob/main/src/collections/Pages/endpoints/duplicatePageToTenant.ts

This is problematic because it copies relationship blocks as well (there is a TODO about media in that file but this applies to all relationship blocks). These relationship blocks are all tenant-scoped so it doesn't make sense to point to the source tenant's documents. 

We should either:
1. Remove all relationship blocks and leave block placeholders that will throw validation errors after duplication when attempting to save/publish the new document, forcing the user to repopulate those blocks. 
2. Create duplicate documents in the destination tenant and update the references before duplicating the page. 

Option 2 This makes sense for some documents like Media but not for others like Biographies. You might want an NWAC version of a photo from the Sierra page) but you don't want an NWAC version of a Sierra-specific Biography (or Sponsor and so on). So option 1 probably makes more sense even though it's more work for the user. 

I suppose you could also remove the relationship blocks completely but that might defeat the purpose of copying?

rustynwac:
I think I like option 1 over 2

PR #649, opened 2025-11-03
Multi-tenant user enhancements

busbyk:
## Description

The key change here allows users to always `read` tenants that they have a tenant role assignment for. This was needed because if a user had `read` permission on `pages` for two tenants but no other permissions, the tenant selector would never be displayed (because the user couldn't `read` tenants). 

Now, a user can always `read` tenants that they have assignments for. 

## Related Issues

A temporary solution related to #625 until we have global pages is to allow centers to see existing centers' content as examples. 

## Key Changes

Uses a similar pattern to `roleAssignments` where [users can read their own assignments](https://github.com/NWACus/web/blob/main/src/collections/RoleAssignments/access/byTenantRoleWithSelfRead.ts#L7) but we [hide the `roleAssignments` collection from the nav if a user _only_ has `read`](https://github.com/NWACus/web/blob/main/src/collections/RoleAssignments/index.ts#L18). The `tenants` collection gets the same pattern now.

### DIFF
--- a/src/access/byTenantRole.ts
+++ b/src/access/byTenantRole.ts
@@ -46,6 +46,12 @@ export const byTenantRole: (method: ruleMethod, collection: ruleCollection) => A
 
       if (tenantFromCookie) {
         return matchingTenantIds.includes(tenantFromCookie)
+      } else {
+        return {
+          tenant: {
+            in: matchingTenantIds,
+          },
+        }
       }
     }
 
--- a/src/access/byTenantRole.ts
+++ b/src/access/byTenantRole.ts
@@ -9,7 +9,7 @@ import { Access, CollectionConfig } from 'payload'
 // tenant-scoped collections
 export const byTenantRole: (method: ruleMethod, collection: ruleCollection) => Access =
   (method: ruleMethod, collection: ruleCollection): Access =>
-  async ({ req: { user, payload } }) => {
+  async ({ req: { user, payload, headers } }) => {
     if (!user) {
       return false
     }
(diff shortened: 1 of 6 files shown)


rchlfryn (changes requested):
For some reason I can create a draft page for the tenant I am assigned read access to. Can we fix this?

If I am a DVAC admin with NWAC read permissions, I can see the Create Page button when NWAC is selected.

It looks like we need to update `accessByTenantRole`

rchlfryn (changes requested):
It looks like the tenant cookie is not getting set on log in. Payload fixed a [similar issue in 59.1.0](https://github.com/payloadcms/payload/pull/14107)
https://www.loom.com/share/238be6d65af140c6bd3d9d4a746d0671

stevekuznetsov (approved):

stevekuznetsov:
Direction lgtm

busbyk:
> It looks like the tenant cookie is not getting set on log in. Payload fixed a [similar issue in 59.1.0](https://github.com/payloadcms/payload/pull/14107) https://www.loom.com/share/238be6d65af140c6bd3d9d4a746d0671

Thanks for catching this!

Looks like [this line](https://github.com/NWACus/web/blob/multi-tenant-user-updates/src/access/byTenantRole.ts#L45) was executing before the tenant cookie was set, resulting in no where clause being returned. 

I also pulled in the change from https://github.com/payloadcms/payload/pull/14107 because why not: [668f941](https://github.com/NWACus/web/pull/649/commits/668f94118e8c04ebea1f35f99974816373ebc27f)

rchlfryn (approved):
Awesome! Those updates work... It is a little weird still being able to see `Media`, `Documents` and `Users`, but there is nothing malicious that can be done. 🚢

Issue #635, opened 2025-10-25
Events migration


rchlfryn:
Ensure migrations have been created

PR #653, opened 2025-11-05
Events: blocks single and list

### LINKED ISSUE #245: Blocks: Events listing
A block that displays a list of events:

Image
These are two different instances of this block with different filters.

Should accept optional:
- heading/title
- categories filter

TBD if we should support this "pinned" or a highlighted event

rchlfryn:
## Description
Updates event blocks

## Related Issues
Fixes #245 

## Key Changes
- Fixes query params for event list block (should have been done in #637)
- Adds thumbnails for blocks
- **Changes autosave on pages and posts to ~include save draft button and~ default time of 800ms**
- Fixes background color not working for events and posts blocks

## Screenshots / Demo
Screenshot 2025-11-05 at 12 33 54
Screenshot 2025-11-05 at 12 34 28

## Migration Explanation
yes but in #635 

- When these blocks are used as the lexical version, the background color selection UI is pretty bad... #654 
- Responsiveness is not great on the bigger tablet size #655


src/blocks/BlogList/Component.tsx
```diff
@@ -46,7 +46,7 @@ export const BlogListBlockComponent = async (args: BlogListComponentProps) => {
   const bgColorClass = `bg-${backgroundColor}`
 
   return (
-    <div className={cn(wrapInContainer && bgColorClass)}>
+    <div className={cn(bgColorClass && `${bgColorClass}`)}>
```
busbyk: Well, there's only space to show the background color if you're wrapping the component in a container. So I actually think we should keep this, right?

busbyk (commented):
We're def getting some conflicts here with what I've been doing in #658 now 😬 because I removed sub types. Maybe it makes sense to pull those changes out into a separate PR to merge in first? Or I'm happy enough with this PR going in and then dealing with the merge conflicts in #658

src/blocks/BlogList/Component.tsx
rchlfryn: we can keep `wrapInContainer` as the logic but that wasnt the problem here. It was just evaluating to whatever the boolean of `wrapInContainer && bgColorClass` was rather than assigning `bgColorClass` as a string. I have a follow up issue in #654

rchlfryn:
>We're def getting some conflicts here with what I've been doing in https://github.com/NWACus/web/pull/658 now 😬 because I removed sub types. Maybe it makes sense to pull those changes out into a separate PR to merge in first? Or I'm happy enough with this PR going in and then dealing with the merge conflicts in https://github.com/NWACus/web/pull/658

It seems like more work to remove them into a separate PR. Since in #658 removing subtypes is necessary, lets just resolve conflicts there (which hopefully will not be too difficult)

rchlfryn:
I changed my mind - I will remove subtype in any of these files

busbyk (approved):

PR #658, opened 2025-11-07
A3 Providers and Courses

### LINKED ISSUE #331: A3 user management in admin panel
External course provider global role -- can only view content related to their course provider
- A3 staff user management (maybe same as above tbh)
- update users collection access control to only display users with global roles that the logged in user also has
- A3 staff global role invites all course provider users
- Either docs or some UI to help with generating the embed script for 3rd party sites

busbyk:
## Description

A3 provider, user, and course management in the admin panel. 

## Related Issues

Resolves #331 

## Key Changes

Main changes:
- Creates a Courses collection for external provider courses
- Creates a Providers collection to represent external providers and grant access to provider courses for users with provider relationships (see `/docs/decisions/012-non-rbac-access-patterns.md`)
- Creates an A3Management global that's used to indicate which global role is the "provider manager"
	- Restricts users with the provider manager role from changing the role selected here even if they have been given access via a global role (unless super admin)
- Global roles & super admin permission check enhancements / fixes: enables non-super admin, global-role-only users in the admin panel
	- accessByGlobalRoleReadOnly function for diagnostics global (pretty unrelated change but noticed this when addressing global role issues so leaving it in)
	- Fixes check to see if user is a "super admin" (i.e. can do every action on every collection/global). Was previously checking to see if the user had permission to do any action on any collection but we never noticed because there was only one global role. Updates usage in DuplicatePageTo.
	- Allows setting globals' slugs in role rules in addition to collection slugs for global roles only
	- Allows reading global roles if authenticated but keeping list view hidden from admin panel if user only has read access (from the authenticated read). This is to allow users with global roles to read their own role in their account view. 
- Removes `course-by-external-provider` as an event type in favor of the Courses collection
	- Removes eventSubTypes in favor of courseTypes
	- Removes event blocks' eventSubType filtering
- Abstracts shared field configs from Events into their own factory functions
	- startAndEndDateFields
	- modeOfTravelField (makes this a multi-select and removes `any` as an option)
	- coordinatesWithMap
- Adds notification emails for Providers being approved (published) and Providers' approved courseTypes changing
- Users updates
	- Adds a Providers relationship to users which grants provider-user access
	- Updates Users collection access function to respect provider managers and provider users
	- Enables inviting users with Provider relationships, including for provider managers
- Hides Media and Documents from the nav if the user does not have explicit global or tenant role read access to them (used to show because those collections have public read for displaying on tenant websites but we don't want that in the admin panel)
- Restricts authenticated read on EventGroups, EventTags, Events
- Updated seed script accordingly
- Allows the slug field to be used for collections without a tenant
- Enables access-style functions that accept a `User | ClientUser` and check for specific permission (see changes under `/utilities/rbac`)
- Decision doc: `/docs/decisions/012-non-rbac-access-patterns.md`

Misc. small changes:
- Adds `_status` = published filter to /events/[slug] generateStaticParams
- Moves stateOptions to a constants file under the location field
- Re-uses the BlogList and BlogPost block thumbnails for the EventList and SingleEvent blocks since those thumbnails match the event blocks' UI too

## Migration Explanation

Will add migration in #635 

The actual embeds will be in a future PR.


src/app/(frontend)/[center]/events/[slug]/page.tsx
```diff
@@ -25,6 +25,11 @@ export async function generateStaticParams() {
       tenant: true,
       slug: true,
     },
+    where: {
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/events/[slug]/page.tsx:30]: 👏  and weird we still have to do this

src/collections/Courses/components/CourseTypeField.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type Props = SelectFieldClientProps
+
+export const CourseTypeField = (props: Props) => {
+  const providerField = useFormFields(([fields]) => fields.provider)
+  const { value: courseTypeValue, setValue: setCourseTypeValue } = useField<string>({
+    path: 'courseType',
+  })
+
+  const [providerCourseTypes, setProviderCourseTypes] = React.useState<string[] | null>(null)
```
rchlfryn, line 18: Can we import these instead to match the rest of our code?
```ts
import {useState} from React;
const [providerCourseTypes, setProviderCourseTypes] = useState<string[] | null>(null)
```

src/collections/Courses/components/CourseTypeField.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  // Fetch provider's approved courseTypes when provider changes
+  React.useEffect(() => {
```
rchlfryn: Same for `useEffect`

src/blocks/EventList/config.ts
```diff
@@ -161,7 +159,7 @@ const staticEventRelatedFields: Field[] = [
 const eventListBlockWithFields = (fields: Field[]): Block => ({
   slug: 'eventList',
   interfaceName: 'EventListBlock',
-  imageURL: '/thumbnail/EventListThumbnail.jpg',
+  imageURL: '/thumbnail/BlogListThumbnail.jpg',
```
rchlfryn: this should be 
```suggestion
  imageURL: '/thumbnail/EventListThumbnail.jpg',
```

src/blocks/SingleEvent/config.ts
```diff
@@ -19,7 +19,7 @@ const defaultFields: Field[] = [
 const singleEventWithFields = (fields?: Field[]): Block => ({
   slug: 'singleEvent',
   interfaceName: 'SingleEventBlock',
-  imageURL: '/thumbnail/SingleEventThumbnail.jpg',
+  imageURL: '/thumbnail/SingleBlogPostThumbnail.jpg',
```
rchlfryn: Same here. this should be 
```suggestion
  imageURL: '/thumbnail/SingleEventThumbnail.jpg',
```

src/collections/Users/index.ts
```diff
@@ -70,6 +70,18 @@ export const Users: CollectionConfig = {
...
       },
     },
+    {
+      name: 'providers',
+      type: 'relationship',
+      relationTo: 'providers',
+      hasMany: true,
+      saveToJWT: true,
+      maxDepth: 3,
+      // TODO
```
rchlfryn: Do we still need this TODO?

src/utilities/validateUrl.ts
rchlfryn, line 1: 👏

src/utilities/validateUrl.ts
```diff
@@ -1,5 +1,60 @@
...
-import isAbsoluteUrl from './isAbsoluteUrl'
 
-export const validateExternalUrl: TextFieldSingleValidation = (val) =>
-  isAbsoluteUrl(val) || 'URL must be an absolute url with a protocol. I.e. https://www.example.com.'
+/**
+ * Validates whether a string is a valid, full URL with protocol, host, and proper TLD
+ */
+const isValidFullUrl = (url?: string | null): boolean => {
+  if (url == null || typeof url !== 'string' || url.trim() === '') {
+    return false
```
rchlfryn, line 8: I think all of these `false` returns need to be a string. I see the || at the bottom but it might be nice to return some specific errors where necessary.

src/collections/Courses/components/CourseTypeField.tsx
```diff
@@ -0,0 +1,77 @@
...
+      } catch (error) {
+        console.error('Failed to fetch provider data:', error)
+      }
+    }
+
+    fetchProviderData()
+  }, [providerField?.value])
+
+  // Filter options based on type and provider's approved courseTypes
+  const filteredOptions = React.useMemo(() => {
```
rchlfryn: Same here. useMemo

src/collections/Courses/index.ts
```diff
@@ -0,0 +1,216 @@
...
+    {
+      name: 'timezone',
+      type: 'select',
+      options: TIMEZONE_OPTIONS,
+      admin: {
+        description: 'Event timezone',
+      },
+    },
+    {
+      name: 'location',
```
rchlfryn, line 53: Is this the same location object as events? Do we want to reuse it?

src/collections/Providers/index.ts
```diff
@@ -0,0 +1,165 @@
...
+        },
+        {
+          name: 'website',
+          type: 'text',
+          validate: validateExternalUrl,
+        },
+      ],
+    },
+    {
+      name: 'location',
```
rchlfryn, line 56: Could this also be reused with events and course? If the answer is no that's fine too

src/collections/Roles/components/CollectionsField.tsx
```diff
@@ -1,25 +1,39 @@
...
-  readOnly,
-}) => {
+export const CollectionsField: TextFieldServerComponent = async (props) => {
+  const { payload, path, field, clientField, readOnly } = props
   const fieldPath = (path || field?.name || '') as string
   const { type: _type, admin, ...clientFields } = clientField
 
+  // Get includeGlobals from serverProps
+  const serverProps =
+    typeof field.admin?.components?.Field === 'object' &&
```
rchlfryn, line 12: These type checks still make me sad

src/emails/provider-application-approved.tsx
```diff
@@ -0,0 +1,76 @@
...
+  courseTypes,
+}: ProviderApplicationApprovedProps) {
+  const providersUrl = `${appUrl}/providers`
+
+  return (
+    <EmailLayout appUrl={appUrl}>
+      <Heading as="h1" style={{ textAlign: 'center' }}>
+        Application Approved!
+      </Heading>
+      <Text style={{ textAlign: 'center', fontSize: '16px', marginBottom: '24px' }}>
```
rchlfryn, line 31: Is there a reason we have so many `<Text>` tags? Styling I assume?

src/collections/Courses/index.ts
```diff
@@ -0,0 +1,216 @@
...
+    {
+      name: 'registrationDeadline',
+      type: 'date',
+      admin: {
+        date: {
+          pickerAppearance: 'dayAndTime',
+        },
+        description: 'Registration cutoff',
+      },
+      validate: (value, { siblingData }) => {
```
rchlfryn, line 121: :pick: I think you can avoid the `as` here by defining typing `siblingData: LinkFields`

src/app/(frontend)/[center]/events/events-filters.tsx
Comment (rchlfryn) [inline src/app/(frontend)/[center]/events/events-filters.tsx:1]: I think this was a merge conflict. This file is not used anywhere. We should delete it

src/collections/Roles/components/CollectionsField.tsx
```diff
@@ -1,25 +1,39 @@
...
-export const CollectionsField: TextFieldServerComponent = async ({
-  payload,
-  path,
-  field,
-  clientField,
-  readOnly,
-}) => {
+export const CollectionsField: TextFieldServerComponent = async (props) => {
+  const { payload, path, field, clientField, readOnly } = props
   const fieldPath = (path || field?.name || '') as string
```
rchlfryn: ⛏️ 
```suggestion
  const fieldPath: string = (path || field?.name || '')
```

src/emails/provider-application-approved.tsx
```diff
@@ -0,0 +1,76 @@
...
+      </Text>
+    </EmailLayout>
+  )
+}
+
+ProviderApplicationApproved.PreviewProps = {
+  appUrl: 'http://localhost:3000',
+  providerName: 'Mountain Safety Academy',
+  courseTypes: ['rec-1', 'rec-2', 'pro-1'],
+} as ProviderApplicationApprovedProps
```
rchlfryn: I am confused why we are nesting with `PreviewProps` here. Was that intentional?
Can we try to remove the type casting here?

rchlfryn (changes requested):
I took a very high level look at this. The review is mostly syntactically focused, but I did pull down everything and look at it. There are 2 things I think we should consider:

- Can we hide the docs or change the link on the dashboard?
Image

- I really think we should do an events migration before merging this in and then another one for A3 events. I am worried we will miss something or had a bad migration.. Thoughts?

src/app/(frontend)/[center]/events/[slug]/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/events/[slug]/page.tsx:30]: Yea it's definitely a bit unexpected in terms of the default functionality but it's def the _intended_ functionality by the Payload team: https://github.com/payloadcms/payload/issues/8077#issuecomment-2419432940

src/app/(frontend)/[center]/events/events-filters.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/events/events-filters.tsx:1]: Yep! Thought I accepted this incoming change but apparently not.

src/collections/Courses/components/CourseTypeField.tsx
busbyk: I found all other usages of React.{hook} and replaced with direct hook references except for those in `/components/ui` files because that's the patterned preferred by shadcn and 1. that would be a lot of unnecessary changes and 2. it might be easier to diff in the future if we just keep their style.

src/collections/Courses/index.ts
busbyk: It's slightly different. The differences:
- omits additional info
- omits is virtual fields
- state is required here but not in events (because state filtering is so important for courses)

I considered an abstraction to share these but I think duplication is simpler and easier to reason about.

src/collections/Courses/index.ts
busbyk: The validate function is typed so I can't just type that one argument. `siblingData` is typed as `siblingData: Partial<unknown>` from the `DateFieldValidation` type. I think casting is appropriate here, although I do agree it should be more specific. I went with `as Partial<Course>`.

src/collections/Providers/index.ts
busbyk: This is still different than both courses and events 😅

This omits placeName, coordinates, additional info.

src/collections/Roles/components/CollectionsField.tsx
busbyk: Yes, agreed.

src/collections/Roles/components/CollectionsField.tsx
busbyk: Changed 👍

src/collections/Users/index.ts
busbyk: Ah, probably not. Originally, I was going to hide this field from tenant-only uses but I think it's fine for this field to be visible. They can't add anything to it if they don't have access to the providers collection (which they wouldn't unless they have a global role, are a provider manager role, or have provider relationships themselves. Per `src/access/byProviderRelationship.ts`

busbyk: Removed.

src/emails/provider-application-approved.tsx
busbyk: Yea, they are the equivalent of `<p>` tags. So they result in a new line. See https://react.email/docs/components/text

src/emails/provider-application-approved.tsx
busbyk: Hm yea weird. This is just copypasta but I got it originally from their examples (see: https://demo.react.email/preview/notifications/vercel-invite-user?view=source&width=1024&height=600 and go to the code tab). But I removed them and we don't get any typescript errors or any issues running the email dev server. So I'm removing from all email templates. Good catch.

src/utilities/validateUrl.ts
busbyk: I wrote `isValidFullUrl` as a re-usable utility function that always returns boolean. The `validateExternalUrl` function is a `TextFieldSingleValidation` which should return either `true` or an error string. So if `isValidFullUrl` returns false, `validateExternalUrl` will return the error string. This way we can use `isValidFullUrl` elsewhere.

busbyk:
> I took a very high level look at this. The review is mostly syntactically focused, but I did pull down everything and look at it. There are 2 things I think we should consider:

> * Can we hide the docs or change the link on the dashboard?

We have public [A3 provider/courses docs](https://www.notion.so/avy-fx/A3-External-Provider-Courses-2a45af40f1988042bc5cc9ac3f48128c) there too so it feels appropriate to me. Thoughts?

> * I really think we should do an events migration before merging this in and then another one for A3 events. I am worried we will miss something or had a bad migration.. Thoughts?

Since we're only adding things in all of these, I personally don't have any immediate concerns with doing one big migration for the `events` branch. But if it makes you more comfortable to do them progressively, I'm OK with that. 

Since this is getting merged into `events` it logically makes sense to me to have a single migration for `events` -> `main` as well.

rchlfryn (approved):
> We have public [A3 provider/courses docs](https://www.notion.so/avy-fx/A3-External-Provider-Courses-2a45af40f1988042bc5cc9ac3f48128c) there too so it feels appropriate to me. Thoughts?

Maybe we just link to the A3 page? I think either way we need to build out our docs for A3 and maybe we can just wait to change it then.

> Since this is getting merged into events it logically makes sense to me to have a single migration for events -> main as well.

I did a [draft PR of a migration](#635) to see how we were doing, and I think we are okay to wait.

rustynwac:
FWIW I think it's okay to have A3 documentation in the general Avy documentation.

Issue #657, opened 2025-11-06
Clicking on Preview When Creating Content Sends You to an Error Screen


andyanderso:
**Describe the bug**
I get an error screen when I click on on the preview button when I am editting and creating posts.

**To Reproduce**
1. Create or edit some content - I tested it with both pages and posts
2. Click on preview

**Expected behavior**
See the preview of the content displayed

**Screenshots**

<img width="1977" height="1341" alt="Image" src="https://github.com/user-attachments/assets/3d88cc77-0b51-4bc4-bdc5-c0bc749dd758" />

<img width="2357" height="1323" alt="Image" src="https://github.com/user-attachments/assets/adc0e75c-8ad2-483c-95de-daad8ce6cf67" />

**Environment details**

- Desktop - Linux
- Browser - Chrome

**Existential musings**
anyone pining for wordpress yet ;) - that is supposed to make you laugh...

busbyk:
> **Existential musings**
> anyone pining for wordpress yet ;) - that is supposed to make you laugh...

Haha, shots fired! 

Got a fix awaiting review: https://github.com/NWACus/web/pull/664

Should get that released tomorrow or Tuesday.

busbyk:
Fixed by #664

PR #664, opened 2025-11-10
Bugfix: Previews broken for pages, posts

busbyk:
## Description

Previews stopped working. The root cause seems to be that the tenant was no longer being provided as a populated tenant object in the preview functions when it was before. 

## Related Issues

#657 

## Key Changes

- Updates Pages, Posts, HomePages, Navigations `preview` functions to resolve tenants correctly (i.e. fetching the full tenant if tenant is passed as a number)
- Updates livePreview and preview functions to use the same pattern everywhere
- Creates an `isTenantValue` function for type assertions in vaguely typed preview functions (and elsewhere if ever needed)
- Leaves a comment on the defaultPopulate of the tenants collection to remind devs to update `isTenantValue` if the default populate changes

### DIFF
--- a/src/collections/HomePages/index.tsx
+++ b/src/collections/HomePages/index.tsx
@@ -27,6 +27,8 @@ import { quickLinksField } from '@/fields/quickLinksFields'
 import { populatePublishedAt } from '@/hooks/populatePublishedAt'
 import { Tenant } from '@/payload-types'
 import { generatePreviewPath } from '@/utilities/generatePreviewPath'
+import { isTenantValue } from '@/utilities/isTenantValue'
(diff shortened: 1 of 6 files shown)


stevekuznetsov:
@busbyk would it be technically possible to trigger the preview flow programatically? UI testing is hard but "known flow gets no 500s" might be easier?

busbyk:
> @busbyk would it be technically possible to trigger the preview flow programatically? UI testing is hard but "known flow gets no 500s" might be easier?

Yea, we could use Playwright to test the preview flows.

rchlfryn:
Looks like we are still getting preview errors.. Was there a reason you merged this @stevekuznetsov ?

PR #668, opened 2025-11-11
Events - fix filters

rchlfryn:
## Description
When interacting with events page filters and then selecting `Events` from the nav, the params would clear from teh URL but not the UI.

## Related Issues
No issue made. Discovered when working on (event table PR)

## Key Changes
- Clears UI filters on initial mount for date and type filter
- Date filter will select if dates match up with quick filters

## Screenshots / Demo
Before
before_filters

With fix
after_eventsfilters

## Migration Explanation
Nope

Do we need to hide the filters on load until fully loaded to avoid seeing the filters set?

### DIFF
--- a/src/app/(frontend)/[center]/events/events-date-filter.tsx
+++ b/src/app/(frontend)/[center]/events/events-date-filter.tsx
@@ -11,10 +11,10 @@ type Props = {
 }
 
 export const EventsDatePicker = ({ startDate, endDate }: Props) => {
-  const [filterType, setFilterType] = useState('')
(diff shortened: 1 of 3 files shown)


busbyk (approved):
> Do we need to hide the filters on load until fully loaded to avoid seeing the filters set?

My initial reaction is no. I'd rather there be a brief flash of the old filters than a layout shift tbh. 

Although I do wonder if we shouldn't just set the url state based on the application state on client side navigations...

rchlfryn:
I think we will the UI state for now

PR #669, opened 2025-11-11
Events group page and table

### LINKED ISSUE #632: Event table
Add a table of events component that can be used on landing pages for Event Types, Event SubTypes,single Event pages, Event Groups, etc. Just use on single Event page for now.

Image

rchlfryn:
## Description
This PR adds the ability to add a page for event group and adds an event table

## Related Issues
Fixes #632 

## Key Changes
- Add ability to make Event Group page
  - URL: `/events/g/[slug]`
  - Updates breadcrumbs component to hide `g` from UI
- Adds responsive events table block
  - Gives the user to select events individually or by event type 
- Adds table shadcn component 
- Updates seed script 
  - adds event group page 
  - fixes event seeding with registration date past end date
- Consolidate events query config for `EventList` and `EventTable`

## Screenshots / Demo
Meet-Your-Forecaster-Northwest-Avalanche-Center-11-12-2025_11_35_AM

Screenshot 2025-11-12 at 11 58 04

## Migration Explanation
Yes - but done in #635 

TODO: Tablet view of events group


src/fields/EventQuery/QueriedEventsComponent.tsx
rchlfryn, line 1: Where should these files live? There doesn't seem to be a good spot

src/app/(frontend)/[center]/events/g/[slug]/page.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { draftMode } from 'next/headers'
+import { getPayload } from 'payload'
+
+import { LivePreviewListener } from '@/components/LivePreviewListener'
+import { Media } from '@/components/Media'
+import { generateMetaForEvent } from '@/utilities/generateMeta'
+import { cn } from '@/utilities/ui'
+import { Metadata, ResolvedMetadata } from 'next'
+
+export const dynamic = 'force-dynamic'
```
Comment (busbyk) [inline src/app/(frontend)/[center]/events/g/[slug]/page.tsx]: Sounds like we're going to remove these routes but couldn't this route be static?

src/app/(frontend)/[center]/events/g/[slug]/page.tsx
```diff
@@ -0,0 +1,131 @@
...
+      slug: true,
+    },
+  })
+
+  const params: PathArgs[] = []
+  for (const event of events.docs) {
+    if (typeof event.tenant === 'number') {
+      payload.logger.error(`got number for event tenant`)
+      continue
+    }
```
Comment (busbyk) [inline src/app/(frontend)/[center]/events/g/[slug]/page.tsx]: Don't forget you can use `resolveTenant` to ensure we have a tenant instead of throwing if tenant is returned as the tenant id.

src/blocks/EventList/config.ts
```diff
@@ -1,160 +1,26 @@
...
-        label: 'Choose events',
-        relationTo: 'events',
-        hasMany: true,
-        admin: {
-          description: 'Choose new event from dropdown and/or drag and drop to change order',
-        },
-        filterOptions: getTenantFilter,
-      },
-    ],
+const sortByField = (): Field => ({
```
busbyk: We don't have this on the /events list page. I wonder if we even want this? I see the use case for filtering by next registration deadline but I don't see the use cases for latest first for start date and registration date.

src/blocks/EventTable/Component.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  if (filterByEventTypes && filterByEventTypes.length > 0) {
+    eventsLinkQueryParams.set('types', filterByEventTypes.join(','))
+  }
+
+  if (!events) {
+    return null
+  }
+
+  return (
```
busbyk, line 97: Do we want the wrapInContainer + backgroundColor pattern here?

src/fields/EventQuery/config.ts
```diff
@@ -0,0 +1,144 @@
...
+      },
+      {
+        label: 'Let me choose',
+        value: 'static',
+      },
+    ],
+  },
+]
+
+export const dynamicEventRelatedFields = (additionalFilters?: Field[]): Field[] => [
```
busbyk, line 58: What do you think about matching these filters to the filters we have on the /events listing page? i.e. startDate, endDate (via quick filters or custom dates but here it would just be custom dates), event type (which you already have), eventually event groups, and event tags. 

Adding maxEvents def makes sense but I don't know about showUpcomingOnly.

src/fields/EventQuery/QueriedEventsComponent.tsx
busbyk: I think this folder should live in the fields directory since these are re-usable field configs with their associated hooks and components.

src/collections/Events/index.ts
busbyk, line 1: I think we need to add a tenant filter to the event groups field.

busbyk: Tags too

busbyk (commented):
Based on our chat, I think we're going to remove the dedicated event group routes so I didn't review those files. 

I had a hard time reviewing the events table because we need a filter on the event groups in the events collection.

src/app/(frontend)/[center]/events/g/[slug]/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/events/g/[slug]/page.tsx]: I ended up removing this page.

src/app/(frontend)/[center]/events/g/[slug]/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/events/g/[slug]/page.tsx]: This page got removed

src/blocks/EventList/config.ts
rchlfryn: Relooking at this, it doesn't make much sense to keep it.

src/blocks/EventTable/Component.tsx
rchlfryn: I didn't include the background color picker in the event table since I don't think it made sense. Do you think we need it?

src/fields/EventQuery/config.ts
rchlfryn: I am all for keeping DRY code, but I am not sure how we do this efficiently while keeping the admin descriptions for `EventList` and `EventTable`.

I have come around to  `showUpcomingOnly`. I think it makes sense to hide past events on the table but less so on the list... it might be nice to just show everything.

Thoughts?

src/blocks/EventTable/Component.tsx
busbyk: There isn't a specific ask for it or anything but, personally, I think it would be nice for most of our blocks to have the same basic styling options since a user might come to expect to be able to use them for most blocks. If you think it doesn't make sense for this block specifically, I'd hear your argument.

src/components/EventsTable/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+          >
+            Load More
+          </Button>
+        </div>
+      )}
+
+      {/* Results info */}
+      <div className="mt-4 text-sm text-gray-600">
+        Showing {displayedEvents.length} of {sortedEvents.length} events
+      </div>
```
busbyk: I think it would be nice to only show this if the displayed events count is less than the total events count.

src/components/Breadcrumbs/Breadcrumbs.client.tsx
```diff
@@ -53,7 +53,12 @@ const processNestedSegments = (
 
 export function Breadcrumbs() {
   const segments = useSelectedLayoutSegments()
-  const decodedSegments = segments.map(decodeURIComponent)
+  // Exclude the 'g' segment from breadcrumbs, but only when 'g' immediately follows 'events' in the URL path
+  // (e.g., for /events/g/[slug] routes)
+  const decodedSegments = segments
+    .map(decodeURIComponent)
+    .filter((segment, index, arr) => !(segment === 'g' && arr[index - 1] === 'events'))
+
```
busbyk: I think we can revert this change now right since you removed the built in event group pages right?

src/fields/EventQuery/config.ts
```diff
@@ -0,0 +1,167 @@
...
+      },
+      {
+        name: 'showUpcomingOnly',
+        type: 'checkbox',
+        defaultValue: true,
+        label: 'Show Upcoming Events Only',
+        admin: {
+          description: 'Only display events that have not yet occurred.',
+        },
+      },
```
busbyk: I know we've discussed this before but I'm now leaning towards removing this completely. Without a sort option (which I still support having removed) this would be pretty unhelpful to uncheck to show past events. Once a center has more than a few past events, you're just going to see a table of past events from the first event they ever created onwards. 

I think if a center wanted to highlight some past events they could just use the static events option for this block.

src/collections/Events/index.ts
```diff
@@ -32,7 +33,7 @@ import { eventTypesData } from './constants'
 
 export const Events: CollectionConfig = {
   slug: 'events',
-  access: accessByTenantRole('events'),
+  access: accessByTenantRoleWithPermissiveRead('events'),
```
busbyk: If we use permissive read access we should also add a "hidden-if-they-don't-have-explicit-read-permission" like we do with Media and Documents: `hidden: ({ user }) => !hasGlobalOrTenantRolePermission({ method: 'read', collection: 'media', user }),`. Otherwise, any global user in the admin panel will be able to see all events for all tenants even if they don't have a role granting access. Tenant users will always only be able to see their tenant's events because of the tenant baseFilter but a user without events read would still be able to read their tenant's events which might be undesirable for some roles. 

Personally, I would lean towards not granting permissive read and using a custom endpoint/route handler that accepts only the query params you need for the EventList and EventTable block components and returns only the data those blocks need. Limit what data potentially-malicious users could fetch to only what is available on public pages anyways.

src/components/EventsTable/EventsTableWrapper.tsx
busbyk, line 1: This file isn't used anywhere, can you remove?

src/components/EventsTable/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+} from 'lucide-react'
+import { Fragment, useMemo, useState } from 'react'
+import { CMSLink } from '../Link'
+
+export function EventTable({ events = [] }: { events: Event[] }) {
+  const [sortConfig, setSortConfig] = useState({ key: 'startDate', direction: 'asc' })
+  const [displayCount, setDisplayCount] = useState(10)
+  const [expandedRows, setExpandedRows] = useState<Set<string>>(new Set())
+
+  const ITEMS_PER_LOAD = 10
```
busbyk: A few thoughts that are related:
- This component isn't actually fetching additional events when "Load More" is clicked. If we keep this I think we should change this to "Show More". Primarily because the sort applies to _all_ of the events not just the displayed ones and I think that's a bit confusing. 
- I think we should remove the logic to show a different amount of events than are passed in the props. The user already chose a maxEvents prop in the block, so why should we restrict the number shown?
- I don't see any bugs with the way you implemented search (works great) but if/when we want to sort by location (which would be a sorting on just part of the location field - name) we may want to reach for [Tanstack Table](https://tanstack.com/table/latest) (formerly React Table).

src/components/EventsTable/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+                      <>
+                        <CMSLink
+                          appearance="default"
+                          size="sm"
+                          className="group-hover:opacity-90 transition-opacity text-sm"
+                          url={event.registrationUrl}
+                        >
+                          Register
+                          <ExternalLink className="w-3 h-3 sm:w-4 sm:h-4 flex-shrink-0 ml-1 sm:ml-2 -mt-0.5 text-muted hidden sm:inline" />
+                        </CMSLink>
```
busbyk, line 250: Can you update this to open in a new tab?

src/components/EventsTable/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+                      <div className="font-medium">{getLocation(event)}</div>
+                      <div className="text-gray-500 text-xs">{getAddress(event)}</div>
+                    </div>
+                  </TableCell>
+
+                  {/* Register button */}
+                  <TableCell className="hidden sm:table-cell text-right px-2">
+                    {event.registrationUrl && !isPast && !isRegistrationClosed ? (
+                      <>
+                        <CMSLink
```
busbyk, line 242: What are your thoughts on using CMSLink vs. <Link><Button>{content}</Button></Link>. 

It feels like CMSLink should be used for link field data from the CMS and we should use Link + Button when passing just an href. 

But there's nothing functionality wrong here. Just thinking about patterns we want to be using or not.

src/components/EventsTable/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+                      <Button variant="default" disabled={true}>
+                        Closed
+                      </Button>
+                    ) : (
+                      <span className="text-gray-400 text-xs">—</span>
+                    )}
+                  </TableCell>
+                </TableRow>
+                {/* Expanded row for details on smaller screens */}
+                {isExpanded && (
```
busbyk, line 262: I like the expanded UI on smaller screens - nice.

src/collections/EventGroups/index.ts
```diff
@@ -16,13 +16,18 @@ export const EventGroups: CollectionConfig = {
...
+        {
+          name: 'title',
+          type: 'text',
+          required: true,
+        },
+        {
+          name: 'description',
+          type: 'textarea',
+        },
+      ],
```
busbyk: Just to make it look better in the admin panel? Should we make the same change to event tags too? Feels unnecessary to me tbh since these are the only fields in the main document section.

src/blocks/EventTable/useDynamicEvents.tsx
```diff
@@ -0,0 +1,92 @@
...
+            params.append('tags', tagIds.join(','))
+          }
+        }
+
+        if (showUpcomingOnly) {
+          params.append('upcomingOnly', 'true')
+        }
+
+        const response = await fetch(`/api/events?${params.toString()}`, {
+          cache: 'no-store',
```
busbyk: I wonder if we really do want to use 'no-store' here? The approach I took in #675 was to return cache-control headers from a custom endpoint/route handler and let the browser and CDN handle caching based on that.

src/blocks/EventTable/useDynamicEvents.tsx
```diff
@@ -0,0 +1,92 @@
...
+}
+export const useDynamicEvents = ({
+  tenant,
+  filterByEventTypes,
+  filterByEventGroups,
+  filterByEventTags,
+  showUpcomingOnly,
+  maxEvents,
+}: UseEventsProps) => {
+  // ... rest of code
```
busbyk: Can you remove?

src/collections/EventGroups/index.ts
rchlfryn: This was leftover from when the event group had a landing page. removed

src/fields/EventQuery/config.ts
rchlfryn: I removed `showUpcomingOnly` from the Event Query. Now it will only show upcoming events

src/components/EventsTable/index.tsx
rchlfryn: I think CMSLink could use some love and a refactor because I agree with you. The name implies exactly what you said. I just used it here to get the styles for a Link + Button & event tracking `onClickWithCapture`

src/components/EventsTable/index.tsx
rchlfryn: I removed the show more option and will add an issue to add search to the table

src/components/EventsTable/index.tsx
busbyk: Ooo hadn't thought about the `onClickWithCapture` functionality. 

I agree. Created a follow up issue: https://github.com/NWACus/web/issues/685

src/collections/Events/index.ts
rchlfryn: Good callout with the "hidden-if-they-don't-have-explicit-read-permission". I created a custom endpoint in [0108f98](https://github.com/NWACus/web/pull/669/commits/0108f98863e8c9d94bcd719bd1c145010a8c013e). Take a look and let me know if this is what you meant

src/blocks/EventTable/useDynamicEvents.tsx
rchlfryn: Updating caching to match what we are doing with courses

src/blocks/EventTable/Component.tsx
rchlfryn: This was more of an issue when we had a lot of columns, but since the table is full width with a background, the only real color that will show is on the sides.

src/app/api/events/route.ts
busbyk, line 1: I think the only thing missing from the existing https://github.com/NWACus/web/blob/events/src/app/api/%5Bcenter%5D/events/route.ts that you have here is a depth query param. So could you just add that there if you need it?

src/utilities/queries/getEvents.ts
```diff
@@ -35,7 +35,7 @@ export async function getEvents(params: GetEventsParams): Promise<GetEventsResul
 
     const conditions: Where[] = [
       {
-        'tenant.slug': {
```
busbyk, line 38: If you use https://github.com/NWACus/web/blob/events/src/app/api/%5Bcenter%5D/events/route.ts you can keep using slug because we have the center slug as a route param (which the events page expects).

busbyk (commented):
Some initial feedback on these changes.

src/utilities/queries/getEvents.ts
rchlfryn: @busbyk We can sadly only get the tenant id from `src/fields/EventQuery/QueriedEventsComponent.tsx`

src/app/api/events/route.ts
rchlfryn: Is there a reason not to consolidate these files?

src/utilities/queries/getEvents.ts
busbyk: Ah I see. OK well this change (the expectation that the `center` param is an id not a slug) breaks `src/app/(frontend)/[center]/events/page.tsx` because it gets `center` from the route param which is a slug, not an id. `getEvents` is used there too. 

I agree that it would be difficult to get slug not id in `src/fields/EventQuery/QueriedEventsComponent.tsx`. But here's the thing: do we even need `src/fields/EventQuery/QueriedEventsComponent.tsx` anymore? Since we're dynamically fetching events in EventList and EventTable, does it really make sense to show what events will show up when that might change based on when a visitor is visiting the site? 

If we did want to keep a preview of what events will be shown, we could use a virtual field instead since the results don't need to be persisted for revalidation anymore. But we would still have the issue of needing to filter by tenant id not slug. So removing `src/fields/EventQuery/QueriedEventsComponent.tsx` would solve that problem for us. And it's pretty reasonable for a user to preview what events would be shown by opening the live preview view like they do for other changes.

src/app/api/events/route.ts
busbyk: Only https://github.com/NWACus/web/pull/669#discussion_r2542851780 but yea that's what I was suggesting.

busbyk (changes requested):
Main issue is that the /events page is now broken because of the mismatch between tenant slug and id expected in getEvents.

src/utilities/queries/getEvents.ts
rchlfryn: I think you're right in making `Preview` do the work. We don't do that with any other blocks. I did need to update `EventListBlock` to use the endpoint since it was relying on `queriedEvents`, but that is fine. 

We're duplicating the `fetchEvents` data fetch across these two files (table and list). Should we extract this into a custom hook  like, `useFetchEvents` so both files can share it? This would reduce duplication and make future updates easier, but I don't know if we want to start this pattern

src/blocks/EventList/Component.tsx
```diff
@@ -1,43 +1,95 @@
...
+
+      const data = await response.json()
+      setFetchedEvents(data.events || [])
+    }
+
+    fetchEvents()
+  }, [eventOptions, filterByEventTypes, filterByEventGroups, filterByEventTags, maxEvents, tenant])
+
+  let displayEvents
+  if (eventOptions === 'static') displayEvents = staticEvents as Event[]
```
busbyk: You could avoid this type cast by using a similar approach as BlogList to determine whether to show static or dynamic events:

`src/blocks/BlogList/Component.tsx`:
```
  let posts: Post[] = filterValidPublishedRelationships(staticPosts)

  const hasStaticPosts = staticPosts && staticPosts.length > 0

  if (!staticPosts || (staticPosts.length === 0 && queriedPosts && queriedPosts.length > 0)) {
    posts = filterValidPublishedRelationships(queriedPosts)
  }
```

src/blocks/EventTable/Component.tsx
```diff
@@ -0,0 +1,120 @@
...
+      } finally {
+        setLoading(false)
+      }
+    }
+
+    fetchEvents()
+  }, [eventOptions, filterByEventTypes, filterByEventGroups, filterByEventTags, maxEvents, tenant])
+
+  let displayEvents
+  if (eventOptions === 'static') displayEvents = staticEvents as Event[]
```
busbyk: Same thought on casting as https://github.com/NWACus/web/pull/669/files#r2546688310

src/utilities/queries/getEvents.ts
busbyk: What about using the same component for EventList and EventTable with an extra prop that dictates whether to render the event list UI or the event table UI?

busbyk (approved):

src/blocks/EventList/Component.tsx
rchlfryn: Oh good call - added!

Issue #358, opened 2025-07-26
Block(s) to display A3 courses on AC websites


rustynwac:
Rusty to discuss with ACs how to filter. Lat/Long not in scope for v1

rchlfryn:
We will use generic embed block for v1

Issue #317, opened 2025-07-08
A3 courses 3rd party embed


rustynwac:
Filter by course provider, filter by avalanche centers. Also:
Location
Date
Course Level/Type
Course Specificity: Moto/Alpine/Women’s etc.
Reg link

PR #675, opened 2025-11-13
A3 course provider embeds

busbyk:
## Description

This PR introduces two new, non-center-scoped routes: `/embeds/providers` and `/embed/courses`. These routes are intended to be embedded in iframes but they can be visited directly. 

This also introduces:
- an infinite scrolling approach to courses and events and the blog (instead of pagination). 
- shared filter components

## Related Issues

#317 (will wait to close that until the follow up PR) 
Potentially #358 too if we're happy using the Generic Embed block for this. 

### Providers embed
- Matches avalanche.org style for listing providers by state
- Providers could appear in multiple states if they have more than one state in their `statesServiced` field
- Popover shows provider details when provider name is clicked
- This route is statically generated, does not include time-based revalidation, and the Providers collection has revalidation hooks on change/delete
- Query params:
  - backgroundColor: allows users to set a custom background color for the embed
  - title: allows users to set an optional header for the embed

### Courses embed
- Fully dynamic courses listing embed displays a list of courses sorted by date by default, each course is an external link to the provider's course listing
- Query params:
  - backgroundColor: allows users to set a custom background color for the embed
  - title: allows users to set an optional header for the embed
  - showFilters: defaults to false but if true, shows a filters sidebar on desktop and drawer on mobile (like events)
  - a query param for each filter included: course types, providers, states, affinity groups, modes of travel, start date, end date
    - states only show states that appear in published events

### Shared components and patterns with events and blog
- Shared filter components `CheckboxFilter`, `DateRangeFilter`, `MobileFiltersDrawer`
- Infinite scrolling pattern favored over the pagination pattern previously used in events and blog
- Shared states, modes of travel, date quick filters, providers filter
- Adds event groups, event tags, and modes of travel filters to events page

### Misc
- Added iframe embeds in GenericEmbed blocks to the courses-by-external-providers Page created in the seed script for each tenant
- Added an A3 theme that's used in the /embeds routes
- Added A3 fonts used in /embeds routes

## Screenshots / Demo

https://www.loom.com/share/71e8774f100f4176933166cb24df9aaa

A follow-up PR will include:
- ensure PostHog analytics are recorded correctly with the /embeds routes are embedded on third-party sites
- implement iframe-resizer to fit the iframes to the embed content height (particularly important for the accordions on the providers embed)
- ensure our cors policy allows embedding these iframes on third-party sites
- docs

Questions:
- I'm choosing to filter some filters to only those that are actually used in their respective collection like states in courses. But I'm displaying all for things like providers and course types because 1. providers probably all want to be displayed and 2. course types are fixed and users would be looking for specific course types. But things like mode of travel - maybe those should be filtered by only those used. So I'm torn on when to show filtered filters and when to show all. I think the way I have it now is OK but just something to think about. 
  - We'll need a better solution for filtering things like state by only those that are present in courses (same for blog and event filters that use this logic) going forward to avoid performance problems on those queries. A kv store does seem like an option there.


src/utilities/queries/getCourses.ts
```diff
@@ -0,0 +1,173 @@
...
+  try {
+    const { types, providers, states, affinityGroups, modesOfTravel, startDate, endDate } = params
+
+    const offset = params.offset || 0
+    const limit = params.limit || 10
+
+    const conditions: Where[] = []
+
+    if (types) {
+      const selectedTypes = types.split(',').filter(Boolean)
```
stevekuznetsov, line 45: Why does `getCourses()` want users to provide a comma-delimited string as opposed to an array here?

src/utilities/queries/getCourses.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    conditions.push({
+      _status: {
+        equals: 'published',
+      },
+    })
+
+    const where: Where = conditions.length > 0 ? { and: conditions } : {}
+
+    const result = await payload.find({
```
stevekuznetsov, line 148: Does this error/404 if there are none matching your predicates?

src/actions/getCoursesStates.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export async function getCoursesStates(): Promise<GetCoursesStatesResults> {
+  try {
+    const payload = await getPayload({ config })
+
+    // Query all to get unique states
+    const result = await payload.find({
+      collection: 'courses',
+      limit: 10000,
```
stevekuznetsov: RIP our server?

src/actions/getCoursesStates.ts
```diff
@@ -0,0 +1,46 @@
...
+      select: {
+        location: true,
+      },
+    })
+
+    // Extract unique state values
```
stevekuznetsov: Do we lose the ability to `SELECT DISTINCT` across the drizzle bridge? This is an ... inefficient way to get at this info 😓

src/actions/getEventGroups.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export interface GetEventGroupsResults {
+  eventGroups: { label: string; value: string }[]
+}
+
+export async function getEventGroups(params: GetEventGroupsParams): Promise<GetEventGroupsResults> {
+  try {
+    const payload = await getPayload({ config })
+
+    // Query all events for the center to get unique event groups
```
stevekuznetsov: Ditto here - loading 10k event groups into memory to filter them in the JS layer seems like a weird thing to do unless it's our only option

src/app/(frontend)/[center]/events/page.tsx
```diff
@@ -53,88 +64,64 @@ export default async function Page({ params, searchParams }: Args) {
...
-      },
-    })
-  } else if (selectedEndDate) {
-    conditions.push({
-      startDate: {
-        less_than_equal: selectedEndDate,
-      },
-    })
+  const filters = {
+    types: selectedTypes || null,
```
Comment (stevekuznetsov) [inline src/app/(frontend)/[center]/events/page.tsx]: I see, the comma-delimited thing ends up from query params? Why not parse it into an array here and let the server function work over structured data?

src/components/filters/ProvidersFilter.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+type ProvidersFilterProps = {
+  providers: Provider[]
+} & Partial<CheckboxFilterProps>
+
+export const ProvidersFilter = ({ providers, ...props }: ProvidersFilterProps) => {
+  const options = providers.map((provider) => ({
+    label: provider.name,
+    value: String(provider.id),
```
stevekuznetsov: Any chance we could use a slug instead so the user has any chance of understanding the URI?

src/actions/getCourses.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+export async function getCourses(params: GetCoursesParams): Promise<GetCoursesResult> {
+  const payload = await getPayload({ config })
+
+  try {
+    const { types, providers, states, affinityGroups, modesOfTravel, startDate, endDate } = params
+
+    const offset = params.offset || 0
+    const limit = params.limit || 10
```
rchlfryn: Should we make these defaults a const?

src/actions/getEventGroups.ts
```diff
@@ -0,0 +1,73 @@
...
+    const result = await payload.find({
+      collection: 'events',
+      limit: 10000,
```
rchlfryn: Same here... Let's just [disable pagination](https://payloadcms.com/docs/queries/pagination#disabling-pagination) here

src/utilities/queries/getEvents.ts
```diff
@@ -0,0 +1,124 @@
...
+  error?: string
+}
+
+export async function getEvents(params: GetEventsParams): Promise<GetEventsResult> {
+  const payload = await getPayload({ config })
+
+  try {
+    const { types, startDate, endDate, groups, tags, modesOfTravel, center } = params
+
+    const offset = params.offset || 0
```
rchlfryn, line 33: Let's make these default values consts and reuse them from above

src/utilities/queries/getPosts.ts
```diff
@@ -0,0 +1,84 @@
...
+  error?: string
+}
+
+export async function getPosts(params: GetPostsParams): Promise<GetPostsResult> {
+  const payload = await getPayload({ config })
+
+  try {
+    const { tags, sort, center } = params
+
+    const offset = params.offset || 0
```
rchlfryn, line 29: Same as above.. Let's make these defaults consts

src/actions/getProviders.ts
```diff
@@ -0,0 +1,35 @@
...
+  providers: { id: number; name: string }[]
+}
+
+export async function getProviders(): Promise<GetProvidersResult> {
+  try {
+    const payload = await getPayload({ config })
+
+    const result = await payload.find({
+      collection: 'providers',
+      limit: 1000,
```
rchlfryn: Do we want to just [disable pagination](https://payloadcms.com/docs/queries/pagination#disabling-pagination) here to avoid setting a high limit?

src/app/(frontend)/embeds/providers/page.tsx
```diff
@@ -0,0 +1,117 @@
...
+    where: {
+      _status: {
+        equals: 'published',
+      },
+    },
+    limit: 1000,
+    depth: 0,
+  })
+
+  const providers = result.docs as Provider[]
```
rchlfryn: does this work?

```suggestion
  const providers: Provider[] = result.docs
```

src/app/(frontend)/embeds/courses/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const hasActiveFilters = Boolean(
+    types || providers || states || affinityGroups || modesOfTravel || startDate || endDate,
+  )
+
+  return (
+    <FiltersTotalProvider initialTotal={total}>
+      <div
+        style={
+          backgroundColor && typeof backgroundColor === 'string' ? { backgroundColor } : undefined
+        }
```
rchlfryn: Can this be cleaned up?

```suggestion
        style={backgroundColor ? { backgroundColor } : undefined}
```

src/constants/quickDateFilters.ts
```diff
@@ -0,0 +1,74 @@
...
+      startDate: () => fmt(subYears(today(), 1)),
+      endDate: () => fmt(yesterday()),
+      sortDirection: 'desc',
+    },
+  ]
+}
+
+// Default exports for events and courses
+export const QUICK_DATE_FILTERS_EVENTS = createQuickDateFilters('Past Events')
+export const QUICK_DATE_FILTERS_COURSES = createQuickDateFilters('Past Courses')
```
rchlfryn: ⛏️ Can we just move these to their constant files to avoid assigning twice?

src/app/(frontend)/embeds/courses/courses-date-filter.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { DateRangeFilter } from '@/components/filters/DateRangeFilter'
+import { QUICK_DATE_FILTERS_COURSES } from '@/constants/quickDateFilters'
+
+type Props = {
+  startDate: string
+  endDate: string
+}
+
+export const CoursesDateFilter = ({ startDate, endDate }: Props) => {
+  return (
```
rchlfryn: Can we just assign the filters here rather than in the `quickDateFilters` file?

```suggestion
  const QUICK_DATE_FILTERS = createQuickDateFilters('Past Courses')
  return (
```

src/collections/Events/constants.ts
```diff
@@ -54,61 +44,4 @@ export const eventTypesData: EventType[] = [
...
-    {
-      id: 'past-events',
-      label: 'Past Events',
-      startDate: () => fmt(subYears(today(), 1)),
-      endDate: () => fmt(yesterday()),
-      sortDirection: 'desc',
-    },
-  ]
-})()
+export const QUICK_DATE_FILTERS = QUICK_DATE_FILTERS_EVENTS
```
rchlfryn: Can we just assign the filters here rather than in the `quickDateFilters` file? Or even just define it in `src/app/(frontend)/[center]/events/events-date-filter.tsx` similarly to how I suggested we assign this for courses, but I think it is used in other files.

```suggestion
export const QUICK_DATE_FILTERS = createQuickDateFilters('Past Events')
```

src/actions/getCoursesStates.ts
rchlfryn: Agreed with Steve.. we could use something like [findDistinct](https://payloadcms.com/docs/local-api/overview#collection-find-distinct) above:
```ts
  const result = await payload.findDistinct({
    collection: 'courses',
    field: 'location.state',
  })
```

which will return:
```ts
{
    "hasNextPage": false,
    "hasPrevPage": false,
    "limit": 0,
    "page": 1,
    "pagingCounter": 1,
    "totalDocs": 4,
    "totalPages": 1,
    "values": [
        {
            "location.state": "CA"
        },
        {
            "location.state": "CO"
        },
        {
            "location.state": "WA"
        },
        {
            "location.state": "WY"
        }
    ]
}
```

This one is weird because state is nested in location but it will work well with the other files.

src/actions/getEventGroups.ts
```diff
@@ -0,0 +1,73 @@
...
+export interface GetEventGroupsResults {
+  eventGroups: { label: string; value: string }[]
+}
+
+export async function getEventGroups(params: GetEventGroupsParams): Promise<GetEventGroupsResults> {
+  try {
+    const payload = await getPayload({ config })
+
+    // Query all events for the center to get unique event groups
+    const result = await payload.find({
```
rchlfryn: Same here - lets use `findDistinct` on `eventGroups`

src/actions/getEventTags.ts
```diff
@@ -0,0 +1,73 @@
...
+export interface GetEventTagsResults {
+  eventTags: { label: string; value: string }[]
+}
+
+export async function getEventTags(params: GetEventTagsParams): Promise<GetEventTagsResults> {
+  try {
+    const payload = await getPayload({ config })
+
+    // Query all events for the center to get unique event tags
+    const result = await payload.find({
```
rchlfryn: Same here - lets use `findDistinct` on `eventTags`

src/app/(frontend)/embeds/providers/page.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const states = Object.keys(providersByState).sort()
+
+  // Split states into two columns for vertical flow
+  const midpoint = Math.ceil(states.length / 2)
+  const leftColumnStates = states.slice(0, midpoint)
+  const rightColumnStates = states.slice(midpoint)
+
+  return (
+    <div style={backgroundColor ? { backgroundColor } : undefined}>
```
rchlfryn, line 70: Can we add a loading state here? It think the load time will reduce significantly with the query updates but it is not clear anything is loading your first time visiting the page with the embed.

src/app/(frontend)/embeds/courses/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const { providers: providersList } = showFilters ? await getProviders() : { providers: [] }
+  const { states: statesList } = showFilters ? await getCoursesStates() : { states: [] }
+
+  const hasActiveFilters = Boolean(
+    types || providers || states || affinityGroups || modesOfTravel || startDate || endDate,
+  )
+
+  return (
+    <FiltersTotalProvider initialTotal={total}>
```
rchlfryn, line 62: Can we add a loading state here? It think the load time will reduce significantly with the query updates but it is not clear anything is loading your first time visiting the page with the embed.

src/actions/getEventTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+export interface GetEventTypesResults {
+  eventTypes: { label: string; value: string; description?: string | null }[]
+}
+
+export async function getEventTypes(params: GetEventTypesParams): Promise<GetEventTypesResults> {
+  try {
+    const payload = await getPayload({ config })
+
+    // Query all events for the center to get unique event types
+    const result = await payload.find({
```
rchlfryn: Same here - lets use `findDistinct` on `types`

src/actions/getTags.ts
```diff
@@ -0,0 +1,57 @@
...
+  tags: Tag[]
+  error?: string
+}
+
+export async function getTags(center: string): Promise<GetTagsResult> {
+  const payload = await getPayload({ config })
+
+  try {
+    // Get all published posts for this center with tags populated
+    const posts = await payload.find({
```
rchlfryn: Same here - lets use `findDistinct` on `tags`

src/components/filters/DateRangeFilter.tsx
```diff
@@ -0,0 +1,174 @@
...
+      startDate: parseAsString.withDefault(''),
+      endDate: parseAsString.withDefault(''),
+    },
+    {
+      history: 'push',
+      shallow: false,
+    },
+  )
+
+  const [filterType, setFilterType] = useState('')
```
rchlfryn: ⛏️ I know I wrote this, but `filterType` doesn't make sense to me. Can we change it to `selectedFilter`, or something else?

src/components/filters/ProvidersFilter.tsx
rchlfryn: Do you mean slug instead of id? 

But this makes me think.. do we want to avoid what we did with tenant and make the slug for providers the id? Probably a follow up PR

src/components/EventsList.tsx
rchlfryn, line 1: Was this supposed to replace `EventCollection`? If so, we should remove the file

src/components/PostsList.tsx
rchlfryn, line 1: Was this supposed to replace `PostsCollection`? If so, we should remove the file

src/fields/affinityGroupField.ts
```diff
@@ -0,0 +1,17 @@
+import { Field } from 'payload'
+
+export const affinityGroupOptions = [
+  { label: 'LGBTQ+', value: 'lgbtq' },
+  { label: "Women's Specific", value: 'womens-specific' },
+  { label: 'Youth Specific', value: 'youth-specific' },
```
rchlfryn, line 6: Very picky on for the naming of this, but I want us to get this right. Can we not use specific? For me it, it emphasizes exclusion over inclusion. I also think `LGBTQ+` should have the same label. Maybe just simplify it??:
- For LGBTQ+, For Women, For Youth
- LGBTQ+, Women, Youth

src/components/CoursePreviewSmallRow.tsx
```diff
@@ -0,0 +1,152 @@
...
+    courseType,
+    subtitle,
+    description,
+    registrationDeadline,
+    timezone,
+    provider,
+  } = doc
+
+  // Format dates using the same logic as EventPreviewSmallRow and EventInfo
+  const getDateDisplay = () => {
```
rchlfryn: Should we make this a utility function since it is used in events?

src/components/CoursePreviewSmallRow.tsx
```diff
@@ -0,0 +1,152 @@
...
+    } else if (endDate) {
+      return `${formatDateTime(startDate, timezone, 'MMM d, yyyy')} - ${formatDateTime(endDate, timezone, 'MMM d, yyyy')}`
+    } else {
+      return formatDateTime(startDate, timezone, 'MMM d, yyyy')
+    }
+  }
+
+  const dateDisplay = getDateDisplay()
+
+  const formatRegistrationDeadline = (dateString: string) => {
```
rchlfryn: Should we make this a utility function since it is used in events?

src/app/(frontend)/embeds/courses/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <div
+        style={
+          backgroundColor && typeof backgroundColor === 'string' ? { backgroundColor } : undefined
+        }
+      >
+        {title && (
```
rchlfryn, line 64: Can we use [getTextColorFromBgColor.ts](https://github.com/NWACus/web/blob/main/src/utilities/getTextColorFromBgColor.ts)?

src/app/(frontend)/embeds/providers/page.tsx
```diff
@@ -0,0 +1,117 @@
...
+      {title && (
```
rchlfryn, line 71: Same here.. can we use `getTextColorFromBgColor`?

src/app/(frontend)/embeds/courses/page.tsx
rchlfryn: Any reason we are using `style` here or can we use tailwind classes?

src/app/(frontend)/embeds/providers/page.tsx
rchlfryn: Any reason we are using style here or can we use tailwind classes?

src/components/filters/CheckboxFilter.tsx
rchlfryn, line 1: Styling for the embeds is a little strange with the narrow accordion text labels, it makes the font look smaller than the check box labels. Not sure what to do about this. Maybe make the `a3 h3` size bigger? or all title sizes for that matter since it is a condensed font

Image

src/app/(frontend)/embeds/courses/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+              hasActiveFilters={hasActiveFilters}
+              startDate={startDate || ''}
+              endDate={endDate || ''}
+            />
+          </div>
+        )}
+
+        <div className="flex gap-6">
+          <div className="flex-1">
+            <CoursesList initialCourses={courses} initialHasMore={hasMore} initialError={error} />
```
rchlfryn, line 84: ⛏️ I am wondering if we want to restyle this at all. there is a lot of dead space

Image

src/components/CoursePreviewSmallRow.tsx
```diff
@@ -0,0 +1,152 @@
...
+      <div className="flex items-start justify-between gap-2">
+        <div className="flex flex-col gap-1 flex-1">
+          {courseTypeDisplay && (
+            <div className="text-xs text-muted-foreground">{courseTypeDisplay}</div>
+          )}
+          <h3 className={cn('text-lg leading-tight', hasExternalUrl && 'group-hover:underline')}>
+            {title}
+          </h3>
+          {providerName && (
+            <p className="text-sm font-medium text-muted-foreground">{providerName}</p>
```
rchlfryn: It might be nice to add `Provided by:` here.. 
```suggestion
            <p className="text-sm font-medium text-muted-foreground">Provided by: {providerName}</p>
```

src/components/CoursesList.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  return (
+    <div className="space-y-4">
+      {courses.map((course) => (
+        <div key={course.id} className="border-b pb-4 last:border-b-0">
+          <CoursePreviewSmallRow doc={course} />
+        </div>
+      ))}
+
+      {hasMoreData && (
```
rchlfryn, line 190: Do you have an example of what having this Loader looks like?

src/app/(frontend)/colors.css
```diff
@@ -165,3 +165,34 @@
...
+.a3 {
+  /* A3 BRAND COLORS */
+  --mountain-slate: hsl(220 25% 20%); /* RGB: 38 / 46 / 63 */
+  --couloir-blue: hsl(200 81% 95%); /* RGB: 233 / 247 / 254 */
+  --patrol-red: hsl(358 85% 52%); /* RGB: 237 / 28 / 36 */
+  /*--------------------------*/
+
+  --primary: var(--mountain-slate);
+  --primary-hover: hsla(220 25% 20% / 0.8);
+  --primary-foreground: hsl(0 0% 98%);
```
rchlfryn, line 178: I am curious why we stuck with hsl values rather than just using `white` or `black` in this file. Ah well...

src/app/(frontend)/[center]/events/events-date-filter.tsx
Comment (rchlfryn) [inline src/app/(frontend)/[center]/events/events-date-filter.tsx:1]: Should we nest filter files in a filter parent file? Same for courses..

src/app/(frontend)/[center]/events/events-mobile-filters.tsx
```diff
@@ -1,102 +1,31 @@
...
-          </Button>
-        </div>
-      </div>
-    </>
+    <MobileFiltersDrawer docLabel="events" docCount={total} hasActiveFilters={hasActiveFilters}>
+      <EventsTypeFilter types={types} />
+      <EventsGroupsFilter groups={groups} />
+      <EventsTagsFilter tags={tags} />
+      <ModesOfTravelFilter />
+      <EventsDatePicker startDate="" endDate="" />
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/events/events-mobile-filters.tsx]: Let's add a check on these like we have in `blog-mobile-filter`.

```ts
{item.length > 1 && <Filter item={item} />}
```
Probably should put it in the Files themselves since they are reused

src/app/(frontend)/embeds/courses/courses-mobile-filters.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { total } = useFiltersTotalContext()
+
+  return (
+    <MobileFiltersDrawer docLabel="courses" docCount={total} hasActiveFilters={hasActiveFilters}>
+      <CoursesDateFilter startDate={startDate} endDate={endDate} />
+      <CoursesTypeFilter />
+      <ProvidersFilter providers={providers} />
+      <StatesFilter states={states} />
+      <AffinityGroupsFilter />
+      <ModesOfTravelFilter />
```
rchlfryn: Let's add a check on these like we have in `blog-mobile-filter`.

```ts
{item.length > 1 && <Filter item={item} />}
```
Probably should put it in the Files themselves since they are reused

src/app/(frontend)/embeds/providers/page.tsx
```diff
@@ -0,0 +1,117 @@
...
+        <div className="mb-6">
+          <h1 className="text-2xl font-bold mb-2">{title}</h1>
+        </div>
+      )}
+      <div className="grid sm:grid-cols-2 gap-x-4">
+        <StatesAccordion states={leftColumnStates} providersByState={providersByState} />
+        <StatesAccordion states={rightColumnStates} providersByState={providersByState} />
```
rchlfryn, line 78: Bummer we can't clean this up by using [`CSS columns`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/columns)

src/app/(frontend)/embeds/providers/page.tsx
```diff
@@ -0,0 +1,117 @@
...
+      })
+    }
+  })
+
+  const states = Object.keys(providersByState).sort()
+
+  // Split states into two columns for vertical flow
+  const midpoint = Math.ceil(states.length / 2)
+  const leftColumnStates = states.slice(0, midpoint)
+  const rightColumnStates = states.slice(midpoint)
```
rchlfryn: Reminder to remove this based on https://github.com/NWACus/web/pull/675/files#r2530430838

src/app/(frontend)/embeds/courses/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+        }
+      >
+        {title && (
+          <div className="mb-6">
+            <h1 className="text-2xl font-bold mb-2">{title}</h1>
+          </div>
+        )}
+
+        {showFilters && (
+          <div className="md:hidden mb-4">
```
rchlfryn, line 71: I wonder if this should be sticky? probably okay to leave it for v1

src/components/ProviderPreview.tsx
rchlfryn, line 1: The styling of the pop up is quite different than what is on A3s site. Did we intentionally leave off the Name, description, other states? Maybe we add a `width` & `min-height` to help?

Image

rchlfryn (changes requested):

src/utilities/queries/getCourses.ts
busbyk: No, if there are no docs matching the where conditions you would get a return object like:

```
{
  courses: 0,
  hasMore: false,
  total: 0
}
```

src/constants/quickDateFilters.ts
busbyk: I'll just move these to where they're used.

src/app/(frontend)/embeds/courses/courses-date-filter.tsx
busbyk: Yup!

src/actions/getCourses.ts
busbyk: Sure

src/utilities/queries/getEvents.ts
busbyk: 👍 - but leaving offset hardcoded because it should always be zero by default.

src/app/(frontend)/[center]/events/page.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/events/page.tsx]: It is getting parsed into an array by `nuqs`: https://github.com/NWACus/web/blob/a3-course-provider-embeds/src/app/(frontend)/%5Bcenter%5D/events/page.tsx#L21

src/app/(frontend)/embeds/courses/page.tsx
busbyk: Yep, can remove the typeof check now. This was leftover from before I was using `nuqs` to make the query params type safe - so now we don't need this type check. 

`style` is necessary here as opposed to tailwind classes because a user can pass any color - really any string - to the backgroundColor query param and you can't use arbitrary values that are dynamic: https://v3.tailwindcss.com/docs/content-configuration#dynamic-class-names

src/app/(frontend)/embeds/courses/page.tsx
busbyk: No, because the backgroundColor is not a class name - it's any [background-color](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/background-color) css property.

src/app/(frontend)/embeds/providers/page.tsx
busbyk: See https://github.com/NWACus/web/pull/675#discussion_r2530296440

src/app/(frontend)/embeds/providers/page.tsx
busbyk: See https://github.com/NWACus/web/pull/675#discussion_r2530304759 RE style vs. tailwind classes.

src/app/(frontend)/embeds/providers/page.tsx
busbyk: Uh, why can't we? 

Personally, I don't think this isn't a clean solution 😅 Just three lines of code to split the array and abstracting some shared code into a component. 

But if you really don't like it or think it's not performant or something, we could use a .css file or a style object to use css columns...

src/fields/affinityGroupField.ts
busbyk: All good. Those were the values that @rustynwac told me to use. I like the "For ___" variant. Rusty, do you agree?

src/app/(frontend)/colors.css
busbyk: Can you explain what you mean? You're just saying that using hsl values for white is overkill? I think it's 1. just sticking with the pattern and 2. I believe we need to use hsl or rgba for [opacity modifiers to work](https://v3.tailwindcss.com/docs/customizing-colors#using-css-variables)

src/app/(frontend)/embeds/providers/page.tsx
busbyk: Oh we don't need to explicitly type this at all - just slop from Claude.

src/app/(frontend)/embeds/providers/page.tsx
busbyk: Do you mean a [loading state in a Suspense boundary](https://nextjs.org/docs/app/api-reference/file-conventions/loading)? This is a server component so there's no client side data fetching happening here. 

I agree a loading skeleton in a Suspense boundary would be nice but it would add to this PR and you already felt like it was too big. I can add that in a separate PR.

src/components/CoursePreviewSmallRow.tsx
busbyk: I think it's fine to just use the same `{formatDateTime(event.registrationDeadline, event.timezone, 'MMM d, yyyy')}`

src/components/ProviderPreview.tsx
busbyk: Since this page is embedded in an iframe we don't have the same capability to use a full page modal like the existing providers pop up on avalanche.org. We need to be careful that whatever element we use, it does not exceed the bounds of the iframe (hence the little bit of vertical padding I use on the /embeds/providers page). So I went for a very concise set of info. But we should include description. 

I am very open to suggestions on design here though. What are you thinking? I considered another accordion for each provider (since the future implementation of iframe-resizer I'm working on will resize the embed accordingly) but then we've got two accordions nested inside of each other which doesn't feel right. 

We could include a modal and associated script with the <iframe> when we provide the embed code that would listen for "open modal" and "close modal" messages from our iframe but that felt like more work than it was worth.

src/components/filters/CheckboxFilter.tsx
busbyk: Updated to allow passing titleClassName in a few places to enable:

Image

src/app/(frontend)/embeds/courses/courses-mobile-filters.tsx
busbyk: The CheckboxFilter at least has a no options fallback UI: https://github.com/NWACus/web/blob/a3-course-provider-embeds/src/components/filters/CheckboxFilter.tsx#L141

But you think it makes more sense to hide the filters completely?

src/app/(frontend)/embeds/courses/page.tsx
busbyk: Same comment as https://github.com/NWACus/web/pull/675/files#r2531172475 -- will add a loading skeleton in a separate PR.

src/app/(frontend)/embeds/courses/page.tsx
busbyk: Hmm yea I was expecting this to be embedded in narrower widths but it won't be on avalanche.org at the least so I agree. What about multiple columns of courses? We don't have images to take up space so that might be our best option. Or we could add a max width.

src/components/CoursesList.tsx
busbyk: Yea: 
CleanShot 2025-11-15 at 22 57 49

Lower the getCourses limit and throttle your network to see it.

src/components/filters/DateRangeFilter.tsx
busbyk: How about `quickFilter`

src/app/(frontend)/embeds/courses/page.tsx
rchlfryn: Make sense. I forgot the context of a user was passing in the params via string.

src/app/(frontend)/embeds/providers/page.tsx
rchlfryn: I was talking about adding a loading skeleton. When I first visited the page I was confused because it was just a white page with titles, and stayed that way for a few seconds.

src/app/(frontend)/embeds/courses/page.tsx
rchlfryn: My originally comment made more sense, but my comment was about the title text color. It would be nice to be able to change that with a string since we give the user the option to pass in a backgoround color. We can do it automatically like [getTextColorFromBgColor.ts](https://github.com/NWACus/web/blob/main/src/utilities/getTextColorFromBgColor.ts) or add a param similarly to background color

src/app/(frontend)/embeds/providers/page.tsx
rchlfryn: I was just calling out there are still pain points between css and js styling. 

I think you came up with the right solution here. I used CSS columns for event info and it worked well, but wont work here with the Accordion toggle interaction. See example with CSS below

Kapture 2025-11-16 at 14 34 01

Just a review comment.. no changes needed.

src/app/(frontend)/embeds/courses/page.tsx
rchlfryn: Maybe we wait for feedback on this one because if everyone uses it on smaller widths, we wont need to update anything.

src/app/(frontend)/embeds/courses/courses-mobile-filters.tsx
rchlfryn: I think it would be cleaner to hide the filer completely if there are no options. Just thinking of UX

src/components/ProviderPreview.tsx
rchlfryn: I think using the popover is fine. I was suggesting to make it look more consistent regardless of what data is entered for v1. I like the idea of an extra accordion. I did something similar on [EventTable](https://github.com/NWACus/web/blob/ac80dd1b3ae0cbec78ed3517577fd989a20bd039/src/components/EventsTable/index.tsx#L267) without using accordion, but I have no idea how that will work in an iframe.

I messed around with it a bit. Take or leave the things you like.

Screenshot 2025-11-16 at 14 57 31
Screenshot 2025-11-16 at 14 57 37
Screenshot 2025-11-16 at 14 57 47

```ts
'use client'

import Link from 'next/link'

import type { Provider } from '@/payload-types'

import { courseTypesData } from '@/constants/courseTypes'
import { isValidFullUrl } from '@/utilities/validateUrl'
import { ExternalLink, Mail, MapPin, Phone } from 'lucide-react'
import { Popover, PopoverContent, PopoverTrigger } from './ui/popover'

export const ProviderPreview = (props: { doc?: Provider }) => {
  const { doc } = props

  if (!doc) return null

  const { name, location, website, email, phone, courseTypes, details } = doc

  const getLocationText = () => {
    if (!location) return null
    if (location.city && location.state) return `${location.city}, ${location.state}`
    if (location.city) return location.city
    if (location.state) return location.state
    return 'Location'
  }

  const locationText = getLocationText()

  const courseTypeLabels = courseTypes
    ?.map((ct) => courseTypesData.find((ctd) => ctd.value === ct)?.label)
    .filter(Boolean)

  const validWebsite = website && isValidFullUrl(website) ? website : null

  return (
    <Popover>
      <PopoverTrigger className="cursor-help hover:underline text-base w-fit">
        {name}
      </PopoverTrigger>
      <PopoverContent
        className="flex flex-col gap-1 py-2 w-[80vw] max-w-[500px]"
        align="start"
        sideOffset={8}
      >
        {courseTypeLabels && courseTypeLabels.length > 0 && (
          <p>
            <b>Course offered:</b> {courseTypeLabels.join(', ')}
          </p>
        )}
        {details && <p className="hidden md:block text-sm">{details}</p>}
        <div className="sm:columns-2">
          {locationText && (
            <div className="flex items-center gap-1 text-sm">
              <MapPin className="h-3.5 w-3.5 flex-shrink-0" />
              <span>{locationText}</span>
            </div>
          )}
          {validWebsite && (
            <Link
              href={validWebsite}
              target="_blank"
              rel="noopener noreferrer"
              className="flex items-center gap-1 text-sm text-blue-600 hover:text-blue-800 hover:underline"
            >
              <ExternalLink className="h-3.5 w-3.5 flex-shrink-0" />
              <span>Website</span>
            </Link>
          )}
          {email && (
            <a href={`mailto:${email}`} className="flex items-center gap-1 text-sm">
              <Mail className="h-3.5 w-3.5 flex-shrink-0" />
              <span>{email}</span>
            </a>
          )}
          {phone && (
            <a href={`tel:${phone}`} className="flex items-center gap-1 text-sm">
              <Phone className="h-3.5 w-3.5 flex-shrink-0" />
              <span>{phone}</span>
            </a>
          )}
        </div>
      </PopoverContent>
    </Popover>
  )
}
```

src/app/(frontend)/embeds/courses/courses-mobile-filters.tsx
busbyk: Cool, that's how it works: https://github.com/NWACus/web/blob/a3-course-provider-embeds/src/components/filters/CheckboxFilter.tsx#L83

src/app/(frontend)/embeds/courses/page.tsx
busbyk: Oh I thought you meant literally using [https://github.com/NWACus/web/blob/main/src/utilities/getTextColorFromBgColor.ts](getTextColorFromBgColor). 

But yea I realized the need for this in my follow up PR as I was testing out embedding on other sites - I'm making that change in my next PR for these features: https://github.com/NWACus/web/commit/40c89223545dd64bda5a1dc8ad1fd658d37a6b8a#diff-b6cf5b98f0306fe6fb11683f53604e0611552219bced19b4cf2818dabd1d27f4

src/app/(frontend)/embeds/courses/page.tsx
busbyk: Well we'll definitely need to embed this full width on avalanche.org and a3's website so I'm down to try out a two column layout

src/components/ProviderPreview.tsx
busbyk: I like adding `details` in here for sure. We can't use an element that doesn't contribute to the height of the document if we're going to do that though or we'll run into this with the providers towards the bottom of the iframe: 
CleanShot 2025-11-17 at 10 25 24

Using additional accordions is probably our best option. I feel like the provider accordions should be closed when the state accordion is closed though, hmm.

src/app/(frontend)/[center]/layout.tsx
```diff
@@ -58,7 +60,9 @@ export default async function RootLayout({ children, params }: Args) {
     },
   })
   const tenant = tenantsRes.docs.length >= 1 ? tenantsRes.docs[0] : null
-  invariant(tenant, `Could not determine tenant for center value: ${center}`)
+  // invariant(tenant, `Could not determine tenant for center value: ${center}`)
```
Comment (rchlfryn) [inline src/app/(frontend)/[center]/layout.tsx]: Remove this?
```suggestion
```

src/components/ProviderPreview.tsx
busbyk: Yanno, a Dialog with a transparent background actually doesn't look too bad, and the iframe height will be much larger in production with many providers. So I think this could work:

CleanShot 2025-11-17 at 11 23 56

src/fields/affinityGroupField.ts
rustynwac: Lets go with LGBTQ+, Women, Youth

src/components/ProviderPreview.tsx
```diff
@@ -0,0 +1,80 @@
...
+    ?.map((ct) => courseTypesData.find((ctd) => ctd.value === ct)?.label)
+    .filter(Boolean)
+
+  const validWebsite = website && isValidFullUrl(website) ? website : null
+
+  const contactItems = [locationText, validWebsite, email, phone].filter(Boolean)
+
+  return (
+    <Dialog>
+      <DialogTrigger className="cursor-help hover:underline text-base w-fit">{name}</DialogTrigger>
```
rchlfryn, line 40: Looks like we are missing `<DialogTitle>` here
Screenshot 2025-11-17 at 12 29 18

src/components/ProviderPreview.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <DialogContent className="flex flex-col gap-2 py-2" overlayClassName="bg-transparent">
+        {courseTypeLabels && courseTypeLabels.length > 0 && (
+          <p>Offers: {courseTypeLabels.join(', ')}</p>
```
rchlfryn: ⛏️ If you're in here and want to wrap `Offers:` in a different font weight, that would rock.

rchlfryn (approved):
Did a _light_ re-review but looks good pending a few clean up changes

src/components/ProviderPreview.tsx
busbyk: Yea sure, used the same font-semibold as EventInfo

src/components/ProviderPreview.tsx
busbyk: Nice, good call.

src/app/(frontend)/[center]/layout.tsx
Comment (busbyk) [reply src/app/(frontend)/[center]/layout.tsx]: Woops, yea didn't mean to leave this in. I would like to change this at some point tho - we get a lot of unnecessary logging for this since invalid paths match the [center] dynamic path part (like /.well-known for example).

PR #689, opened 2025-11-19
Migration snapshot diff script

busbyk:
## Description

The suggested one-liner in `docs/migration-safety.md` was OK but this script gives us a much more concise comparison of the two different snapshots. 

I also felt like the migration docs were a little messy so cleaned those up.

### DIFF
--- a/docs/migration-safety.md
+++ b/docs/migration-safety.md
@@ -15,17 +15,15 @@ Background reading:
 - https://github.com/ariga/atlas/issues/3317
 - https://github.com/launchbadge/sqlx/issues/2085#issuecomment-1237474188
 
-## Automated Checks
+## Static Analysis Check
 
-### Static Analysis
+Runs on `git commit` via Husky pre-commit hook.
 
-**What it does:** Scans migration files for dangerous patterns.
+Runs in CI: PRs with new or modified migrations:
+- The `migration-safety` CI job runs automatically
+- If issues are detected, a comment is posted on the PR with details
 
-**When it runs:**
-- Automatically on `git commit` via Husky pre-commit hook
-- In CI when migrations are added or modified in pull requests
-
-**Manual usage:**
+Run locally:
 ```bash
 # Check all migrations
 pnpm migrate:check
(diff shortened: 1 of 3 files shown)


rchlfryn (approved):
Love the new diff. Thanks for cleaning this up

docs/migration-safety.md
```diff
@@ -74,45 +73,21 @@ To manually test the effects of a migration locally, follow these steps:
...
-## Writing Custom Migrations to Replace Unsafe Auto-Generated Ones
+#### 1 - Preferred: Write a custom migration
 
-Sometimes Payload/Drizzle generates migrations that would cause data loss in production due to the `PRAGMA foreign_keys=OFF` issue. In these cases, you need to:
 1. Manually simplify the migration to avoid the unsafe patterns
 2. Preserve the JSON schema snapshot so Payload's diffing logic recognizes the schema is in sync
 
+#### 2 - Keep old fields, marked as hidden
+
+If you can't write a custom migration that
```
stevekuznetsov, line 89: :D

src/scripts/analyze-migration-diff.ts
```diff
@@ -0,0 +1,219 @@
...
+function diffTableFields(oldTable: TableSchema, newTable: TableSchema): TableFieldDiff {
+  const oldColumns = new Set(Object.keys(oldTable.columns))
+  const newColumns = new Set(Object.keys(newTable.columns))
+
+  const fieldsAdded: string[] = []
+  const fieldsRemoved: string[] = []
+  const fieldsModified: string[] = []
+
+  // Find added columns
+  for (const column of newColumns) {
```
stevekuznetsov, line 52: You have two `Set()` - you could just use symmetric difference methods for these

src/scripts/analyze-migration-diff.ts
```diff
@@ -0,0 +1,219 @@
...
+      }
+      if (diff.fieldsModified.length > 0) {
+        output += `     ~ Fields modified: ${diff.fieldsModified.join(', ')}\n`
+      }
+    })
+    output += '\n'
+  }
+
+  if (tablesAdded.length === 0 && tablesRemoved.length === 0 && tablesModified.size === 0) {
+    output += '📭 No table-level changes detected\n\n'
```
stevekuznetsov, line 179: What does the mailbox mean?

stevekuznetsov (approved):

src/scripts/analyze-migration-diff.ts
busbyk: haha great question. I need to tell Claude to stop using so many emojis.

PR #694, opened 2025-11-20
ESLint rule to warn on type assertion

busbyk:
## Description

Adds `@typescript-eslint/consistent-type-assertions` with `assertionStyle: never` to warn on usage of type assertions like `const variable = value as Type` and `const variable = <Type>value`. 

I fixed or decided it was safe to ignore many instances of these but I time-boxed myself on this and ran out of time. Anyone else, please feel free to jump in here and fix or decide to ignore any of the remaining usages:

```
./src/app/(frontend)/[center]/next/preview/route.ts
18:22  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
46:14  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
46:14  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/collections/Pages/endpoints/duplicatePageToTenant.ts
7:28  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
64:12  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
67:12  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/collections/Sponsors/index.ts
13:23  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/collections/Users/components/InviteUserDrawer.tsx
66:27  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/components/Header/utils.ts
73:20  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
91:27  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/components/TenantSelector/TenantSelector.client.tsx
35:25  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
61:16  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/endpoints/seed/index.ts
481:32  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
486:37  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
491:32  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/endpoints/seed/upsert.ts
62:17  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
184:17  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/fields/defaultLexical.ts
49:25  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
50:22  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/fields/slug/SlugComponent.tsx
28:12  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/middleware.ts
42:22  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/services/vercel.ts
62:22  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/utilities/findDocumentsWithBlockReferences.ts
16:18  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
31:16  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
40:15  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/utilities/generateOGImage.tsx
93:35  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/utilities/removeNonDeterministicKeys.ts
9:22  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
14:7  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions

./src/utilities/useClickableCard.ts
37:24  Warning: Do not use any type assertions.  @typescript-eslint/consistent-type-assertions
```

### DIFF
--- a/src/collections/Pages/components/DuplicatePageFor/index.tsx
+++ b/src/collections/Pages/components/DuplicatePageFor/index.tsx
@@ -1,18 +1,23 @@
 import { byGlobalRole } from '@/access/byGlobalRole'
-import { User } from '@/payload-types'
 import { roleAssignmentsForUser } from '@/utilities/rbac/roleAssignmentsForUser'
 import { ruleMatches } from '@/utilities/rbac/ruleMatches'
 import { EditMenuItemsServerProps, PayloadRequest } from 'payload'
 import { DuplicatePageForDrawer } from './DuplicatePageForDrawer'
 
 export const DuplicatePageFor = ({ user, payload }: EditMenuItemsServerProps) => {
+  if (!user) {
+    return null
+  }
+
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   const mockedPayloadReq = {
     user,
     payload,
   } as PayloadRequest
+
   const isSuperAdmin = byGlobalRole('*', '*')({ req: mockedPayloadReq })
 
-  const roleAssignments = roleAssignmentsForUser(payload.logger, user as User)
+  const roleAssignments = roleAssignmentsForUser(payload.logger, user)
   const matchingTenantIds = roleAssignments
     .filter(
       (assignment) =>
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -36,6 +36,12 @@ const eslintConfig = [
           caughtErrorsIgnorePattern: '^(_|ignore)',
         },
       ],
+      '@typescript-eslint/consistent-type-assertions': [
+        'warn',
+        {
+          assertionStyle: 'never',
+        },
+      ],
     },
   },
   globalIgnores(['.next/', 'src/migrations/']),
--- a/src/app/(frontend)/[center]/[...segments]/page.tsx
+++ b/src/app/(frontend)/[center]/[...segments]/page.tsx
@@ -106,6 +106,7 @@ export async function generateMetadata(
   },
   parent: Promise<ResolvedMetadata>,
 ): Promise<Metadata> {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   const parentMeta = (await parent) as Metadata
   const { center, segments } = await paramsPromise
 
--- a/src/app/(frontend)/[center]/[slug]/page.tsx
+++ b/src/app/(frontend)/[center]/[slug]/page.tsx
@@ -107,6 +107,7 @@ export async function generateMetadata(
   },
   parent: Promise<ResolvedMetadata>,
 ): Promise<Metadata> {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   const parentMeta = (await parent) as Metadata
   const { center, slug } = await paramsPromise
 
--- a/src/app/(frontend)/[center]/blog/[slug]/page.tsx
+++ b/src/app/(frontend)/[center]/blog/[slug]/page.tsx
@@ -107,6 +107,7 @@ export async function generateMetadata(
   { params: paramsPromise }: Args,
   parent: Promise<ResolvedMetadata>,
 ): Promise<Metadata> {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   const parentMeta = (await parent) as Metadata
   const { center, slug = '' } = await paramsPromise
   const post = await queryPostBySlug({ center: center, slug: slug })
--- a/src/app/(frontend)/[center]/blog/page.tsx
+++ b/src/app/(frontend)/[center]/blog/page.tsx
@@ -97,6 +97,7 @@ export async function generateMetadata(
   _props: Args,
   parent: Promise<ResolvedMetadata>,
 ): Promise<Metadata> {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
   const parentMeta = (await parent) as Metadata
 
   const parentTitle =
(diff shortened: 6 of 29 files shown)


rchlfryn (approved):
wonderful! did we want to update the tests as well?

busbyk:
Ope, sorry this was meant to be a draft! I don't think we should merge this until we address all of the existing warnings - there are still several unaddressed in the PR description. 

> wonderful! did we want to update the tests as well?

Update the tests?

stevekuznetsov:
> decided it was safe to ignore

What criteria did you use to determine this?

src/collections/Pages/components/DuplicatePageFor/index.tsx
```diff
@@ -1,18 +1,23 @@
...
 import { ruleMatches } from '@/utilities/rbac/ruleMatches'
 import { EditMenuItemsServerProps, PayloadRequest } from 'payload'
 import { DuplicatePageForDrawer } from './DuplicatePageForDrawer'
 
 export const DuplicatePageFor = ({ user, payload }: EditMenuItemsServerProps) => {
+  if (!user) {
+    return null
+  }
+
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
stevekuznetsov, line 12: eg here - why did we need the `as`? if it's to omit otherwise required fields, how do we know that all downstream users don't need any of those? if they don't, wouldn't the real fix be to make the methods we pass this into require a smaller subset of this type?

busbyk:
> > decided it was safe to ignore
> 
> What criteria did you use to determine this?

Primarily, "Is it unreasonable to make typescript happy using a different approach and is it reasonably safe to use a type assertion here"

With a little bit of, "This is a time-boxed tech debt task, don't do unnecessary work that provides no value to the client" sprinkled on top.

src/collections/Pages/components/DuplicatePageFor/index.tsx
busbyk: We have two options here:
1. We could use this type assertion approach when creating a mocked PayloadRequest object when using the `by___` access functions outside of Payload configs.
2. We could change the type of our `by___` access functions to accept an object of user and payload. Which I suppose would be fine - we'd update all usages of our access functions to be wrapped in a function that extracts user and payload from the request object and pass those the access functions directly.

stevekuznetsov: > we'd update all usages of our access functions to be wrapped in a function that extracts user and payload from the request object and pass those the access functions directly.

Would we need to do this? TS should have stuctural typing, I thought passing a superset would be fine?

stevekuznetsov:
Yeah, I guess I'm making the (admittedly pedantic) point that we had implicitly made the same determination when writing `as` in the past, so if we're going to turn on the lint, the hope is that we'd be no longer doing that - unless we're in the habit of using the `as` chainsaw for surgery without justification normally?

busbyk:
> Yeah, I guess I'm making the (admittedly pedantic) point that we had implicitly made the same determination when writing `as` in the past, so if we're going to turn on the lint, the hope is that we'd be no longer doing that - unless we're in the habit of using the `as` chainsaw for surgery without justification normally?

I'm turning on the lint because I frequently get PR feedback requesting the removal of `as` and I would like to be warned about it so I don't have to get that feedback.

src/collections/Pages/components/DuplicatePageFor/index.tsx
stevekuznetsov: See this [playground](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgM4FcAO1UTMgbwChlTkZgIAbAEwEYAuNMKUAcwG4SyLqaAmJgCMA9iKoQ4ILgF8iRUJFiIUGIbnzEy5SrUHJR4ydKJyiMdCARhgIkMgR2MAWwgAKVEzUaAlIW6kjiCoRgB0VCJsHqG8ej6m8kGo+GAAFuxeWDh4yAC8hLH0TABEAO6pInDFADQ6fPrwVLgyXEnorm5p7D5AA)

stevekuznetsov:
Makes sense. Even one `as` makes the consumers and producers unreliable, which is why I assume we ended up with a Payload form hook returning a value incorrectly typed as always being non-null or whatever. Sometimes you need the chainsaw to bash some other library into submission but ideally we just have none of them -.-

busbyk:
> Makes sense. Even one `as` makes the consumers and producers unreliable, which is why I assume we ended up with a Payload form hook returning a value incorrectly typed as always being non-null or whatever. Sometimes you need the chainsaw to bash some other library into submission but ideally we just have none of them -.-

From the description:
> but I time-boxed myself on this and ran out of time. Anyone else, please feel free to jump in here and fix

@stevekuznetsov get in there brother

stevekuznetsov:
Sure - the pattern I have used in the past that helps here is:

1. add the lint rule
2. mark every existing instance of failure with an ignore comment (I hope there's automation for it, maybe something like [this](https://stackoverflow.com/a/64017715) would work)
3. add a `pnpm <>` command to find all ignore comments in our repo for this rule, sort them, and write them to a file
4. add a CI task to ensure that the file is up-to-date on merge by running 2 and doing a `git diff`
5. do not approve PRs that add to this list

That way you get to add the lint rule today, stop any new usages of `as`, and, crucially, you can punt on fixing existing ones, so there's no rush to figure out more durable solutions than squinting at an issue hard enough to see if it's "reasonable" to have the `as` in perpetuity.

PR #698, opened 2025-11-21
Bugfix: blank screen of death on block reorder when content block present

### LINKED ISSUE #683: Issues removing blocks when page building
**Describe the bug**
I'm trying to work on a few periphery pages (About, Who We Are, Beacon Parks, etc...) to familiarize myself with the new website formatting. I'm running into a black screen every time I try to remove a content block... This only started going to a black screen in the last half hour.

**To Reproduce**
Go to the Beacon Parks pages on SNFAC.
Remove a block
See error

**Expected behavior**
No black screens of death

**Screenshots**
[Vids](https://drive.google.com/file/d/18SQllBJ4hyiCtTD2_W3SLvXGY_3zfWUB/view?usp=sharing)

**Environment details**
MacBook Air (OS 15.6.1) using Chrome (v142.0.7444.162)

busbyk:
## Description

When a content block was present in a blocks type field, if you re-ordered a block the `src/components/ColumnLayoutPicker/index.tsx` component would throw an uncaught error. It seems that the fields returned from `useFormFields` would be empty for at least one render during this re-order and we didn't handle that case. 

I updated the way we access the sibling field to match `src/blocks/Content/components/DefaultColumnAdder.tsx`.

## Related Issues

Fixes #683

### DIFF
--- a/src/components/ColumnLayoutPicker/index.tsx
+++ b/src/components/ColumnLayoutPicker/index.tsx
@@ -1,7 +1,7 @@
 'use client'
 
 import { ToggleGroup, ToggleGroupItem } from '@/components/ui/toggle-group'
-import { FieldLabel, useField, useFormFields } from '@payloadcms/ui'
+import { FieldLabel, useField } from '@payloadcms/ui'
 import { Columns2, Columns3, Columns4, Square } from 'lucide-react'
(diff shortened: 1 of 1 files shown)


rchlfryn (approved):
LG

PR #703, opened 2025-11-22
Events - fix past dates shown on `/events` & change `Field Class` label

rchlfryn:
## Related Issues
I was trying to get ready for events demo videos and needed to update

## Key Changes
- Since the change with #675 , nuqs was not setting `startDate` on inital load. I could not figure this out so I reverted to old code.
- Change `Field Class` label

Probably want to go back to `nuqs`


src/app/(frontend)/colors.css
```diff
@@ -12,8 +12,8 @@
   --primary-hover: hsla(240 4% 11% / 0.8);
   --primary-foreground: hsl(0 0% 98%);
 
-  --secondary: hsl(240 30% 57%);
-  --secondary-hover: hsla(240 30% 57% / 0.8);
+  --secondary: hsl(240, 23%, 39%);
+  --secondary-hover: hsla(240, 25%, 39%, 0.8);
```
busbyk, line 16: Why the change to the default theme too?

src/constants/eventTypes.ts
busbyk, line 1: Was this change based on feedback from Rusty or the ACs?

I'm wondering because these values came from the Events Google Doc. I know we expect them to change based on AC feedback but it feels weird to remove "by AC" from Field Class but leave it for Events by AC. Although, I get why you did it - because just "Events" without "by AC" feels weird too. Just want to know why we decided to change this, not arguing against it necessarily.

rchlfryn: Yes - I talked to Rusty about this but we are going to change a lot of this anyways

PR #704, opened 2025-11-22
eslint.config.mjs: forbid `as` assertions

stevekuznetsov:
eslint.config.mjs: forbid `as` assertions

TypeScript is only helpful inasmuch as we don't use the escape hatch.

---

*: grandfather all existing `as` assertions

---

The above commit was generated with this Python script operating over the result of running `eslint . -f json -o /tmp/warnings.json`:

```python
#!/usr/bin/env python

import os
import json

warnings_file="/tmp/warnings.json"

warnings = []
with open(warnings_file) as raw:
	warnings = json.load(raw)

for file in warnings:
	lines = []
	for message in file["messages"]:
		if message["ruleId"] == "@typescript-eslint/consistent-type-assertions":
			lines.append(int(message["line"])) #n.b. these are 1-indexed

	if len(lines) > 0:
		segments = []
		with open(file["filePath"]) as raw:
			existing = raw.readlines()
			for i in range(len(lines)+1):
				before = 0
				if i > 0:
					before = lines[i-1] - 1
				after = len(existing)
				if i < len(lines):
					after = lines[i] - 1
				segments.append(existing[before:after])

		with open(file["filePath"], "w") as raw:
			for i in range(len(segments)):
				if i != 0:
					raw.write("// eslint-disable-next-line @typescript-eslint/consistent-type-assertions\n")
				raw.writelines(segments[i])
```

---

*: add a tracker for existing type assertions, validate it

We want to be able to remove type assertions from the codebase at our
own pace while not allowing any new ones. The tracker file lets us
determine which opt-outs we have for existing assertions; validation in
CI keeps us from adding new ones. As we remove existing assertions by
refactoring to use proper types, the `pnpm update-type-assertions`
helper will automate the removal of those entries from our tracker.

---


stevekuznetsov:
Need to be on a branch from the repo, not a fork.

PR #707, opened 2025-11-24
Events date range filter re-implementation with fix

busbyk:
## Description

Re-implementing the `nuqs`-based version of DateRangeFilter + fixes to the "View all" button on the EventList block. 

## Key Changes

See commit messages. 

If you don't pass startDate or endDate to the /api/[center]/events endpoint you will now get future events only. In order to get past events you need to pass an endDate. I think this is only a problem if we wanted to get _all_ past and future events since you would need to pass an endDate that includes all future events. But I think that's fine.

### DIFF
--- a/src/blocks/EventList/Component.tsx
+++ b/src/blocks/EventList/Component.tsx
@@ -32,7 +32,7 @@ export const EventListBlockComponent = (args: EventListComponentProps) => {
 
   const { tenant } = useTenant()
   const [fetchedEvents, setFetchedEvents] = useState<Event[]>([])
-  const [eventParams, setEventParams] = useState<string>('')
+  const [eventsPageParams, setEventsPageParams] = useState<string>('')
 
   useEffect(() => {
     if (eventOptions !== 'dynamic') return
@@ -42,13 +42,13 @@ export const EventListBlockComponent = (args: EventListComponentProps) => {
       if (!tenantSlug) return
 
       const params = new URLSearchParams({
-        center: tenantSlug,
         limit: String(maxEvents || 4),
-        depth: '1',
       })
+      // params supported by the events page
+      const eventsPageParams = new URLSearchParams()
 
       if (filterByEventTypes?.length) {
-        params.append('types', filterByEventTypes.join(','))
+        eventsPageParams.append('types', filterByEventTypes.join(','))
       }
 
       if (filterByEventGroups?.length) {
@@ -56,7 +56,7 @@ export const EventListBlockComponent = (args: EventListComponentProps) => {
           .map((g) => (typeof g === 'object' ? g.id : g))
           .filter(Boolean)
         if (groupIds.length) {
-          params.append('groups', groupIds.join(','))
+          eventsPageParams.append('groups', groupIds.join(','))
         }
       }
 
@@ -65,13 +65,15 @@ export const EventListBlockComponent = (args: EventListComponentProps) => {
           .map((t) => (typeof t === 'object' ? t.id : t))
           .filter(Boolean)
         if (tagIds.length) {
-          params.append('tags', tagIds.join(','))
+          eventsPageParams.append('tags', tagIds.join(','))
         }
       }
-      params.append('startDate', format(new Date(), 'MM-dd-yyyy'))
-      setEventParams(params.toString())
+      eventsPageParams.append('startDate', format(new Date(), 'MM-dd-yyyy'))
+      setEventsPageParams(eventsPageParams.toString())
 
-      const response = await fetch(`/api/${tenantSlug}/events?${params.toString()}`, {
+      const allParams = new URLSearchParams([...params, ...eventsPageParams])
+
+      const response = await fetch(`/api/${tenantSlug}/events?${allParams.toString()}`, {
         cache: 'no-store',
       })
 
@@ -128,9 +130,9 @@ export const EventListBlockComponent = (args: EventListComponentProps) => {
               <h3>There are no events matching these results.</h3>
             )}
           </div>
-          {eventOptions === 'static' && (
+          {eventOptions === 'dynamic' && (
             <Button asChild className="not-prose md:self-start">
-              <Link href={`/events?${eventParams}`}>View all {heading}</Link>
+              <Link href={`/events?${eventsPageParams}`}>View all {heading}</Link>
             </Button>
           )}
         </div>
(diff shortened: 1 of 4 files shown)


src/components/filters/DateRangeFilter.tsx
```diff
@@ -95,40 +81,39 @@ export const DateRangeFilter = ({
...
+  useEffect(
+    function determineMatchingQuickFilter() {
+      if (isInitialMount.current) {
+        // Use initial props for first mount
+        const currentStart = startDate || initialStartDate
+        const currentEnd = endDate || initialEndDate
+
+        if (!currentStart && !currentEnd) {
+          // No dates set, set start date and set quick filter to upcoming
+          handleQuickFilter('upcoming')
```
rchlfryn, line 93: I had this , but it was only working on a refresh. Did you experience this as well?

rchlfryn (changes requested):
One minor thing...I click `Events` in nav while on it, it should reset the quick date filter. It looks like the events update but we need to update the filter as well.
Kapture 2025-11-24 at 16 43 35

src/components/filters/DateRangeFilter.tsx
rchlfryn: I think it might have been the addition of the query params... I was so close!

src/components/filters/DateRangeFilter.tsx
```diff
@@ -95,40 +81,39 @@ export const DateRangeFilter = ({
...
+        } else if (currentStart && !currentEnd) {
+          // Only start date, default to upcoming
+          setQuickFilter('upcoming')
```
rchlfryn: Shouldn't this be handled the same way as line 91? I dont know if we want people choosing any random start date as a param and no end date

src/blocks/EventList/Component.tsx
```diff
@@ -128,9 +130,9 @@ export const EventListBlockComponent = (args: EventListComponentProps) => {
               <h3>There are no events matching these results.</h3>
             )}
           </div>
-          {eventOptions === 'static' && (
+          {eventOptions === 'dynamic' && (
```
rchlfryn, line 133: Do we only want to show this for dynamic? It seems like it should be its own setting

busbyk:
> One minor thing...I click `Events` in nav while on it, it should reset the quick date filter. It looks like the events update but we need to update the filter as well. 

Ah yes... that's because, this being a client side navigation, the `isInitialMount` ref is still true (because the component hasn't been unmounted so we skip the check. I'll try always checking for `!startDate && !endDate` and see if that would result in any issues. That seems reasonable.

src/components/filters/DateRangeFilter.tsx
busbyk: All good! Yea the query params had to be set too.

src/blocks/EventList/Component.tsx
busbyk: Hm yea that's interesting, we could have a "view all button" group with a label and link field. But the way it is right now, I think it only makes sense to show a view all button for dynamic because we can use the same params from the block in the `/events?(params.toString())` url. For static options, where would the link go? We'd need the user to set that. 

That could be a cool future enhancement. I'd want to ask current users if they would use that, though.

busbyk: This is the same logic as for the BlogList block: https://github.com/NWACus/web/blob/events-date-range-filter/src/blocks/BlogList/Component.tsx#L78 (although written differently because of the different way we query)

src/components/filters/DateRangeFilter.tsx
busbyk: Yea good call. I ended up removing this else block and trying to match against quick filters if there is a startDate (regardless of if there is an endDate or not) so that the "custom" quick filter will be matched if a user enters a startDate different from the current date. I think that's the right behavior here. They would need to manually be modifying the url params of course but users may do that.

busbyk:
> > One minor thing...I click `Events` in nav while on it, it should reset the quick date filter. It looks like the events update but we need to update the filter as well.
> 
> Ah yes... that's because, this being a client side navigation, the `isInitialMount` ref is still true (because the component hasn't been unmounted so we skip the check. I'll try always checking for `!startDate && !endDate` and see if that would result in any issues. That seems reasonable.

This was indeed more complicated than I initially realized.

See https://github.com/NWACus/web/pull/707/commits/42196f086bf5ca7b606787df4c330d71cbf4e0e8 for updates. What do you think about this approach? 

This works for page reloads, navigating to the events page via a Link component from other pages and also from the event page. 

https://www.loom.com/share/c9a6d1f67196420180bceb0a7d7f1f46

rchlfryn:
This looks great! 

> > > One minor thing...I click `Events` in nav while on it, it should reset the quick date filter. It looks like the events update but we need to update the filter as well.
> > 
> > 
> > Ah yes... that's because, this being a client side navigation, the `isInitialMount` ref is still true (because the component hasn't been unmounted so we skip the check. I'll try always checking for `!startDate && !endDate` and see if that would result in any issues. That seems reasonable.
> 
> This was indeed more complicated than I initially realized.
> 
> See [42196f0](https://github.com/NWACus/web/commit/42196f086bf5ca7b606787df4c330d71cbf4e0e8) for updates. What do you think about this approach?
> 
> This works for page reloads, navigating to the events page via a Link component from other pages and also from the event page.

This looks great! I think this approach will work.

src/blocks/EventList/Component.tsx
rchlfryn: Let's keep it as a future enhancement. I was just curious if this was something requested or just randomly decided

src/components/filters/DateRangeFilter.tsx
```diff
@@ -28,43 +27,31 @@ export const DateRangeFilter = ({
...
-    [pathname, router, searchParams],
   )
 
+  const [quickFilter, setQuickFilter] = useState('')
+  const [isOpen, setIsOpen] = useState(defaultOpen)
+  const isInitialMount = useRef(true)
+
   const updateDateSelection = useCallback(
     (filter: string, start: string, end: string) => {
+      console.log(`Setting filter to ${filter} and startDate: ${start} and endDate: ${end}.`)
```
rchlfryn:
```suggestion
```

rchlfryn (approved):
LG - one minor thing

src/components/filters/DateRangeFilter.tsx
busbyk: Ah a classic "forgot to remove my debug logs 🤦 " haha. Thanks.

src/blocks/EventList/Component.tsx
busbyk: Cool, yea just based on what made sense to me. We only have query params when using the 'dynamic' option and just making the button link to the /events page doesn't feel like it fits the use case of picking specific events.

src/components/filters/DateRangeFilter.tsx
busbyk: Oh hilarious - I'd already removed it and forgot to push. Even more classic 😏

PR #716, opened 2025-11-25
Events - fix table error & desktop nav z-index

rchlfryn:
## Description
- Removes improper error handling on events table
- Fix desktop nav z-index with introduction of featured image page

## Key Changes
- Remove try block from events table
  - If you only had past events in a table, the table would completely error out the page. 
- Add `z-index` to  desktop nav

## Migration Explanation
Just added to  `events` branch migration


src/blocks/EventTable/Component.tsx
```diff
@@ -22,63 +22,55 @@ export const EventTableBlockComponent = (args: EventTableComponentProps) => {
...
 
     const fetchEvents = async () => {
-      try {
-        setError(null)
+      const tenantSlug = typeof tenant === 'object' && tenant?.slug
+      if (!tenantSlug) return
 
-        const tenantSlug = typeof tenant === 'object' && tenant?.slug
-        if (!tenantSlug) return
+      const params = new URLSearchParams({
```
busbyk, line 34: Could you implement the same pattern as [here](https://github.com/NWACus/web/pull/707/files#diff-67cc6a6795494de445c30ee26d665ec143c354ff793024296715c5961a80a1a1) where only support query params are sent to `getEvents` via `src/app/api/[center]/events/route.ts`? `center` and `depth` are not supported in `src/app/api/[center]/events/route.ts`, for example. 

They'll be safely ignored but I don't want future us to think they're required for this api call.

src/app/(frontend)/[center]/events/[slug]/page.tsx
```diff
@@ -84,7 +84,7 @@ export default async function Event({ params: paramsPromise }: Args) {
         {event.featuredImage && (
           <Media
             resource={event.featuredImage}
-            className="w-full absolute"
+            className="w-full absolute z-1"
```
Comment (busbyk) [inline src/app/(frontend)/[center]/events/[slug]/page.tsx]: Hey I think we could do without this - just setting `z-20` on the `NavigationMenu` in `DesktopNav.client.tsx` seems to be sufficient. Were you seeing something different?

Good catch on this btw, I never noticed that!

src/blocks/EventTable/Component.tsx
```diff
@@ -22,63 +22,55 @@ export const EventTableBlockComponent = (args: EventTableComponentProps) => {
...
-        })
+      if (filterByEventTypes?.length) {
+        params.append('types', filterByEventTypes.join(','))
+      }
 
-        if (filterByEventTypes?.length) {
-          params.append('types', filterByEventTypes.join(','))
+      if (filterByEventGroups?.length) {
+        const groupIds = filterByEventGroups
+          .map((g) => (typeof g === 'object' ? g.id : g))
```
busbyk, line 47: This might be a nice utility function at some point -- like get id from value that might be a document object or a number which is the id. Just an aside.

busbyk (commented):
Looks good. Just a few comments. 

Last we chatted, we'd talked about just using tags for affinity groups for events and centers could add them or not. But it's nice that centers won't have to add these manually / it encourages them to use these ones specifically so I like this addition.

src/app/(frontend)/[center]/events/[slug]/page.tsx
Comment (rchlfryn) [reply src/app/(frontend)/[center]/events/[slug]/page.tsx]: Ah you're right - removed the z-index..

src/blocks/EventTable/Component.tsx
rchlfryn: I like this! Updated

src/blocks/EventTable/Component.tsx
```diff
@@ -22,67 +22,74 @@ export const EventTableBlockComponent = (args: EventTableComponentProps) => {
...
-        setError(null)
+      const tenantSlug = typeof tenant === 'object' && tenant?.slug
+      if (!tenantSlug) return
 
-        const tenantSlug = typeof tenant === 'object' && tenant?.slug
-        if (!tenantSlug) return
+      const params = new URLSearchParams({
+        center: tenantSlug,
+        limit: String(maxEvents || 4),
+        depth: '1',
```
busbyk: You can remove `center` and `depth` here. They don't do anything in `src/app/api/[center]/events/route.ts`

src/blocks/EventTable/Component.tsx
```diff
@@ -22,67 +22,74 @@ export const EventTableBlockComponent = (args: EventTableComponentProps) => {
...
+      })
 
-        const params = new URLSearchParams({
-          center: tenantSlug,
-          limit: String(maxEvents || 4),
-          depth: '1',
-        })
+      // params supported by the events page
+      const eventsTableParams = new URLSearchParams()
```
busbyk: Can you name this `eventsPageParams`?

busbyk (changes requested):

busbyk (approved):
Thanks!

PR #718, opened 2025-11-26
WidgetRouterHandler: allow deep linking to widget routes

### LINKED ISSUE #567: Linking externally to filtered widget page not working properly

busbyk:
## Description

This allows linking to a different widget route than the initial route. 

The `afp-public-widgets` that have more than a single view use `vue-router`'s [`createWebHashHistory`](https://router.vuejs.org/api/functions/createWebHashHistory.html) function, enabling `vue-router`'s [hash mode](https://router.vuejs.org/guide/essentials/history-mode.html#Hash-Mode) for routing. When embedded in AvyWeb, this means there are two routers at play: 1) our router (the next.js routing system) which uses the url's path and 2) the widgets' router which uses the url's hash. 

So we need to make these play nicely together. Before this PR we were _always_ resetting the hash to the `initialHash` by passing this into the former `WidgetHashHandler`. 

The key issue with this for us was that you couldn't "deep link" to a widget route besides the `initialHash` we had set.

This PR allows that kind of "deep linking" and also allows query params in the hash to persist and be used by the widgets. 

## Related Issues

Resolves #567 

## Key Changes

- Adds `src/components/NACWidget/widgetRouter.ts` which contains a subset of all allowed routes for each widget except that they have been grouped by our widget pages. Our widget pages don't always align with the widget types from `afp-public-widgets` so I made the decision to align routes by our widget pages. 
  - Widget routes for reference: [forecasts](https://github.com/NationalAvalancheCenter/afp-public-widgets/blob/main/src/apps/forecasts/router/index.js), [observations](https://github.com/NationalAvalancheCenter/afp-public-widgets/blob/main/src/apps/observations/router/index.js), [stations](https://github.com/NationalAvalancheCenter/afp-public-widgets/blob/main/src/apps/stations/router/index.js)
- Refactors `WidgetHashHandler` into `WidgetRouterHandler` (makes more sense to me to name it this way)
  - `useHash` isn't set on the first render so it falls back to `window.location.hash`. We still need `useHash` to watch for hash changes so we can "redirect" to our `initialPath` if an invalid route is set programmatically. 
  - We check for validity of the widget route (from the url hash) and "redirect" to the `initialPath` if not. This works for when there is no hash as well as when there is an "invalid" (i.e. not in the `widgetRouter`) hash. 

## Screenshots / Demo

https://www.loom.com/share/869cd799a29143b7ad1b335e554fd00d
https://www.loom.com/share/ca082066a8784a75bde459a9d75e335f

The browser still moves the scroll position down to what I believe is a matching html element to the hash (i.e an anchor) based on it's id attribute. This is annoying to me but does not affect functionality so I will leave that for a future PR. 

I left out the following path from the router even though it is a valid path for the forecast widget router:
`/media/:id` -- I couldn't find any links to this view. There are just some [commented-out links](https://github.com/NationalAvalancheCenter/afp-public-widgets/blob/staging/src/components/forecasts/views/Media.vue#L4), seemingly for testing. This route is mentioned [here](https://nationalavalanchecenter.atlassian.net/wiki/spaces/AFP/pages/6619179/Avalanche+Forecast+Widget#:~:text=%23/media/%3Cid%3E%20%2D%20Page%20showing%20the%20media%20and%20its%20associated%20metadata.%20Grab%20these%20links%20from%20the%20Dashboard%20Media%20Gallery.) but I'm guessing this might be outdated. 

I initially left out the following paths because I thought they weren't used but realized that they are for the ["Conditions Blog" product](https://nationalavalanchecenter.atlassian.net/wiki/spaces/AFP/pages/10354695/Tutorial+Avalanche+Forecast+Module+-+Blog+Product) which Sawtooth uses currently. So I've left these in:

- `'/blog', '/blog/:id'` -- These are redirect-only routes. `/blog` -> `/post` and `/blog/:id` -> `/post/:id`
- `'/post', '/post/:id'` -- These are the Conditions Blog routes - there is a tab on the forecast zone view for this:
CleanShot 2025-11-26 at 08 59 46@2x

Limiting the allowed paths for the widgets' routers isn't strictly necessary. We could change the logic for the `WidgetRouterHandler` to only redirect to the `initialPath` if there is _no_ hash and let the widgets themselves handle invalid paths. I like being able to limit the valid paths for the widget based on which AvyWeb page the visitor is on, though. This ensures that a user can't end up on `#/view/observations` while on our `/observations/submit` route, for example. Do we think this is overkill or a nice level of control? 

At the very least, I like having the allowed paths listed here by page for reference.


src/components/NACWidget/WidgetRouterHandler.client.tsx
```diff
@@ -0,0 +1,48 @@
...
+  PathMatcher,
+  WidgetPageWithRouterKey,
+} from './widgetRouter'
+
+export function WidgetRouterHandler({
+  initialPath,
+  widgetPageKey,
+}: {
+  initialPath: string
+  widgetPageKey?: WidgetPageWithRouterKey
```
rchlfryn: I see that this is optional, but is there a reason to not make it required? I didn't see a case where it wasn't being used.

rchlfryn (commented):
I am very excited about these changes. One minor question about the `widgetPageKey`

> Do we think this is overkill or a nice level of control?
> At the very least, I like having the allowed paths listed here by page for reference.

I think it is a nice level of control, and also like limiting the allowed paths for the widgets' routers

src/components/NACWidget/WidgetRouterHandler.client.tsx
busbyk: I remembered there being one widget that needed an initial path but not having any other allowed paths. I think it must have been the forecasts widget and I just added it as a widgetPage. Either way, good catch - I'll make that a required prop.

busbyk:
> > Do we think this is overkill or a nice level of control?
> > At the very least, I like having the allowed paths listed here by page for reference.
> 
> I think it is a nice level of control, and also like limiting the allowed paths for the widgets' routers

Great, I agree.

rchlfryn (approved):

PR #719, opened 2025-11-26
Events - update types, skill level 

rchlfryn:
## Description
Update events and skill level

## Issues
When selecting a type, the selector was selecting the same type multiple time. Turn out it was a react key warning error. I was able to solve this by adding a key to the options. I looked elsewhere and we don't need to add this.
>  Encountered two children with the same key, Event-event. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version

## Migration Explanation
Need to merge #716 and then rerun migration


busbyk (approved):
🤘

PR #720, opened 2025-11-27
Single observation page

### LINKED ISSUE #643: Opening Observation in a new tab reverts back to Obs page
**Describe the bug**
When opening the unique link for each observation it sends you back to the obs landing page

**To Reproduce**

1. Go to https://www.sierraavalanchecenter.org/observations/#/view/observations
2. Click on a new observation
3. Click this button: 
Image

Another way would just be to go to the link of an observation. Someone at SAC was trying to [share this link](https://www.sierraavalanchecenter.org/observations/#/observation/012d1cd0-8498-43da-bdda-e2cafe0eeff2) with a bud.

I would expect it to open in a new tab.

busbyk:
## Description

Adds a dedicated single observation page as a widget, built-in page. 

There are two variants to this page: one for observations and one for avalanches. This is necessary because they need different `initialPath` values.

I didn't add these to the built-in pages in the README because these are dynamic pages based on the observation/avalanche id. 

We just missed this route when building the widget pages initially. 

## Related Issues

Resolves #643 

## Key Changes

- Adds two new routes: `/observations/[id]` and `/observations/avalanches/[id]`.
- Uses css in favor of the NACContainerRemover component which removed the `nac-container` class from the element. We just didn't need to use javascript for this. Silly past me. 
- Adds extra padding between the heading and widget on the `/observations/submit` page
- Adds a `BreadcrumbProvider` and hook which allows the use of `BreadcrumbSetter` to override the page label (i.e. last breadcrumb). This was necessary because the observation id used as the last breadcrumb looked terrible. So I used the widget heading as the last breadcrumb label. 

## Screenshots / Demo

https://www.loom.com/share/bd2bb895ba494b919eb4119e8c23c288

1. In an attempt to have to make a call to the avalanche.org API (to not expand the scope of resolving #643) I am overriding the page breadcrumb. This results in a flash of the id before it's updated on the client: 
CleanShot 2025-11-26 at 18 04 48

It's not ideal but I think we can accept this for now rather than depending on a request to the avalanche.org API to get the observation/avalanche title. This is primarily a time-constraint decision. Thoughts?

2. I used a single widget page in `src/components/NACWidget/widgetRouter.ts`: `  'single-observation': ['/observation/:id', '/avalanche/:id'],`. Is that confusing? Should I just use two different widget pages?

3. When hijacking the observation link button, this doesn't open it in a new tab like the widget's behavior. Should it? 
CleanShot 2025-11-26 at 18 31 11@2x


src/components/Breadcrumbs/BreadcrumbSetter.client.tsx
```diff
@@ -0,0 +1,15 @@
+'use client'
```
stevekuznetsov, line 1: Why does this behavior need to live client-side? If we know the segments of the path server-side when rendering this page and we override it there won't we remove the flicker?

busbyk: Good question. Let me explain a bit first:

`<Breadcrumbs />` is [only rendered once](https://github.com/NWACus/web/blob/single-observation-page/src/app/(frontend)/%5Bcenter%5D/layout.tsx#L80), in `src/app/(frontend)/[center]/layout.tsx` 

That's a client component but (and you may already know this) client components are actually rendered on the server during SSR - they just can't be statically rendered. But they will render without hydration. So `useEffects` and other react hooks will not run during SSR. 

But since `<Breadcrumbs />` is only rendered in the layout, in order for a child component to update it, we need to use a provider/context/hook pattern. We use the hook to update the provider's context, which is used by `<Breadcrumbs />`. But since we must use a `useEffect` for this, that use effect only runs after client side hydration. This results in the flash of SSR content before hydration. 

`<Breadcrumbs />` uses `useSelectedLayoutSegments()` which uses the next.js router to determine the path segments. This works during SSR because it's using the next.js router and doesn't depend on browser APIs. Breadcrumbs also uses other client hooks like `useAnalytics` and `useNotFound` which make it more difficult to refactor to use a server first approach. 

A couple alternative approaches to breadcrumbs we could use:
1. Render `<Breadcrumbs />` on every page instead of in the layout and accept a pageLabel prop. This would work during SSR, we would just need to ensure that we always include a `<Breadcrumbs />` on every page. Not a huge deal for the refactor but I could see that being forgotten when adding a new route. 
2. Use parallel routes. Here's an article discussing an approach to do that: https://jeremykreutzbender.com/blog/app-router-dynamic-breadcrumbs

__tests__/client/Breadcrumbs.client.test.tsx
```diff
@@ -80,4 +90,23 @@ describe('Breadcrumbs', () => {
       'href',
     )
   })
+
+  it('overrides the last breadcrumb label when BreadcrumbSetter is used', () => {
```
rchlfryn, line 94: 👏🏼

src/components/Breadcrumbs/BreadcrumbSetter.client.tsx
busbyk: Created a follow up issue: https://github.com/NWACus/web/issues/728

If we really hate this flicker, we could implement alternative 1?

rchlfryn (approved):
> 1. This is primarily a time-constraint decision. Thoughts?

I think this is fine to leave for now, especially with making the follow up issue to look into rendering breadcrumbs as a sever component

> 2. I used a single widget page in src/components/NACWidget/widgetRouter.ts:   'single-observation': ['/observation/:id', '/avalanche/:id'],. Is that confusing? Should I just use two different widget pages?

I think it is fine to leave them as 1 page because it will just be the same page. We will likely change this if widgets get updated.

> 3. When hijacking the observation link button, this doesn't open it in a new tab like the widget's behavior. Should it?

I think it should remain in the same window. A bummer we cant update the icon somehow.

PR #724, opened 2025-11-30
Events - do not use `dbName` & fix `futureDate` and `pastDate`

### LINKED ISSUE #714: Error setting filters in event list block
**Describe the bug**
We get a "cannot read properties of undefined (reading 'tenant')" error when changing the event types filter on an event list block on home page - needs further investigation. Happening on events preview environment.

**To Reproduce**

1. Visit a homePages document on the `events` branch
2. Change the event types filter on the event list block that was created from the seed script on every home page
3. See the "cannot read properties of undefined (reading 'tenant')" error appear in a toast
4. Try to save the document and observe the FOREIGN KEY constraint error

**Expected behavior**
No errors for this flow

**Screenshots**
https://www.loom.com/share/56e917402ed94a50a5bf37c36cd4c790

**Environment details**

- Desktop
- Chrome

rchlfryn:
## Description
Shortens database names to avoid using `dbName` which was causing a foreign key error
https://github.com/payloadcms/payload/issues/14350

Fixes `futureDate` and `pastDate` seed functions

## Related Issues
Fixes #714 

## Key Changes
- Shorten db names
- `futureDate` and `pastDate` functions were erroring out because the month offset was trying to set months to 13 & 14 which don't exist. I fixed this by adding a mod function and year to wrap the dates. 

## Migration Explanation
Need migration to update names

### DIFF
--- a/src/fields/EventQuery/config.ts
+++ b/src/fields/EventQuery/config.ts
@@ -39,16 +39,16 @@ export const defaultStylingFields = (additionalFilters?: Field[]): Field[] => [
   {
     type: 'radio',
     name: 'eventOptions',
-    label: 'How do you want to choose your events?',
+    label: 'How do you want to select your events?',
     defaultValue: 'dynamic',
     required: true,
     options: [
       {
-        label: 'Do it for me',
+        label: 'Use filters',
         value: 'dynamic',
       },
       {
-        label: 'Let me choose',
+        label: 'Manually',
         value: 'static',
       },
     ],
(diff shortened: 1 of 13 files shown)


src/fields/EventQuery/config.ts
```diff
@@ -39,16 +39,16 @@ export const defaultStylingFields = (additionalFilters?: Field[]): Field[] => [
...
     required: true,
     options: [
       {
-        label: 'Do it for me',
+        label: 'Use filters',
         value: 'dynamic',
       },
       {
-        label: 'Let me choose',
+        label: 'Manually',
```
busbyk, line 51: 👏

busbyk (approved):
Looks good!

PR #738, opened 2025-12-03
Event timezones on date fields

busbyk:
## Description

The timezone field on events and courses was misleading. If a user set the timezone, the assumption would be that the start/end dates would be set for that timezone (i.e. 1pm in PST). We were correctly formatting the dates on the frontend based on the timezone field but the issue was that Payload was storing the dates based on the browser's timezone where the admin panel user saved the event. This could be different from the timezone field value. 

This PR utilizes Payload's date field timezone config which _does_ affect the date being stored in the database. Dates are (and were before) stored as ISO strings but using the timezone config influences the ISO string generated and stored in the db. See the [docs](https://payloadcms.com/docs/fields/date#timezones). Using this config results in an additional db column for a date field {date field name}_tz. This displays a timezone picker (with our custom timezone labels and subset of timezones) below the date field:
CleanShot 2025-12-02 at 18 05 30

I also updated the date formatting to always include timezone and to use a 4 digit year instead of 2 (it was confusing next to the date). I think we may want to mess around with making the date displays more concise but I'd rather err on the side of accuracy for now. 

I decided not to add a timezone to the tenants collection for now, although I think we may want to use that for the default value of dates' timezone fields in the future. For now, we default the startDate timezone to the browser timezone and make timezone required. We also set endDate and registrationDeadline timezones to always match startDate. 

## Key Changes

- Removes timezone field from events and courses
- Adds timezone config to the date fields: startDate, endDate, registrationDeadline for events and courses
- Adds a UI component to set the initial value for startDate_tz if not set and to keep endDate_tz and registrationDeadline_tz timezones in sync
- Adds beforeValidate hooks to events and courses collections (mainly to check that the timezones are all the same but we also check that endDate is after startDate)
- Updated seed script
- Updates EventTable to use the same formatDateTime function that the rest of the events and courses components use.

## Screenshots / Demo

https://www.loom.com/share/06b146d97fef40318f7d558a07041e38

## Migration Explanation

Updated the migration accordingly. 

I think adding a default timezone to tenants in the future would be nice but this seems fine for now.


rchlfryn (approved):
LG - future request, we should just make out own custom date picker that accommodates a readonly timezone picker.

busbyk:
> LG - future request, we should just make out own custom date picker that accommodates a readonly timezone picker.

I think it would be sweet to contribute to Payload's - I like that UI design. It would also be nice to accept a function for the timezone default value.

PR #759, opened 2025-12-04
Update BlogList block

### LINKED ISSUE #601: Bug: BlogListBlock maxPosts validation can cause infinite loop with a decimal
**Describe the bug**
When you use a non whole number in the maxPosts field of the BlogListBlock, the block continually fetches posts resulting in an infinite loop. 

**To Reproduce**
1. Add a BlogListBlock to a blocks or richText field
2. Change the maxPosts value to 4.5 (or any non whole number)
3. Observe an infinite refetching of the Preview Posts (may be helpful to open the dev tools but should be obvious)

**Expected behavior**
A validation error is thrown and fetching is delayed until this field is valid.

**Screenshots**

Image

**Environment details**

- Desktop
- Chrome

**Existential musings**
Marmots are all cute and adorable until they chew through your backpack strap while you're on a climb and you have to duck tape it back together to hike out.

rchlfryn:
## Description
Payload was making this more difficult than this needed to be. The `step` admin setting on the number field in the admin config does not work how we would want it to and making a custom component seems like overkill, especially since we removed queried __ in events. Because of that I just removed it here and edited PostList to use `getPosts`

## Related Issues
Fixes #601 

## Key Changes
Remove queried posts from BlogList and update `postOptions` labels

## Migration Explanation
None

We will probably want to add `showViewButton` to other blocks.


src/blocks/BlogList/config.ts
```diff
@@ -43,11 +43,11 @@ const defaultStylingFields: Field[] = [
     required: true,
     options: [
       {
-        label: 'Do it for me',
+        label: 'Use filters',
         value: 'dynamic',
       },
       {
-        label: 'Let me choose',
+        label: 'Manually',
```
busbyk, line 50: 🧡

src/blocks/BlogList/Component.tsx
```diff
@@ -1,50 +1,95 @@
+'use client'
```
busbyk, line 1: I think this could be a server component. 

Events render differently over time without there being an event that triggers a revalidation. i.e. time passing changes their state and we can't revalidate pages that have been statically generated based on that (although our time based revalidation might be good enough here 🤔). 

But blog posts don't have that same behavior. So using a server component here seems like it would be preferred. We could avoid the whole process of creating query params from filters and just call `getPosts` directly. 

Right?

busbyk (changes requested):
I do like the shift from storing the results of the block's filters directly on the block to querying in the component based on the filters. I think there was a lingering bug here where if something changed related to a post (like a tag) that should affect the _result_ of the query based on the filters, that wouldn't trigger the admin panel to re-query and update the queriedPosts field. So I think this a good call. 

But this doesn't address the fact that you can set a non whole number for maxPosts. This still results in an error: 
Image

So we need to enforce that maxPosts is a whole number. We could use a validate function if we don't want to go to the trouble of creating a custom component. Would probably want to use a validate function in conjunction with a custom component anyways.

busbyk:
> We will probably want to add `showViewButton` to other blocks.

Did you mean to add this to this PR? I like the current logic of only showing a view all button for dynamic posts/events. But you were thinking it would be nice to have a more generic call to action button that's configurable with a label, link (page/post/built-in reference for internal, href for external), etc. ?

rchlfryn:
> Did you mean to add this to this PR? I like the current logic of only showing a view all button for dynamic posts/events. But you were thinking it would be nice to have a more generic call to action button that's configurable with a label, link (page/post/built-in reference for internal, href for external), etc. ?

I did not want to add the code change to this PR, but editing the config made me think about adding an option to show/hide view all button to all configs with this set up. I think the view all button can be used for static events too. I will make a follow up issue for us to discuss and decide.

rchlfryn:
As we discussed, the validation `validateMaxPosts` prevents decimals from being saved.
Screenshot 2025-12-04 at 11 45 48

But this does not do anything to help live preview. To help live preview we updated the queries [b37a6a9](https://github.com/NWACus/web/pull/759/commits/b37a6a96ec48ee6201405f9764913a296118ffa8) to round all decimal numbers down. Tested and it works
Screenshot 2025-12-04 at 11 59 39

busbyk (approved):

PR #760, opened 2025-12-04
Bump Next.js and Payload versions

busbyk:
## Description

Bumps Next.js to 15.4.8 to address [CVE-2025-66478](https://github.com/vercel/next.js/security/advisories/GHSA-9qr9-h5gf-34mp)

Bumps Payload to 3.66.0


rchlfryn (changes requested):
Can we include why we make all these other file changes in the description? Are the necessary for the upgrade, or nice to have with the upgrade?

src/app/(frontend)/[center]/next/preview/route.ts
```diff
@@ -1,17 +1,10 @@
 import configPromise from '@payload-config'
 import { draftMode } from 'next/headers'
 import { redirect } from 'next/navigation'
+import { type NextRequest } from 'next/server'
 import { CollectionSlug, getPayload, type PayloadRequest } from 'payload'
 
-export async function GET(
```
Comment (busbyk) [inline src/app/(frontend)/[center]/next/preview/route.ts:6]: The new version of next.js complained about this type.

src/payload-types.ts
```diff
@@ -151,6 +151,7 @@ export interface Config {
   db: {
     defaultIDType: number;
   };
+  fallbackLocale: null;
```
busbyk, line 154: This was added by the new version of Payload.

eslint.config.mjs
```diff
@@ -44,7 +44,7 @@ const eslintConfig = [
       ],
     },
   },
-  globalIgnores(['.next/', 'src/migrations/']),
+  globalIgnores(['.next/', 'src/migrations/', 'next-env.d.ts']),
```
busbyk, line 47: This was added to ignore an eslint error in `next-env.d.ts`. It makes sense to ignore this file completely because it is auto-generated by next.js.

next-env.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference path="./.next/types/routes.d.ts" />
```
busbyk, line 3: This was added by the new version of Next.js

busbyk (commented):
Notes on additional changes beyond version bumps.

busbyk:
> Can we include why we make all these other file changes in the description? Are the necessary for the upgrade, or nice to have with the upgrade?

Sure, I left comments on the specific files: https://github.com/NWACus/web/pull/760#pullrequestreview-3541593135

rchlfryn (approved):
Awesome! Thanks for doing that. I assumed it was auto generated but wanted to be sure.

Issue #743, opened 2025-12-03
Rusty cannot invite users


rustynwac:
**Describe the bug**
I get a blank screen upon clicking "Add User"

**To Reproduce**

1. Login as Provider Manager user or Super admin user
2. Users
3. Click invite user

**Expected behavior**
I would expect the user module to open

**Screenshots**
https://www.loom.com/share/425b9a2e90494589997b8d47e65e24b9

Here's a screenshot of the provider user I was logged in as (from my super admin user)

<img width="899" height="717" alt="Image" src="https://github.com/user-attachments/assets/0622319b-278b-462e-b712-085d4bf10c9b" />

**Environment details**

- Chrome
- Could also replicate with Safari

**Existential musings**
What else you got for us?

rustynwac:
Resolved itself

rustynwac:
Getting this error for all users right now

PR #764, opened 2025-12-05
Bugfix: invite user drawer + Sentry config

busbyk:
## Description

I can only reproduce #743 in production (not in dev, not locally with a production build). I can only assume this is caused by data in the production environment. I'm attempting to add defensive checks to see if they will resolve the error along with capturing additional error statuses in the Payload Sentry plugin. 

## Related Issues

#743


src/collections/Users/components/InviteUserDrawer.tsx
```diff
@@ -95,19 +95,17 @@ const initialState: FormState = {
...
-    valid: true, // Valid by default since it's optional
+    valid: true,
     value: [],
-    rows: [],
   },
   providers: {
     initialValue: [],
-    valid: true, // Valid by default since it's optional
+    valid: true,
     value: [],
```
busbyk, line 109: Removed the empty rows array to start because it seemed like Payload was resetting this anyways and it's not necessary - thought maybe it's messing with the form state. Removed the comments just because they seem unnecessary.

src/collections/Users/components/InviteUserDrawer.tsx
```diff
@@ -128,33 +126,148 @@ export function InviteUserDrawer({
...
+      rolesData?.docs
+        ?.filter(
+          (role: Role | null): role is Role => role != null && role.name != null && role.id != null,
+        )
+        .map((role: Role) => ({
+          label: role.name,
+          value: String(role.id),
+        })) || [],
+    [rolesData],
+  )
```
busbyk, line 140: Memoizing just as a good practice for performance. Filtering to ensure we get all of the data we expect and label and value are always valid.

src/collections/Users/components/InviteUserDrawer.tsx
```diff
@@ -128,33 +126,148 @@ export function InviteUserDrawer({
...
+            globalRole != null && globalRole.name != null && globalRole.id != null,
+        )
+        .map((globalRole: GlobalRole) => ({
+          label: globalRole.name,
+          value: String(globalRole.id),
+        })) || [],
+    [globalRolesData],
+  )
+
+  const [{ data: providersData }] = usePayloadAPI('/api/providers?limit=0')
```
busbyk, line 172: Changed `limit=100` to `limit=0` to fetch all providers since I know we'll at least always have 150 providers after we create the existing ones. Not super performant but the "make it work" step of make it work, make it right, make it fast.

src/collections/Users/components/InviteUserDrawer.tsx
```diff
@@ -128,33 +126,148 @@ export function InviteUserDrawer({
...
+            provider != null && provider.name != null && provider.id != null,
+        )
+        .map((provider: Provider) => ({
+          label: provider.name,
+          value: String(provider.id),
+        })) || [],
+    [providersData],
+  )
+
+  // Debug logging for production issue investigation
```
busbyk, line 187: Will remove this once we've resolved this issue.

src/collections/Users/components/InviteUserDrawer.tsx
```diff
@@ -351,10 +509,10 @@ export function InviteUserDrawer({
             </div>
           )}
 
-          {canCreateGlobalRoleAssignments && (
+          {canCreateGlobalRoleAssignments && globalRoleOptions.length > 0 && (
             <SelectField
               field={{
-                name: 'globalRoles',
+                name: 'globalRoleAssignments',
```
busbyk, line 515: I was curious if this name mismatch might cause issues with Payload's form state.

src/collections/Users/components/InviteUserDrawer.tsx
```diff
@@ -351,10 +509,10 @@ export function InviteUserDrawer({
             </div>
           )}
 
-          {canCreateGlobalRoleAssignments && (
+          {canCreateGlobalRoleAssignments && globalRoleOptions.length > 0 && (
```
busbyk, line 512: Only rendering if there are options.

src/plugins/index.ts
```diff
@@ -79,5 +79,12 @@ export const plugins: Plugin[] = [
...
   }),
-  sentryPlugin({ Sentry }),
+  sentryPlugin({
+    Sentry,
+    options: {
+      debug: true,
+      // >= 500 status codes are captured automatically in addition to these
+      captureErrors: [400, 401, 403, 404],
+    },
+  }),
```
busbyk, line 89: We haven't gotten any Sentry issue logged from Payload. I would have expected the sentry plugin's error boundary to catch this so I'm setting debug: true and explicitly capturing more than just the default 500 errors (although I don't think that would change anything here).

src/collections/Users/components/InviteUserDrawer.tsx
stevekuznetsov: Won't the react compiler do this for us?

src/collections/Users/components/InviteUserDrawer.tsx
stevekuznetsov: If it's used as a form-fill for "which of our providers will you manage" I don't know that you can do better than showing all of them

busbyk: Yea I think pulling all of them is totally fine for this case. But Payload's select components do have the capability to have an infinite scrolling style implementation where additional elements are fetched as you scroll. You'll see this in places in the admin panel like selecting a page, post, or built-in page for a link field.

src/collections/Users/components/InviteUserDrawer.tsx
busbyk: I believe it would, yea. We don't have react-compiler enabled for this project. The first stable version of it only came out in October: https://react.dev/blog/2025/10/07/react-compiler-1. 

I've heard nothing but good things but I haven't read about it in depth yet. I'd want to understand it in a little bit more detail before turning it on. Looks like they have pretty good guidance on things I'm curious about like:
- Should we remove existing `useMemo`, `useCallback` hooks or keep them
- When hooks like those are still necessary 
- Any breaking changes

Created an issue to track that: https://github.com/NWACus/web/issues/766

PR #774, opened 2025-12-09
Update tenant slug readonly

### LINKED ISSUE #758: Tenants collection slug field needs to be editable during create
Currently, the Tenants collection's slug field is always readOnly. This makes it so that you can't set the slug field when creating a new tenant, thereby blocking us from creating new tenants. 

https://www.loom.com/share/82a2d984e54444c6a16b1c02b66c63fd

rchlfryn:
## Description
We cannot edit the tenant slug when we go to make a new tenant. We want to update this to allow a user to add a slug on creation but never be able to edit this after creation.

## Related Issues
Fixes #758 

## Key Changes
- Remove readonly from tenant slug and uses access instead 

## Screenshots / Demo
On creation

Screenshot 2025-12-08 at 22 42 52

---
On edit

Screenshot 2025-12-08 at 22 44 20

### DIFF
--- a/src/collections/Tenants/index.ts
+++ b/src/collections/Tenants/index.ts
@@ -57,7 +57,6 @@ export const Tenants: CollectionConfig = {
       admin: {
         description:
           'Used for subdomains and url paths for previews. This is a unique identifier for a tenant.',
-        readOnly: true,
       },
       index: true,
       required: true,


busbyk (approved):
Incredible

PR #778, opened 2025-12-11
Add NAC media widget as block

### LINKED ISSUE #686: Surface selected media from AFP via media widget
The SNFAC, BTAC, many others use a media widget that pulls selected images from the AFP with their captions onto the site. We may want to use this on various pages and/or the home page. This may be an easy one to integrate.

See:
https://bridgertetonavalanchecenter.org/
https://payetteavalanche.org/
https://www.sawtoothavalanche.com/

Related to #576 but AFP specific

- Add 'media' to NACWidget type enum
- Add a Block that allows embedding this

rchlfryn:
## Description
Adds media widget from [afp-public-widgets](https://github.com/NationalAvalancheCenter/afp-public-widgets/tree/main/src/components/media) as a block

## Related Issues
Fixes #686 

## Key Changes
- Adds `NACMediaWidget` to pages and home page layouts. (not posts or the lexical editor since we cannot control how many times a uses adds the block)
- Uses [conditional blocks](https://payloadcms.com/docs/fields/blocks#conditional-blocks)

## Screenshots / Demo
Screenshot 2025-12-10 at 22 43 32
Screenshot 2025-12-10 at 22 43 58

## Migration Explanation
Yes - adds tables and indices for `nac_media_block`

Do we want to call this AFP or NAC media widget? I tried to copy the already current nomenclature.


src/app/api/[center]/nac-config/route.ts
Comment (busbyk) [inline src/app/api/[center]/nac-config/route.ts:1]: Based on the data returned this is a generic route for avalanche center platforms and the widget config fields. Do we want to name this something more generic? 

Maybe `/[center]/nac-widget-config` or maybe more AFP-generic `/[center]/afp-config/` or just `/[center]/platforms`. Hmmm

src/blocks/NACMediaBlock/Component.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [loading, setLoading] = useState(true)
+
+  const center = typeof tenant === 'object' && tenant !== null ? tenant.slug : null
+
+  useEffect(() => {
+    if (!center) return
+
+    const fetchData = async () => {
+      const response = await fetch(`/api/${center}/nac-media`)
+      const result = await response.json()
```
busbyk: What about using a zod schema and `safeParse` to parse this fetch response with valid types like we do in [`getAllAvalancheCenterCapabilities`](https://github.com/NWACus/web/blob/main/src/services/nac/nac.ts#L121). That way, we wouldn't need to use `any` for the `data` `useState`.

src/blocks/NACMediaBlock/Component.tsx
```diff
@@ -0,0 +1,57 @@
...
+      setData(result)
+      setLoading(false)
+    }
+    fetchData()
+  }, [center])
+
+  const bgColorClass = `bg-${backgroundColor}`
+  const textColor = getTextColorFromBgColor(backgroundColor)
+
+  if (!center) return <div>Error loading media</div>
```
busbyk: I wonder if we should just return `null` here and capture an exception with Sentry. `center` should always exist. I like handling it not existing bc yanno, unexpected errors. Either way I think capturing the exception with Sentry seems like a good idea.

src/blocks/NACMediaBlock/config.ts
```diff
@@ -0,0 +1,37 @@
...
+export const NACMediaBlock: Block = {
+  slug: 'nacMediaBlock',
+  interfaceName: 'NACMediaBlock',
+  labels: {
+    singular: 'NAC Media Block',
+    plural: 'NAC Media Blocks',
+  },
+  imageURL: '/thumbnail/NACMediaBlockThumbnail.jpg',
+  fields: [
+    tenantField(),
```
busbyk: Do we need this here? Could we use `useTenant` in the component instead?

src/blocks/NACMediaBlock/config.ts
```diff
@@ -0,0 +1,37 @@
...
+    colorPickerField('Background color'),
+    {
+      name: 'wrapInContainer',
+      admin: {
+        description:
+          'Checking this will render the block with additional padding around it and using the background color you have selected.',
+      },
+      type: 'checkbox',
+      defaultValue: true,
+    },
```
busbyk: I think we could omit these fields since we'll always want to wrap this in a container if it's only possible to use this as a full width section on Pages, right? 

Like ImageQuote?

src/collections/HomePages/index.tsx
```diff
@@ -166,28 +154,15 @@ export const HomePages: CollectionConfig = {
...
-        TeamBlock,
-      ],
+      blocks: [...DEFAULT_BLOCKS, NACMediaBlock].sort((a, b) => a.slug.localeCompare(b.slug)),
       required: true,
+      filterOptions: ({ data }) => {
+        const layoutBlocks = data?.layout
+        const hasNacMediaBlock = layoutBlocks.some(
+          (block: { blockType: string }) => block.blockType === 'nacMediaBlock',
+        )
+        return hasNacMediaBlock ? DEFAULT_BLOCKS.map((block) => block.slug) : true
```
busbyk: Should we have the check for only a single nac media block here too like there is in the Pages collection?

src/blocks/NACMediaBlock/Component.tsx
rchlfryn: Sounds good - we could remove this completely since I agree center should always be defined, but I just left it for good measure with the Sentry error.

src/blocks/NACMediaBlock/config.ts
rchlfryn: Good call. we don't. I will change it

src/app/api/[center]/nac-config/route.ts
Comment (rchlfryn) [reply src/app/api/[center]/nac-config/route.ts:1]: Oh good call. I made this more generic.

src/blocks/NACMediaBlock/config.ts
rchlfryn: Including this doesn't matter to me, but I don't see why it couldn't be full width if that is what they wanted it to be (besides it might not look good)

src/blocks/NACMediaBlock/Component.tsx
rchlfryn: I hope I did this right. I am not to keen on setting a default empty string for `baseURL` and `version`. Any suggestions?

src/blocks/NACMediaBlock/config.ts
busbyk: Hmm good point. I don't think we have any other blocks that allow for full width (i.e. are not given the `container` class) in Pages blocks. But images would probably be the common case for that. I am skeptical of how it would look compared to other blocks having their max width constrained on larger screens.

src/blocks/NACMediaBlock/Component.tsx
busbyk: I think this is good! The preference would be to not render the NACWidget if there's no `data` so you don't need that logical OR. 

Here's one way you could accomplish that:
- add a hasError boolean for good measure and for making the conditionals more specific/accurate
- return `null` if not loading and there is an error - we don't want to render the padding / container if we're not going to render an error message when there's no data -- unless we want to display and error to the user but I think rendering `null` is the right choice here
- only render NACWidget if `!loading && data`
- capture the /nac-config error using Sentry for good measure

Just a suggestion - tweak as necessary. 

```
export const NACMediaBlockComponent = (props: NACMediaBlockProps) => {
  const { backgroundColor, mode, wrapInContainer } = props
  const { tenant } = useTenant()
  const [data, setData] = useState<NacWidgetConfigurationSchema>()
  const [loading, setLoading] = useState(true)
  const [hasError, setHasError] = useState<boolean>(false)

  const center = typeof tenant === 'object' && tenant !== null ? tenant.slug : null

  useEffect(() => {
    if (!center) return

    const fetchData = async () => {
      const response = await fetch(`/api/${center}/nac-config`)
      const result = await response.json()
      const parsed = nacWidgetConfigurationSchema.safeParse(response)

      if (parsed.success) {
        setData(result)
      } else {
        setHasError(true)
        const errors = parsed.error.message
        Sentry.captureException(`Error parsing /api/${center}/nac-config response: ${errors}`)
      }

      setLoading(false)
    }
    fetchData()
  }, [center])

  const bgColorClass = `bg-${backgroundColor}`
  const textColor = getTextColorFromBgColor(backgroundColor)

  if (!center) {
    Sentry.captureException('NACMediaWidget: center not defined')
    return null
  }

  if (!loading && hasError) return null

  return (
    <div className={cn(!wrapInContainer && bgColorClass)}>
      <div
        className={cn(
          'py-4 w-full text-center',
          textColor,
          wrapInContainer && `${bgColorClass} container`,
        )}
      >
        {(loading || (!data && !hasError)) && <div> Loading...</div>}
        {!loading && data && (
          <NACWidget
            center={center}
            widget={'media'}
            widgetsVersion={data.version}
            widgetsBaseUrl={data.baseUrl}
            mediaMode={mode}
          />
        )}
      </div>
    </div>
  )
}
```

rchlfryn: Great! Implemented in [e4d305a](https://github.com/NWACus/web/pull/778/commits/e4d305a768fe7ac7ac414518a8d82cd21573207d). It was doing its job and erroring so I fixed the schema

busbyk (approved):
LGTM pending the removal of `wrapInContainer` from the config.

PR #781, opened 2025-12-12
Bugfix: enables the use of builtInPages in the navigation

### LINKED ISSUE #752: Cannot use built-in pages in navigation
**Describe the bug**
Built-in pages used as the internal link reference are filtered out of the navigation in [getTopLevelNavItems](https://github.com/NWACus/web/blob/main/src/components/Header/utils.ts#L242). 

**To Reproduce**

1. Add a built-in page as an internal link's "page or post" in the navigation
2. Visit that tenant's home page (or any page)
3. Observe that the link with the built-in page as the internal link's "page or post" does not appear in the nav

**Expected behavior**
The nav item referencing the built-in page should appear in the nav. It should use a url that is _not_ nested under whatever nav item it's nested under. 

**Screenshots**
https://www.loom.com/share/f157303aeeb34c15a854ad0bf53ac73c

**Environment details**

- Desktop
- Chrome

busbyk:
## Description

`builtInPages` weren't being handled correctly in the navigation. They were being filtered out. This allows them in the nav. 

## Related Issues

Fixes #752

## Key Changes

- Enables the use of built-in pages in the navigation
- Uses a built-in page for the weather stations link in the seed script

## Screenshots / Demo

https://www.loom.com/share/134e95648cae46cf8c2dc75fc04fc5e4

## Changes needed after merge

- [ ] Update all centers' weather stations nav link to use their built-in page
- [ ] Communicate that the local accidents can be a built-in page for Sierra / add to their nav. Won't be a significant difference for them but might as well change it.

### DIFF
--- a/src/components/Header/utils.ts
+++ b/src/components/Header/utils.ts
@@ -163,6 +163,7 @@ export function convertToNavLink(
 ): NavLink | undefined {
   let linkLabel: string | undefined = link.label || undefined
 
+  // external links
   if (link.type === 'external' && link.url) {
     invariant(linkLabel, `Label not set for external link with url ${link.url}`)
 
@@ -174,24 +175,14 @@ export function convertToNavLink(
     }
   }
 
-  if (
-    link.type === 'internal' &&
-    link.reference &&
-    (link.reference.relationTo === 'pages' ||
-      link.reference.relationTo === 'posts' ||
-      link.reference.relationTo === 'builtInPages')
-  ) {
+  // internal page/post/builtInPage reference links
+  if (link.type === 'internal' && link.reference) {
     const reference = link.reference
 
     invariant(
       typeof reference.value !== 'number',
       `Link reference.value is a number. Depth not set correctly on navigations collection query.`,
     )
-
-    // Do not render documents in draft state (builtInPages don't have _status)
-    if ('_status' in reference.value && reference.value._status === 'draft') {
-      return undefined
-    }
 
     if (
       !linkLabel &&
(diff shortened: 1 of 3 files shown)


src/components/Header/utils.ts
```diff
@@ -204,13 +207,24 @@ export function convertToNavLink(
...
-    let url =
-      link.reference.relationTo === 'pages'
-        ? `/${reference.value.slug}`
-        : `/blog/${reference.value.slug}`
+    let url: string
+    if (link.reference.relationTo === 'builtInPages' && 'url' in reference.value) {
+      url = normalizePath(reference.value.url, { ensureLeadingSlash: true })
+    } else if (link.reference.relationTo === 'posts' && 'slug' in reference.value) {
+      url = `/blog/${reference.value.slug}`
+    } else if ('slug' in reference.value) {
```
rchlfryn: do we want to add `link.reference.relationTo === 'pages' &&` here as well? That was our initial check and it is not here anymore.

This might also cover external URLs so I could see the case for keeping it out

src/endpoints/seed/navigation.ts
```diff
@@ -45,6 +46,34 @@ export const navigationSeed = (
     }
   }
 
+  const builtInPageLink = ({
```
rchlfryn, line 49: 👏🏼 this was needed

src/endpoints/seed/navigation.ts
```diff
@@ -45,6 +46,34 @@ export const navigationSeed = (
...
+    url,
+    label,
+  }: {
+    url: string
+    label?: string
+  }): NonNullable<NonNullable<Navigation['donate']>['link']> => {
```
rchlfryn: I think this is the wrong type? `Navigation['donate']>['link']`

src/endpoints/seed/navigation.ts
```diff
@@ -45,6 +46,34 @@ export const navigationSeed = (
...
+
+    const builtInPage = builtInPages[tenant.slug][url]
+
+    return {
+      type: 'internal',
+      reference: {
+        value: builtInPage.id,
+        relationTo: 'builtInPages',
+      },
+      label: label || builtInPage.title,
```
rchlfryn, line 73: Since label is optional, do we want to add a default assignment here? It might be nice to test the edge case.

src/components/Header/utils.ts
```diff
@@ -176,7 +177,9 @@ export function convertToNavLink(
   if (
     link.type === 'internal' &&
     link.reference &&
-    (link.reference.relationTo === 'pages' || link.reference.relationTo === 'posts')
+    (link.reference.relationTo === 'pages' ||
+      link.reference.relationTo === 'posts' ||
+      link.reference.relationTo === 'builtInPages')
```
rchlfryn: It is interesting we allow and combine all of these type options only to separate them later in 210. Should we just separate them form here? I am not sure what is cleaner.

rchlfryn (changes requested):
A few thoughts.. good call on this PR

src/components/Header/utils.ts
busbyk: Good call. I like being more specific here. I'll add that back.

src/endpoints/seed/navigation.ts
busbyk: Ope, yea double nested NonNullables. We need this one NonNullable though because Navigation['donate'] is possibly undefined.

src/endpoints/seed/navigation.ts
busbyk: I.e. if builtInPage.title is null?

src/components/Header/utils.ts
busbyk: What do you think about this refactor? [1f91195](https://github.com/NWACus/web/pull/781/commits/1f91195c357649c809685caed3d12eebc721bc99)

Tried to make this a little bit easier to read / cleaner.

src/endpoints/seed/navigation.ts
rchlfryn: I was just saying it is optional on line 54 so it seems weird to have a default here but I get why you did it

src/components/Header/utils.ts
rchlfryn: Awesome!

rchlfryn (approved):

PR #784, opened 2025-12-12
Do not reformat importmap.js

### LINKED ISSUE #783: Avoid constant reformatting of importMap.js
It's annoying during development for `src/app/(payload)/admin/importMap.js` to be regenerated by Payload in a different order than we had previously committed it. 

The issue here is that our `generate:importmap` script runs `prettify` after which re-orders imports. This differs from how Payload generates it without our formatting step. If we removed `prettify` after that script we should be able to avoid this unnecessary formatting of that file. 

We'll just want to make sure that our CI won't try to re-format that file and complain about it. We may need to ignore that file from formatting.

busbyk:
## Description

`src/app/(payload)/admin/importMap.js` appearing in our git changes during local development has been a minor annoyance for us for a while now. I think we've both just been ignoring it. The reason it appears in changes is because `payload generate:importmap` was generating the file and then we had `&& pnpm prettify` which was reformatting the file using prettier which re-ordered imports due to our plugins. 

We committed this prettier-formatted version. And so when Payload generates a new importmap during local development, it wasn't get automatically reformatted with prettier, resulting in a git diff. 

There's no reason to reformat this file. It's autogenerated. 

## Related Issues

Fixes #783 

## Key Changes

- Ignore `src/app/(payload)/admin/importMap.js` from prettier
- Remove `&& pnpm prettify` from `generate:importmap`
- Remove `&& pnpm prettify` from `generate:types` (we had already added the types file to our `.prettierignore` so this formatting step was doing nothing anyways and just making the script take longer because it was checking all files for formatting)
- Commits `src/app/(payload)/admin/importMap.js` in the format that Payload generates it in


rchlfryn (approved):
🎁  🎊  👍🏼

PR #788, opened 2025-12-13
Use OG image API route instead of opengraph-image and twitter-image files

### LINKED ISSUE #603: Bug: OG images not loading on center-scoped routes
**Describe the bug**

OG images aren't loading correctly for center-scoped routes. 

**To Reproduce**

Cannot reproduce locally, production only. 

Inspect the meta tags for a center-scoped route in a deployed environment. 

If you remove the /{center slug}/ url part you get a `Failed to generate OG image` error:
Image

**Expected behavior**
The OG images should load correctly. 

**Assessment**

I suspect there are two problems at play here. 

**Screenshots**

Image

Image

Local:
Image
Aside: check that we're generating OG images with the right dimensions.

**Environment details**

- Desktop
- Chrome

**Existential musings**
Do you remember how great stumbleupon was back in the day?

busbyk:
## Description

During the build the center was getting prepended to the urls generated by these "special" files in Next.js. So when visiting these urls on a production build you would essentially have two center paths instead of one after the rewrite. 

Let's say you get the following url generated for a route's opengraph-image file: `/sac/opengraph-image-zct1d9?51f04e5af21d0f21`. Well when visiting this in production you would get a URL: `https://sac.avy-fx.org/sac/opengraph-image-zct1d9?51f04e5af21d0f21` which behind the scenes the middleware would then rewrite to: `https://avy-fx.org/sac/sac/opengraph-image-zct1d9?51f04e5af21d0f21` which would 404. 

I couldn't think of a way to avoid this since the opengraph|twitter-image files are colocated with the route files in their center-scoped routes so technically this was "right" according to Next.js.

So this uses and API route to generate images instead. This is realistically much more flexible anyways. 

There was a second problem: we were using `fs` to read the font file into a buffer but we can't do this in Vercel serverless environments because that font file isn't bundled with these serverless functions. So changed this to include the font file in our public folder and fetch the font file from their in the API route. 

## Related Issues

Fixes #603 

## Screenshots / Demo

CleanShot 2025-12-12 at 17 31 52@2x


consistent-type-assertions.txt
```diff
@@ -33,6 +33,5 @@ src/fields/slug/SlugComponent.tsx
 src/middleware.ts
 src/services/vercel.ts
 src/utilities/findDocumentsWithBlockReferences.ts
-src/utilities/generateOGImage.tsx
```
rchlfryn, line 36: 🥳

src/app/api/[center]/og/route.tsx
```diff
@@ -209,7 +198,7 @@ export async function generateOGImage({
         fonts: [
           {
             name: 'Lato',
-            data: latoBold,
+            data: fontData,
             style: 'normal',
             weight: 600,
```
Comment (rchlfryn) [inline src/app/api/[center]/og/route.tsx]: Oh interesting. Isn't bold `800`? I am surprised this works with the new import. Should we move semi-bold to the public folder instead?

rchlfryn (approved):
LG - one minor question. Will conditionally approve

src/app/api/[center]/og/route.tsx
Comment (busbyk) [reply src/app/api/[center]/og/route.tsx]: Oh very good point. Bold is 700. I'll update. Good catch!

PR #789, opened 2025-12-13
A3 updates (docs, embed generator, logo)

### LINKED ISSUE #747: Add A3 logo somewhere to A3 embeds
If embed is in xyz provider site or Skidoo, the user knows the experience is brought to them by A3

### LINKED ISSUE #772: A3 Embed documentation
I just need some basic documentation on how to embed A3 courses for 

- ACs
- Providers
- A3/avalanche.org
- something like Skidoo

I can polish from there.

busbyk:
## Description

Three updates for A3 courses:
1. Docs: (events and a3 embeds docs)
2. An embed generator view in the admin panel
3. A3 logo and blurb added to each embed

## Related Issues

Fixes #747 
Fixes #772 

## Key Changes

- Adds events docs
- Adds A3 embeds docs
- Adds an embed generator view to the admin panel
  - There is no link to this view. We plan to add a link to it in our public Notion docs.
  - All options available to embeds are configurable here + the option to include a resizer script or use a fixed height
- Adds the A3 logo and a text blurb above each embed

## Screenshots / Demo

https://www.loom.com/share/be070d20b35348fb99ab0fd77f8ed865


src/app/(embeds)/embeds/courses/page.tsx
```diff
@@ -5,6 +5,7 @@ import { getCoursesStates } from '@/utilities/queries/getCoursesStates'
 import { getProviders } from '@/utilities/queries/getProviders'
 import Script from 'next/script'
 import { createLoader, parseAsBoolean, parseAsString, SearchParams } from 'nuqs/server'
+import { A3Banner } from '../../../../components/A3Banner'
```
rchlfryn: ⛏️ 
```suggestion
import { A3Banner } from '@/components/A3Banner'
```

src/app/(embeds)/embeds/providers/page.tsx
```diff
@@ -11,6 +11,7 @@ import config from '@/payload.config'
 import Script from 'next/script'
 import { createLoader, parseAsString, SearchParams } from 'nuqs/server'
 import { getPayload } from 'payload'
+import { A3Banner } from '../../../../components/A3Banner'
```
rchlfryn: ⛏️ 
```suggestion
import { A3Banner } from '@/components/A3Banner'
```

src/components/A3Banner.tsx
```diff
@@ -0,0 +1,21 @@
...
+      <div className="flex-shrink-0">
+        <Image
+          src="/assets/a3-logo.png"
+          alt="A3 Logo"
+          width={1361}
+          height={444}
+          className="w-auto h-12"
+        />
+      </div>
+      <p className="text-xs text-muted-foreground">
```
rchlfryn, line 15: Will `text-muted-foreground` always be defined? If not, maybe we jsut use a grey or something instead

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
```diff
@@ -0,0 +1,354 @@
...
+  const queryString = params.toString()
+  const src = `${baseUrl}/embeds/${type}${queryString ? '?' + queryString : ''}`
+  const iframeId = `avy-web-embed-${type}`
+
+  // For auto height, use 0 as initial height (iframe-resizer will adjust)
+  // For fixed height, use the user-specified value
+  const height = options.heightMode === 'auto' ? '0' : `${options.fixedHeight}px`
+
+  if (options.heightMode === 'auto') {
+    return `<iframe
```
rchlfryn, line 110: ⛏️  to the max... My OCD brain wants these 2 scripts nicely indented...

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
rchlfryn, line 1: This is awesome!

docs/a3-embeds.md
rchlfryn, line 1: Are these (this and `events-and-courses.md)` supposed to be in Notion or here or both?

src/components/A3Banner.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export function A3Banner() {
+  return (
+    <div className="flex flex-col md:flex-row justify-between md:items-end gap-3 mb-4 border-b pb-3">
+      <div className="flex-shrink-0">
+        <Image
+          src="/assets/a3-logo.png"
+          alt="A3 Logo"
+          width={1361}
+          height={444}
```
rchlfryn, line 11: Interesting size. Is this just the natural size? we might want to update it since it is not/will never be 1361px. Or can we just remove this?

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
```diff
@@ -0,0 +1,354 @@
...
+  types: [],
+  providers: [],
+  states: [],
+  affinityGroups: [],
+  modesOfTravel: [],
+  startDate: undefined,
+  endDate: undefined,
+}
+
+function formatDateForParam(date: Date | undefined): string {
```
rchlfryn, line 77: I know you recently did a bunch of timezone work, does that need to be done here as well? Or are we getting back the correct date from the server?

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
```diff
@@ -0,0 +1,354 @@
...
+            <div>
+              <FieldLabel label="Start Date" />
+              <DatePicker
+                value={options.startDate}
+                onChange={(date) => updateOption('startDate', date)}
+              />
+            </div>
+            <div>
+              <FieldLabel label="End Date" />
+              <DatePicker
```
rchlfryn, line 332: Do we want any light error handling on this? I was able to select an end date that was before the start date.

rchlfryn (commented):
Very nit picky review from me. Overall looks great. Just a few questions

src/components/A3Banner.tsx
busbyk: Yea the `a3` theme inherits from our `:root` variables so we're good here: https://github.com/NWACus/web/blob/a3-updates/src/app/(frontend)/colors.css#L20

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
busbyk: haha I feel ya but we need to format it this way. If we add indentation in the code we get extra indentation in the generated code in the UI:
CleanShot 2025-12-13 at 13 06 42@2x
CleanShot 2025-12-13 at 13 06 36@2x

docs/a3-embeds.md
busbyk: I think it makes sense to have different versions in both places. These are slightly more technical docs than how I'd write this for the Notion docs probably. Once this is approved / decided on I can copy over to Notion. 

Unless you think we could just linked to these files from Notion. This isn't _that_ technical.

src/components/A3Banner.tsx
busbyk: This is the intrinsic size of the image. I believe this is the correct usage of next/image based on: https://nextjs.org/docs/15/app/api-reference/components/image#width-and-height

Alternatively, we could use `fill` but I think the dynamic image optimization likes the intrinsic size for best performance.

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
busbyk: Yea good thought! This implicitly uses the user's browser timezone to generate a date-only string. In this case that's `yyyy-MM-dd`. I _think_ it's OK to leave this as is because it's not being _stored_ anywhere, this is a static value provided as the date for the iframe embed which is always client-side only so it will always filter based on the user's browser's timezone. Thoughts?

This is making me realize that this is a mismatch between how we represent the date in our filter components which is `MM-dd-yyyy`. This actually still works since the `Date` constructor accept both formats. But I'll update to match just for consistency.

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
busbyk: Good call! I added a toast notification onChange if either date is invalid. It still sets the value - just alerts the user it's invalid. Since there's no form submit or anything I think this makes sense.

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
rchlfryn: Stupid payload 😆

docs/a3-embeds.md
rchlfryn: That makes sense to keep the notion ones less technical. I think it's a good base for the A3 stuff

src/components/A3Banner.tsx
rchlfryn: No change needed. I was just curious

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
rchlfryn: I think it is fine to use the browsers timezone as you have it.

rchlfryn (approved):

src/views/EmbedGenerator/EmbedGeneratorForm.tsx
busbyk: Nah nah, just a javascript template literals!

PR #791, opened 2025-12-13
Clean up - Vercel, issue template, prettier for dependabot

### LINKED ISSUE #790: Clean up Vercel
I already created a one time script to clean up all old data.

Now we need to add to the clean up script to clean up environments things in vercel after a PR is merged:
* New deployment
* Wild card alias
    - BRANCH_NAME.preview.avy-fx.org
* ID alias
    - avy-HASH-nwac.vercel.app
- Environment variables    
    ```
    DATABASE_URI
    DATABASE_AUTH_TOKEN
    PAYLOAD_SEED_PASSWORD
    NEXT_PUBLIC_ROOT_DOMAIN
    NEXT_PUBLIC_SERVER_URL
    SERVER_DOMAIN
    ```

* Update script to NOT deploy on merge bot

rchlfryn:
## Description
- Adds more to `cleanup.yml` to clean up Vercel once a PR is merged
- Does not deploy for `merge-bot` 
- Hides the default issue template
- Remove `prettier-for-dependabot` workflow

## Related Issues
Fixes #790 

## Key Changes
- Does not deploy for `merge-bot` 
- Cleans up vercel by when a branch is merged by:
  - Removing environment variables
  - Removing deployments
  - Removing aliases from those deployments 

## Notes
I could not find a good way to manually trigger this. I did not want to change it to on `push` because then it would be a race case and not controlled.


.github/workflows/cleanup.yaml
```diff
@@ -1,12 +1,10 @@
 name: cleanup
 on:
   delete:
-    branches-ignore:
-      - 'main'
-      - 'release'
```
busbyk, line 6: Not that we'll ever intentionally delete these but why remove this? The if check on fromJson does the same thing?

busbyk (approved):
Looks right to me. Just one question to confirm.

.github/workflows/cleanup.yaml
rchlfryn: This was improperly formatted yaml, so I moved the check to line 7. 🤷🏼‍♀️

Issue #710, opened 2025-11-24
Slug field enhancements


busbyk:
I believe it would be a significant UX improvement if a user could omit the required slug field for a document and have it auto-generated based on the `fieldToUse` arg instead of being warned that the slug field must be set and requiring the user to generate it. We have the ability to generate the slug field value for them beforeChange so we might as well eliminate this possible friction for users. 

We can even increment the slug if there is a duplicate slug for the same collection. 

This would be particularly useful for the Courses collection since slugs have to be globally unique, not just unique across a single tenant. 

We made some enhancements to the slug field for the Courses and Providers collection to allow the use of the slug field without a tenant field: https://github.com/NWACus/web/pull/702/files#diff-56a6e2e43d288073a722e7cdae9a9f2c36a41c7af6b78979317df33244fbed1f

busbyk:
Note: Payload now has it's own `slugField` exported from their UI package: https://github.com/payloadcms/payload/blob/main/packages/ui/src/fields/Slug/index.tsx

Worth considering if we should use that over our custom slug field, originally from the website template.

busbyk:
Let's make sure we trim field values before they are slugified.

rustynwac:
Is this a struggle for users?

rchlfryn:
Payload just fixed a bug with their slug field https://github.com/payloadcms/payload/issues/14938#issuecomment-4147219675

PR #806, opened 2025-12-18
Fix course bugs

### LINKED ISSUE #805: Courses - update filter, fix 405 err, slug/title & filters
### Description
On production we had a few blank courses added as drafts. A couple errors sprouted from trying to remove them:
* Check all filters to make sure we are not using draft data (original issue from slack)
* A 405 is thrown when trying to bulk delete
* When filtering, blank options are shown
  * Image
* The slug generator adds a trailing `-` if there is a trailing space in the title
  * Hook to remove trailing space on title
  * Fix generator to not include trailing spaces

rchlfryn:
## Description
* Fixes course filters to not use `draft` data
* Fixes 405 error caused when bulk deleting courses (should probably file payload issue)
* Add `trim()` to `formatSlug` to not have trailing `-` on slug generation

## Related Issues
Fixes #805 

## Key Changes
For 405 error
* Bulk delete operations on courses were returning 405 Method Not Allowed, while individual course deletions worked fine. Individual deletes probably worked because they used a different route pattern (`/api/courses/[id]/route.ts`) which had proper `DELETE` handling. 
* We added a custom file `/api/courses/route.ts` which only defined a `GET` handler. This intercepted REST requests to `/api/courses` before reaching Payload's auto-generated endpoints. 
* When bulk delete sent `DELETE /api/courses?where[...]`, Next.js returned `405` because the `DELETE` method wasn't defined in the custom route file.
* Solution: I moved custom file  `/api/courses` to a different endpoint to avoid shadowing Payload's REST API

Do we want to add a hook to trim all titles?


busbyk:
> ## Key Changes
> For 405 error
> 
> * Bulk delete operations on courses were returning 405 Method Not Allowed, while individual course deletions worked fine. Individual deletes probably worked because they used a different route pattern (`/api/courses/[id]/route.ts`) which had proper `DELETE` handling.
> * We added a custom file `/api/courses/route.ts` which only defined a `GET` handler. This intercepted REST requests to `/api/courses` before reaching Payload's auto-generated endpoints.
> * When bulk delete sent `DELETE /api/courses?where[...]`, Next.js returned `405` because the `DELETE` method wasn't defined in the custom route file.
> * Solution: I moved custom file  `/api/courses` to a different endpoint to avoid shadowing Payload's REST API

🤦‍♂️ makes sense that custom route would conflict with Payload's - didn't consider that. Thanks for fixing!

> ## Future enhancements / Questions
> Do we want to add a hook to trim all titles?

I have [an issue](https://github.com/NWACus/web/issues/710) for general enhancements to our re-usable slug field - I've added a note to that to make sure field values are trimmed before their slugified.

PR #813, opened 2025-12-19
Remove autosave on all collections

### LINKED ISSUE #779: Remove autosave
Autosave is very wonky on the frontend. There is a lot of waiting before you save and it sometimes even deletes data. We should remove it and use the `Save as draft` button instead.

rchlfryn:
## Description
Remove autosave on all collections

## Related Issues
Fixes #779 

## Migration Explanation
Simple migration

### DIFF
--- a/docs/revalidation.md
+++ b/docs/revalidation.md
@@ -233,9 +233,6 @@ All revalidation hooks respect the `context.disableRevalidate` flag to prevent r
 
 We should always respect this in new revalidation hooks.
 
-### Autosave Handling
-Page and Post hooks skip revalidation when `query.autosave === 'true'` to avoid unnecessary revalidation during draft editing.
-
 ## Guidance for Writing Revalidation Functions
 
 ### Reference Collections
--- a/src/app/(frontend)/[center]/[...segments]/page.tsx
+++ b/src/app/(frontend)/[center]/[...segments]/page.tsx
@@ -27,7 +27,7 @@ export async function generateStaticParams() {
       tenant: true,
       slug: true,
     },
-    // Need where clause to ignore autosave bug (https://github.com/NWACus/web/pull/204)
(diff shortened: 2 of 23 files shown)


busbyk (commented):
We have some [`query.autosave`](https://github.com/search?q=repo%3ANWACus%2Fweb%20query.autosave&type=code) checks in hooks that we'll want to remove in this PR too.

busbyk (approved):
Looks good. We should educate users that they can hit ⌘ + s to save a draft - that's pretty helpful when using the preview pane to see changes quickly. 

I don't think that we should remove the where clause - I think it's good to have regardless of autosave being turned on or not but I think I'd be in favor of removing these  comments. Or if you think they should stay so we don't remove that where clause, we could change them to explain that "We only want to statically generate for published documents". 

https://github.com/NWACus/web/blob/remove-autosave/src/app/(frontend)/%5Bcenter%5D/%5B...segments%5D/page.tsx#L30
https://github.com/NWACus/web/blob/remove-autosave/src/app/(frontend)/%5Bcenter%5D/%5Bslug%5D/page.tsx#L32
https://github.com/NWACus/web/blob/remove-autosave/src/app/(frontend)/%5Bcenter%5D/blog/%5Bslug%5D/page.tsx#L29

rchlfryn:
> Looks good. We should educate users that they can hit ⌘ + s to save a draft - that's pretty helpful when using the preview pane to see changes quickly.
> 
> I don't think that we should remove the where clause - I think it's good to have regardless of autosave being turned on or not but I think I'd be in favor of removing these comments. Or if you think they should stay so we don't remove that where clause, we could change them to explain that "We only want to statically generate for published documents".

I think keeping there where clause is important as a user can still make a draft. I can update the comment to remove the autosave explanation.

PR #823, opened 2025-12-22
Debug script - update vercel branch call

rchlfryn:

### DIFF
--- a/.github/workflows/development.yaml
+++ b/.github/workflows/development.yaml
@@ -134,13 +134,17 @@ jobs:
           RESPONSE=$(curl -s -H "Authorization: Bearer ${{ secrets.VERCEL_TOKEN }}" \
             "https://api.vercel.com/v6/deployments?projectId=${{ secrets.VERCEL_PROJECT_ID }}&meta-gh-branch=main&limit=100")
 
+          echo "[DEBUG] Full API response:"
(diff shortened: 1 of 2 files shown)


.github/workflows/development.yaml
```diff
@@ -129,11 +130,17 @@ jobs:
...
           echo "[INFO] Finding all dev deployments for main branch"
 
-          # Get all deployments for main branch, sorted by creation time (newest first)
-          DEPLOYMENTS=$(vercel deployments --json --token=${{ secrets.VERCEL_TOKEN }} 2>/dev/null | jq -r '.deployments[] | select(.meta.githubCommitRef=="main") | .uid' || echo "")
+          # Debug: show raw output first
+          echo "[DEBUG] Raw deployments output:"
+          vercel deployments --token=${{ secrets.VERCEL_TOKEN }} | head -20
+
+          # Get all deployments for main branch - improved filtering
+          DEPLOYMENTS=$(vercel deployments --token=${{ secrets.VERCEL_TOKEN }} | awk '/main/ {print $1}' || echo "")
```
stevekuznetsov: Prefer output JSON and `jq` for operating over structured data to `awk` if you can, less likely to break later in some inscrutable and confusing way when vercel CLI decides some other format is more human-readable

PR #829, opened 2025-12-30
collections/Documents: allow other MIME types

stevekuznetsov:
One of our users has an urgent use-case for uploading XML and PHP files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ node_modules
 
 # Payload default media upload directory
 public/media/
+public/documents
 
 public/robots.txt
 public/sitemap*.xml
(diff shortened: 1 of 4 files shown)


andyanderso:
Thanks guys!

andyanderso:
Thanks, everyone. The KML file  works great. 

The .php file still throws an error when I try to upload it. PHP is a bit trickier, since the server would have to run the PHP file to generate the JSON that we use to create the wx forecast. 
Here is how it currently works:

- The AFP has a button that says load custom wx table. 
- When we click that button, it pings https://archive.sierraavalanchecenter.org/sites/all/libraries/avalanche.org-API-Client/SAC_WxConfigJSON.php
- The php file then pulls info from the NWS api for the current mountain weather forecast, parses it, and outputs a JSON file. 
- The AFP uses that JSON file to build and populate our custom weather table. 

I'm not sure if the avy-web server is configured to support PHP. We could rewrite the script in another language if needed, but I don't have time to do so during the forecasting season. For now, I have it hosted in two places. Surely they won't both go down at the same time :-)

PR #833, opened 2026-01-06
Tenant not defined on post tag

### LINKED ISSUE #820: Cannot add Post Tag slug that is being used in another tenant
**Describe the bug**
I am trying to create the slug `news` for the PAC site and it wont let me

**To Reproduce**

1. Go to Posts
2. Add Tag News with Slug news
3. See error

**Expected behavior**
I would expect to have the same slug for diff tenants 

**Screenshots**

Image

**Environment details**
- Prod
- 
**Existential musings**
Mexico really likes to hide their traffic lights. I ran 3

rchlfryn:
## Description
The problem is, upon tag creation, tenant is not passed to `ensureUniqueSlug` in the `data` or `originalDoc` props.
`data : { slug: 'category', title: 'Category' }`
`originalDoc: {}`

I checked other collections that are using `ensureUniqueSlug` and `tenant` is being passed in both `data` or `originalDoc` props so I am not sure why `tags` is any different.

## Related Issues
Fixes #820 

## Key Changes
Changed to use `getTentantFromCookie` instead of using `data` or `originalDoc`. I don't feel great about this change, since I think tenant should be defined on `data` or `originalDoc` but I figured it was a better fix than debugging payload create method. `getTentantFromCookie` should always be defined.

- Upgrade payload to include their fix
- truly debug why `tenant` is not passed on `tags` initial create request

### DIFF
--- a/src/fields/slug/ensureUniqueSlug.ts
+++ b/src/fields/slug/ensureUniqueSlug.ts
@@ -1,3 +1,4 @@
+import { getTenantFromCookie } from '@/utilities/tenancy/getTenantFromCookie'
 import type { FieldHook, Where } from 'payload'
 
 import { ValidationError } from 'payload'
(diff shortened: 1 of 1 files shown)


llaz00:
Does this work with `tenantsSlug `set to other value than `tenant` in `payload.config.ts`?

rchlfryn:
No longer needed.. fixed in #838

PR #835, opened 2026-01-06
Fix duplicate for all collections

### LINKED ISSUE #827: Slug creation prevents duplicating any collection
**Describe the bug**
I go to clone and event and I get an error: The following field is invalid: slug

**To Reproduce**

1. Go to an event
2. Click duplicate

**Expected behavior**
I'd expect it to clone successfully

**Screenshots**

Image

**Environment details**
- Desktop prod

**Existential musings**
"Let me know if I can do anything", "is there anything you need?", etc are not very helpful gestures

rchlfryn:
## Description
Duplicating a collection fails because the slug isn't unique, and Payload's `unique: true` validator doesn't work correctly in multi-tenant environments. We added custom validation to `ensureUniqueSlug` to handle this. It looks like Payload is working on a solution so we can remove our workaround eventually and use `unique: true` again. https://github.com/payloadcms/payload/issues/14938

Until then, we added `getIndexedTitleAndSlug` when making `duplicatePageToTenant` #322 which will increment the slug and title based on the number of documents found in the current tenant.. I checked payload's docs and they handle duplication like this:
> By Default, unique and required text fields Payload will append "- Copy" to the original document value.

I think we should do the same rather than increment and hopefully they will fix the `unique` prop issue.

## Related Issues
Fixes #827 

## Key Changes
Adds  `- Copy` to title and `-copy` to slugs on duplication

### DIFF
--- a/src/components/Header/utils.ts
+++ b/src/components/Header/utils.ts
@@ -161,7 +161,7 @@ export function convertToNavLink(
   },
   parentItems?: string[],
 ): NavLink | undefined {
-  let linkLabel: string | null | undefined = link.label || undefined
+  let linkLabel: string | undefined = link.label || undefined
 
   // external links
   if (link.type === 'external' && link.url) {
--- a/src/components/Header/utils.ts
+++ b/src/components/Header/utils.ts
@@ -161,7 +161,7 @@ export function convertToNavLink(
   },
   parentItems?: string[],
 ): NavLink | undefined {
-  let linkLabel: string | undefined = link.label || undefined
+  let linkLabel: string | null | undefined = link.label || undefined
 
   // external links
   if (link.type === 'external' && link.url) {
--- a/src/fields/slug/ensureUniqueSlug.ts
+++ b/src/fields/slug/ensureUniqueSlug.ts
@@ -1,6 +1,6 @@
 import type { FieldHook, Where } from 'payload'
 
-import { ValidationError } from 'payload'
+import { APIError } from 'payload'
 import invariant from 'tiny-invariant'
 
 export const ensureUniqueSlug: FieldHook = async (props) => {
@@ -57,14 +57,10 @@ export const ensureUniqueSlug: FieldHook = async (props) => {
   })
 
   if (duplicateDocumentsRes.docs.length > 0 && req.user) {
-    throw new ValidationError({
-      errors: [
-        {
-          message: `A ${collection.labels.singular} with the slug ${value} already exists. Slug must be unique${collectionHasTenantField ? ' per avalanche center' : ''}.`,
-          path: 'slug',
-        },
-      ],
-    })
+    throw new APIError(
+      `A ${collection.labels.singular} with the slug "${value}" already exists. Slug must be unique${collectionHasTenantField ? ' per avalanche center' : ''}.`,
+      400,
+    )
   }
 
   return value
--- a/src/fields/slug/ensureUniqueSlug.ts
+++ b/src/fields/slug/ensureUniqueSlug.ts
@@ -11,6 +11,9 @@ export const ensureUniqueSlug: FieldHook = async (props) => {
     'Collection config missing in ensureUniqueSlug FieldHook. This is likely a misuse of the slug field.',
   )
 
+  // Don't validate on duplicate
+  if (req.url?.includes('duplicate')) return value
+
   const tenantField = collection.fields.find((field) => 'name' in field && field.name === 'tenant')
   const collectionHasTenantField = !!tenantField
   const tenantFieldRequired =
--- a/src/fields/slug/ensureUniqueSlug.ts
+++ b/src/fields/slug/ensureUniqueSlug.ts
@@ -11,9 +11,6 @@ export const ensureUniqueSlug: FieldHook = async (props) => {
     'Collection config missing in ensureUniqueSlug FieldHook. This is likely a misuse of the slug field.',
   )
 
-  // Don't validate on duplicate
-  if (req.url?.includes('duplicate')) return value
-
   const tenantField = collection.fields.find((field) => 'name' in field && field.name === 'tenant')
   const collectionHasTenantField = !!tenantField
   const tenantFieldRequired =
--- a/src/fields/slug/index.ts
+++ b/src/fields/slug/index.ts
@@ -2,7 +2,7 @@ import { ensureUniqueSlug } from '@/fields/slug/ensureUniqueSlug'
 import { validateSlug } from '@/utilities/validateSlug'
 import { FieldHook, TextField } from 'payload'
 
-export const getDuplicateSlug: FieldHook = async ({ value }) => {
+export const setDuplicateSlug: FieldHook = async ({ value }) => {
   if (!value || typeof value !== 'string') {
     return value
   }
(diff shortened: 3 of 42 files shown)


src/fields/slug/ensureUniqueSlug.ts
```diff
@@ -11,6 +11,9 @@ export const ensureUniqueSlug: FieldHook = async (props) => {
     'Collection config missing in ensureUniqueSlug FieldHook. This is likely a misuse of the slug field.',
   )
 
+  // Don't validate on duplicate
+  if (req.url?.includes('duplicate')) return value
```
busbyk: I don't think it's a good idea to skip this check since we still don't want duplicate slug values. We just want to append something to the slug value like we do in the `src/collections/Pages/endpoints/duplicatePageToTenant.ts` server action. 

For example, if we skip checking that the slug is unique on duplicate and then the user doesn't manually change the slug, we'll have two documents with the same slug...

What about using [the `beforeDuplicate` field hook](https://payloadcms.com/docs/hooks/fields#beforeduplicate) to use `getIndexedTitleAndSlug` like you suggested in this PR description (or similar logic).

busbyk (changes requested):

src/fields/slug/ensureUniqueSlug.ts
rchlfryn: > For example, if we skip checking that the slug is unique on duplicate and then the user doesn't manually change the slug, we'll have two documents with the same slug...

~This will not happen. The user will will be forced to update the slug upon saving the newly duplicated page since the validation will catch this. The problem is the creation is failing because of the validation. I should have explained this in the description. I updated it.~

> What about using [the beforeDuplicate field hook](https://payloadcms.com/docs/hooks/fields#beforeduplicate) to use getIndexedTitleAndSlug like you suggested in this PR description (or similar logic).

Let me try this. I think it will work.. Good callout

rchlfryn:
@busbyk Completely rewrote the description for this PR so take a look.

src/fields/title/index.ts
```diff
@@ -0,0 +1,22 @@
+import { FieldHook, TextField } from 'payload'
+
+export const getDuplicateSlug: FieldHook = async ({ value }) => {
```
busbyk: ⛏️can we call this setDuplicateTitle

src/fields/slug/index.ts
```diff
@@ -1,6 +1,14 @@
 import { ensureUniqueSlug } from '@/fields/slug/ensureUniqueSlug'
 import { validateSlug } from '@/utilities/validateSlug'
-import { TextField } from 'payload'
+import { FieldHook, TextField } from 'payload'
+
+export const getDuplicateSlug: FieldHook = async ({ value }) => {
```
busbyk: ⛏️Can we call this setDuplicateSlug

src/components/Header/utils.ts
```diff
@@ -161,7 +161,7 @@ export function convertToNavLink(
   },
   parentItems?: string[],
 ): NavLink | undefined {
-  let linkLabel: string | undefined = link.label || undefined
+  let linkLabel: string | null | undefined = link.label || undefined
```
busbyk: Why add `null` here?

src/payload-types.ts
```diff
@@ -319,7 +319,7 @@ export interface Page {
   /**
    * The main heading for this page. This appears in the browser tab, search results, and as the page heading. Keep it descriptive and under 60 characters for best SEO results.
    */
-  title: string;
+  title?: string | null;
```
busbyk: Hmm, do we want this to be the case? I think we would prefer this to still be `title: string` not possibly `undefined` or `null` since title is required for Pages right? 

I'm guessing this is because `required` is set in the field config based on the value of `isRequired` in the props of the `titleField` function. 

Maybe if we used function overloads for that `titleField` function, Payload would generate the expected types. 

Although it's weird that only the Page collection's title attribute's type changed... 🤔

busbyk (commented):
I like the `- Copy` for title and `-copy` for slug behavior. It's nice to mimic Payload's behavior here. 

I did have one thought though: If you try to copy the same page twice you still get a "The following field is invalid: slug" error on duplicate which is pretty unhelpful to the user. They have to infer that they can't duplicate it because there's another slug with the same value. 

We have a custom error message in `ensureUniqueSlug` but since the duplicate action doesn't submit the form, it just makes an API call, the [`ValidationError`](https://github.com/NWACus/web/blob/fix-uniq-slug/src/fields/slug/ensureUniqueSlug.ts#L59) that we send doesn't get displayed next to the custom slug field because Payload's `FieldError` component only shows when the form has been submitted. 

We'd ideally like it to show this (which currently shows if you save with a duplicate slug bug not when using the duplicate action in the case where there's already a `-copy` version):
Image

So we could change our [`ValidationError`](https://github.com/NWACus/web/blob/fix-uniq-slug/src/fields/slug/ensureUniqueSlug.ts#L59) to an:
```
    throw new APIError(
      `A ${collection.labels.singular} with the slug "${value}" already exists. Slug must be unique${collectionHasTenantField ? ' per avalanche center' : ''}.`,
      400,
    )
```

Which would _only_ show the error message in the toast. This would then be the same for save and for duplicate but we'd just have a slightly different behavior than the norm. That's probably the preferred option here. 

Alternatively, we could combine this `-copy` appending with the indexing logic from the removed `getIndexedTitleAndSlug` so that duplicate will _always_ work even if a user has already copied the document before. 

What are your thoughts on this?

src/components/Header/utils.ts
rchlfryn: I need to add it for the linter. It's related to the issue going on in payload-types so see the other comment

src/payload-types.ts
rchlfryn: Today I learned that parameter default placement matters for TypeScript's type inference. Moving the default from the parameter level to destructuring assignment defaults allowed the function to properly infer and return `title: string`

I also think it is very weird that only the pages collection was changed.

rchlfryn:
@busbyk I went ahead an changed the validation error to an API error.  I think it is the easiest and best solution. I did double check to see how payload handles this with the unique prop enabled (without the multi-tenant bug), and they just share the generic `The following field is invalid` toast.

busbyk:
> @busbyk I went ahead an changed the validation error to an API error. I think it is the easiest and best solution. I did double check to see how payload handles this with the unique prop enabled (without the multi-tenant bug), and they just share the generic `The following field is invalid` toast.

I like it. Seems like the best way to provide the right message to the user in a consistent way.

src/utilities/getDocumentBySlug.ts
busbyk, line 1: Not that I have an issue with this rename since it is more accurate but this isn't used anywhere in this PR right? I'm cool with throwing this in if it's just something you think makes sense (it does). Just curious 😅

busbyk (approved):
Looks great!

src/utilities/getDocumentBySlug.ts
rchlfryn: It is not used anywhere. I changed it because I went to use it and realized we could name it better. I know, a very random change

busbyk: Yea random change but I'm for it. Sounds good!

PR #837, opened 2026-01-07
Allow script tags in generic embeds

### LINKED ISSUE #816: New Sierra donation embed not functioning as expected
**Describe the bug**
Sierra updated their donation platform to accept stock and crypto. The Next button doesnt work when embedded in Avy. 

[Here it is standalone
](https://donorbox.org/sac-donate-alternatives)[Here it is on a test web page](https://www.sierraavalanchecenter.org/donate-app)

Embed code:

`<script type="module" src="https://donorbox.org/widgets.js" async></script><dbox-widget campaign="sac-donate-alternatives" type="donation_form" enable-auto-scroll="true"></dbox-widget>
`
Classic Embed code (same but different):

`<script src="https://donorbox.org/widget.js" paypalExpress="true"></script> <iframe src="https://donorbox.org/embed/sac-donate-alternatives?" name="donorbox" allowpaymentrequest="allowpaymentrequest" seamless="seamless" frameborder="0" scrolling="no" height="900px" width="100%" style="max-width: 500px; min-width: 250px; max-height:none!important" allow="payment"></iframe>
`
**To Reproduce**

1. Go to https://www.sierraavalanchecenter.org/donate-app 
2. Click on Crypto
3. Click Next button
4. Form does not advance

**Expected behavior**
Expect it to work as it does when standalone

**Environment details**

Desktop prod

**Existential musings**
What else you got for us?

busbyk:
## Description

This sets `FORCE_BODY` to true in DOMPurify settings to allow script tags. Apparently, even if you include `scripts` in the `ADD_TAGS` option, scripts will be stripped. This was causing the script that the donorbox embed needed to be stripped from the resulting sanitized html. 

We're still sandboxing all embeds in an iframe so the crypto modal will only open in the iframe document which doesn't look amazing. Since we resize to the height of the iframe content using `@open-iframe-resizer`, there isn't much vertical space for it to occupy:
CleanShot 2026-01-06 at 17 10 30

I still believe that these types of things should be resolved by setting styles on the iframe / embed html rather than trying to account for these things on our end. So I suggest we accept this. 

## Related Issues

Fixes #816

## Key Changes

Adds `FORCE_BODY=true` to DOMPurify settings, allowing script tags in sanitized html.

It might be helpful to write some best practices for styling embeds in our Notion docs.

### DIFF
--- a/src/blocks/GenericEmbed/Component.tsx
+++ b/src/blocks/GenericEmbed/Component.tsx
@@ -49,6 +49,7 @@ export const GenericEmbedBlock = ({
         'width',
         'allowpaymentrequest',
       ],
+      FORCE_BODY: true,
     })
 
     const styleOverrides = `


rchlfryn (approved):
Make sense to me. Hopefully we wont have to update this again.

PR #768, opened 2025-12-08
Fix tenant selector for global & readd to dashboard

rchlfryn:
## Description
Fixed the tenant selector not being set on global view & re-added it to the dashboard. I choose to not fully copy everything over because it was taking longer than I expected and we should make a follow up ticket for this if we want it

## Related Issues
Fixes #712 

## Key Changes
- I copied the provider files from the latest [multi-tenant-plugin..v3.67.0](https://github.com/payloadcms/payload/tree/v3.67.0/packages/plugin-multi-tenant/src/providers/TenantSelectionProvider) files. The main things I did not copy over..
  - updating `getTenantOptions` to use `tenantsArrayFieldName`, tenantsArrayTenantFieldName`, and `userHasAccessToAllTenants` 
  - the `populate-teanant-options` endpoint and use of it
    - https://github.com/payloadcms/payload/blob/v3.67.0/packages/plugin-multi-tenant/src/providers/TenantSelectionProvider/index.client.tsx#L173-L179
    - https://github.com/payloadcms/payload/blob/v3.67.0/packages/plugin-multi-tenant/src/endpoints/getTenantOptionsEndpoint.ts

## Future enhancements / Questions
Update all the other multi-tenant files

PR #838, opened 2026-01-07
Await form initialization before setting tenant field value

### LINKED ISSUE #832: Cannot add Role Assignment Read Only Pages on SAC
**Describe the bug**
I am trying to give Gabrielle at PAC the ability to see SAC pages like we set up for Sawtooth. I get an error (the following field is invalid: tenant) when trying to create the role assignment. 

**To Reproduce**

1. Go to SAC
2. Add Role Assignment (Read Only Pages + director@payetteavalanche.org
3. See error

**Expected behavior**
I used to be able to 
**Screenshots**

Image

**Environment details**

- Prod Chrome desktop

**Existential musings**
Maybe I just forgot how to do this?

busbyk:
## Description

The tenant field (which is hidden in the UI but required to match the currently selected tenant for tenant collections) was not being set on initial render. It's possible that this was introduced in #768 but I couldn't determine a change that would have caused this. Collections with versions and autosave enabled didn't have this problem because the autosave would cause an additional render, setting the tenant field correctly. Collections without versions/autosave wouldn't get the tenant field set in the initial render so when saving during create, we'd get a validation error for the undefined tenant field value. 

## Related Issues

Fixes #832 

## Key Changes

Changes made to TenantFieldComponent.tsx:
  1. Added useFormInitializing hook import
  2. Added formReady check (!formInitializing && !formInitializingContext)
  3. Effect now returns early if the form is still initializing, allowing it to re-run once the form is ready

## Test

Test this by creating a new role assignment successfully. You can also set debug: true in https://github.com/NWACus/web/blob/bugfix/tenant-field-value/src/fields/tenantField/index.ts#L22 temporarily to see the tenant field in the UI.

### DIFF
--- a/src/fields/tenantField/TenantFieldComponent.tsx
+++ b/src/fields/tenantField/TenantFieldComponent.tsx
@@ -2,7 +2,7 @@
 
 import type { RelationshipFieldClientProps } from 'payload'
 
-import { RelationshipField, useField, useFormModified } from '@payloadcms/ui'
+import { RelationshipField, useField, useFormInitializing, useFormModified } from '@payloadcms/ui'
 import { useEffect, useRef } from 'react'
 
 import { useTenantSelection } from '@/providers/TenantSelectionProvider/index.client'
@@ -17,8 +17,9 @@ type Props = {
 
 export const TenantFieldComponent = (args: Props) => {
   const { debug, unique } = args
-  const { setValue, value } = useField<number | string>()
+  const { setValue, value, formInitializing } = useField<number | string>()
   const modified = useFormModified()
+  const formInitializingContext = useFormInitializing()
   const {
     options,
     selectedTenantID,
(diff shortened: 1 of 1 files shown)


rchlfryn (approved):
Oh very interesting. It is unfortunate this will cause us to divert again from payload's `multi-tenant-plugin`. Do we want to make note of this diversion in the file? I think it is fine if we don't.

I wonder if this is a fix will work for #833. I was not confident in the PR I have out so I am excited to test this as a solution.

busbyk:
> Oh very interesting. It is unfortunate this will cause us to divert again from payload's `multi-tenant-plugin`. Do we want to make note of this diversion in the file? I think it is fine if we don't.

Well, in my mind, we have fully diverted from their `multi-tenant-plugin` and pulling in changes from theirs is optional. If we want to utilize their plugin we could revisit the decision in `docs/decisions/010-remove-multi-tenant-plugin.md`. We might be able to make it work for us again based on recent changes.

Issue #841, opened 2026-01-07
404 Error on some Sawtooth Events


rustynwac:
**Describe the bug**
Sawtooth is getting a 404 when you click on Learn More for the two Jan 15 events. This is one of them
https://avy-fx.org/admin/collections/events/35

**To Reproduce**

1. Go to events
2. Click on learn more
3. See error

**Expected behavior**
I would expect the event to open like the others 

**Screenshots**

<img width="944" height="668" alt="Image" src="https://github.com/user-attachments/assets/748f8e8e-9962-47f4-9345-c3ed235c1a68" />

**Environment details**

Desktop Chrome Prod

**Existential musings**
What else you got for us?

busbyk:
Temporarily resolved by re-running the latest release. Still need to do root cause analysis and push a fix so this doesn't happen again.

busbyk:
Notes:
- The issue here is an event that is created after the initial build (during the last deploy)
- Vercel/next is matching ./[...segmentsNotFound] instead of /events/[slug]
- Can't reproduce locally using both subdomains and custom domains (ran a local build and created a new event after being built)
- I wonder if we ran the app locally with the vercel CLI if we could recreate this behavior locally

andyanderso:
This seems to be working now on the Sierra site with both the event landing page and an external url. 
https://www.sierraavalanchecenter.org/events

busbyk:
Fixed by #947 and released in [v1.11.0](https://github.com/NWACus/web/releases/tag/v1.11.0)

PR #847, opened 2026-01-10
Do not cache non-tenant-scoped 404s

busbyk:
## Description

This is an attempt at restricting the non-tenant-scoped catch-all route which serves as the 404 page from being cached. I have a suspicion that this is what was causing #841 but I wasn't able to reproduce locally. 

## Related Issues

Related to #841 but couldn't identify root cause so I'm not sure this will resolve it

## Key Changes

Sets the `dynamic` export to `force-dynamic` in an effort to avoid the caching of 404s

I hate to make a change that I just think _might_ be the solution. I spent a while trying to recreate this locally with no luck so this is me doing _something_ about it. This is a low consequence change - it seems fine (and potentially ideal) to not cache 404s anyways.

### DIFF
--- a/src/app/[...segmentsNotFound]/page.tsx
+++ b/src/app/[...segmentsNotFound]/page.tsx
@@ -2,6 +2,9 @@ import Link from 'next/link'
 
 import { Button } from '@/components/ui/button'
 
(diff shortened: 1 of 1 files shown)


rchlfryn (approved):
Let's try it

Issue #825, opened 2025-12-27
Navigation improvements - donate tab & sort link refs


rchlfryn:
- When selecting an Internal Link , we should sort the pages/posts/built-in alphabetically to make it easier to find what you are looking for
- Change Donation enable option the same as the rest of the navigation
  From this
  <img width="939" height="214" alt="Image" src="https://github.com/user-attachments/assets/88b1a66f-8450-4d9e-8c4b-f86a6fee99c5" />

  To this
  <img width="958" height="185" alt="Image" src="https://github.com/user-attachments/assets/abbfa062-fd87-44b4-bf03-cd7f98e95c56" />

rustynwac:
@rchlfryn tech debt or new center onboarding or just a general enhancement we should consider?

Issue #685, opened 2025-11-17
Guidance on usage of CMSLink vs. Link + Button with clickWithCapture


busbyk:
Let's consolidate the usage of buttons that are links. We have a CMSLink component that is named and written such that it expects fields from the link field but can also accept an href. But there's a lot going on in there just for a Link + Button. 

This issue should consider the best approach to using links with button styling that also capture events. That might mean a refactor and rename of CMSLink or a new component as well. This issue should also add to the `docs/coding-guide.md` docs. 

Some context: https://github.com/NWACus/web/pull/669#discussion_r2535460975

busbyk:
Let's also use this issue to add [`clearIrrelevantLinkValues`](https://github.com/NWACus/web/blob/bugfix/navigation-nesting-issues/src/fields/navLink/index.ts#L13) to `defaultLexical` and `linkToPageOrPost`.

PR #848, opened 2026-01-11
Bugfix: Fix navigation canonical URL resolution and improve admin UX for nav items

### LINKED ISSUE #818: Motorized Classes Page on Sierra Site Triggers 404
**Describe the bug**
Clicking on the Motorized Avalanche Classes menu link results in a 404 error. The menu link points to: [https://www.sierraavalanchecenter.org/education/classes/motorized-avalanche-classes](https://www.sierraavalanchecenter.org/education/classes/motorized-avalanche-classes).
 
Using internal links seems to work fine. The blog post on the home page directs users to the Motorized Classes. I updated the link and the button on the blog post page to be internal links. Those links now point to this URL - [https://www.sierraavalanchecenter.org/education/motorized-avalanche-classes](https://www.sierraavalanchecenter.org/education/motorized-avalanche-classes), which seems to work fine. 

**To Reproduce**
1. Click on the motorized avalanche classes menu item
2. Get a 404 error

**Expected behavior**
Clicking on the motorized avalanche classes menu item takes you to the existing page.

**Screenshots**

Image

**Environment details**
IOS phone and laptop, Linux, Android - Chrome, Safari, Firefox

**Existential musings**
It's raining...

busbyk:
### Summary

- Fix incorrect canonical URLs for pages referenced in navigation items that have sub-items (accordion items)
- Improve admin UX by showing only relevant fields based on nav item type (link vs accordion)
- Clear stale link data when nav items are converted to accordions
- Clear irrelevant link field values when switching between internal/external link types

### Root Cause Analysis

Navigation items with sub-items render as accordion triggers (expand/collapse), not clickable links. However, the canonical URL resolution functions were treating these items as if they were clickable, causing pages to get incorrect canonical URLs pointing to non-navigable paths. It would also only find the first reference to a page.

#### How Navigation Rendering Works

In `DesktopNav.client.tsx`:
- Items **without** sub-items → rendered as clickable `<NavigationMenuLink>`
- Items **with** sub-items → rendered as `<AccordionTrigger>` that expands/collapses; the link URL is never used as a navigation target

#### The Bug

The functions in `utils-pure.ts` (`extractAllInternalUrls`, `findNavigationItemBySlug`, `getNavigationPathForSlug`) recursively traversed ALL navigation items without checking whether they were actually clickable. When a page was referenced on an accordion item, the canonical URL logic would find and use that reference, even though users couldn't actually navigate to it via the menu. So when there was a reference to the same page in the link for a third level accordion that was also in one of that items' sub items, the accordion level link would be used (which wouldn't have a third level of nesting). 

This was illustrated nicely in our seed data (prior to this PR's changes). Education -> Classes had a page reference for the Field Classes page for it's link since an itemsField needed a link to have a label. Field Classes was also the first sub item (Education -> Classes -> Field Classes). So we would expect the canonical URL for Field Classes to be /education/classes/field-classes but since it was being found first at the Classes label itemsField, the canonical URL was being recognized as /education/field-classes and click on the link for /education/classes/field-classes would 404. 

## Related Issues

Fixes #818

## Key Changes

1. **Clear link values when switching types** (`src/fields/navLink/index.ts`)
   - Added `beforeChange` hook that clears `url`/`newTab` when switching to internal, and clears `reference` when switching to external

2. **Add standalone label field with conditional visibility** (`src/collections/Navigations/fields/itemsField.ts`, `src/components/Header/`)
   - Added `label` text field to nav items schema
   - **Label field** only shows when item has sub-items (accordion mode)
   - **Link fields** only show when item has no sub-items (direct link mode)
   - When sub-items are added, `link.label` is automatically copied to standalone `label`
   - When sub-items are added, link data is cleared to avoid stale data
   - Updated `NavItem` type to include optional `label`
   - Desktop/Mobile nav components now use `item.label || item.link?.label`

3. **Skip accordion items in canonical URL resolution** (`src/components/Header/utils-pure.ts`)
   - All three traversal functions now check `hasChildren` and skip items with sub-items (items with children are not clickable links, so they shouldn't contribute to canonical URLs)
   - Canonical URL path resolution now uses standalone `label` field for accordion items when building breadcrumb paths
   - Added tests verifying accordion items are excluded from URL extraction and slug lookup

4. **Use standalone label in canonical URL path resolution** (`src/components/Header/utils.ts`)
   - Updated `topLevelNavItem()` to check `label` field before `link.label` when building parent path segments
   - Ensures accordion-style nav items (with standalone labels) are included in breadcrumb path resolution

## Screenshots / Demo

https://www.loom.com/share/ae46670a162b4670969ca2eb2d1b72a8

☝️☝️ bit long tbh so don't be afraid to 1.5x that 😇

## How to test

- Create nav item without sub-items, verify only Link fields show
- Add sub-items, verify Label field appears and Link fields disappear
- Verify link.label is copied to standalone label when sub-items added
- Verify link data is cleared in database when sub-items added

## Migration Explanation

See `src/migrations/20260111_205454_nav_item_standalone_label_explanation.md`.

I tested this locally by checking out the main branch and performing steps 1-7 from [Testing a migration locally](https://github.com/NWACus/web/blob/bugfix/navigation-nesting-issues/docs/migration-safety.md#L61) and then checking out this PR's branch and performing steps 8-10. The data was correctly migrated from the old link group field to the new standalone label fields. 

There is another bug that we haven't run into yet because we control the navigations strictly but that we should address:

If we have the following three urls from placing the local accident pages throughout the navigation:
```
http://nwac.localhost:3000/accidents/local-accident-reports
http://nwac.localhost:3000/education/classes/local-accident-reports
http://nwac.localhost:3000/education/local-accident-reports
```

The /education/classes/local-accident-reports url will 404 because the canonical url logic finds the /education/local-accident-reports url first and then stops. /accidents/local-accident-reports and /education/local-accident-reports both work. 

I'd like to address this in a separate PR.

### DIFF
--- a/__tests__/server/headerUtils.server.test.ts
+++ b/__tests__/server/headerUtils.server.test.ts
@@ -405,6 +405,40 @@ describe('Header Utilities', () => {
 
       expect(path).toEqual(['/education', '/education/classes', '/education/classes/field-classes'])
     })
+
+    it('handles accordion item with standalone label (no link) as parent', () => {
+      // This tests the scenario where a parent nav item uses standalone label
+      // instead of link (accordion mode) - the child items should still include
+      // the parent's label in the canonical URL path
+      const navWithStandaloneLabel: TopLevelNavItem[] = [
+        {
+          label: 'Education',
+          items: [
+            {
+              id: 'classes-accordion',
+              // Standalone label, no link - this is an accordion trigger
+              label: 'Classes',
+              items: [
+                {
+                  id: 'field-classes',
+                  link: {
+                    type: 'internal',
+                    label: 'Field Classes',
+                    // URL already includes /education/classes/ segment from parent
+                    url: '/education/classes/field-classes',
+                  },
+                },
+              ],
+            },
+          ],
+        },
+      ]
+
+      // The path should include the education and classes segments
+      const path = getNavigationPathForSlug(navWithStandaloneLabel, 'field-classes')
+
+      expect(path).toEqual(['/education/classes/field-classes'])
+    })
   })
 
   describe('getCanonicalUrlsFromNavigation', () => {
--- a/__tests__/server/headerUtils.server.test.ts
+++ b/__tests__/server/headerUtils.server.test.ts
@@ -155,6 +155,53 @@ describe('Header Utilities', () => {
 
       expect(urls).toEqual(['/about', '/blog'])
     })
+
+    it('excludes URLs from items that have children (accordion items)', () => {
+      // Items with children render as accordion triggers, not clickable links
+      // So their URLs should not be included in canonical URL extraction
+      const navWithAccordionItem: TopLevelNavItem[] = [
+        {
+          label: 'Education',
+          items: [
+            {
+              id: 'classes-section',
+              // This item has children AND a link, but the link is not clickable (accordion trigger)
+              link: {
+                type: 'internal',
+                label: 'Classes',
+                url: '/education/classes',
+              },
+              items: [
+                {
+                  id: 'field-class',
+                  link: {
+                    type: 'internal',
+                    label: 'Field Class',
+                    url: '/education/classes/field',
+                  },
+                },
+              ],
+            },
+            {
+              id: 'resources',
+              // This item has NO children, so its link IS clickable
+              link: {
+                type: 'internal',
+                label: 'Resources',
+                url: '/education/resources',
+              },
+            },
+          ],
+        },
+      ]
+
+      const urls = extractAllInternalUrls(navWithAccordionItem)
+
+      // Should include /education/classes/field and /education/resources
+      // Should NOT include /education/classes (accordion item has children)
+      expect(urls).toEqual(['/education/classes/field', '/education/resources'])
+      expect(urls).not.toContain('/education/classes')
+    })
   })
 
   describe('findNavigationItemBySlug', () => {
@@ -217,6 +264,53 @@ describe('Header Utilities', () => {
           type: 'internal',
           label: 'Weather Stations',
           url: '/weather/stations/map',
+        },
+      })
+    })
+
+    it('does not find items that have children (accordion items)', () => {
+      // Items with children render as accordion triggers, not clickable links
+      // So they should not be found when searching for canonical URLs
+      const navWithAccordionItem: TopLevelNavItem[] = [
+        {
+          label: 'Education',
+          items: [
+            {
+              id: 'classes-section',
+              // This item has children AND a link with slug "classes"
+              // But it's an accordion trigger, not a clickable link
+              link: {
+                type: 'internal',
+                label: 'Classes',
+                url: '/education/classes',
+              },
+              items: [
+                {
+                  id: 'field-class',
+                  link: {
+                    type: 'internal',
+                    label: 'Field Class',
+                    url: '/education/classes/field',
+                  },
+                },
+              ],
+            },
+          ],
+        },
+      ]
+
+      // Searching for 'classes' should NOT find the accordion item
+      const item = findNavigationItemBySlug(navWithAccordionItem, 'classes')
+      expect(item).toBeNull()
+
+      // But searching for 'field' should find the child item (no children)
+      const childItem = findNavigationItemBySlug(navWithAccordionItem, 'field')
+      expect(childItem).toEqual({
+        id: 'field-class',
+        link: {
+          type: 'internal',
+          label: 'Field Class',
+          url: '/education/classes/field',
         },
       })
     })
--- a/src/collections/Navigations/fields/itemsField.ts
+++ b/src/collections/Navigations/fields/itemsField.ts
@@ -53,6 +53,7 @@ export const itemsField = ({
         {
           name: 'label',
           type: 'text',
+          required: true,
           admin: {
             description: 'Label for this nav section (shown when item has sub-items)',
             condition: hasSubItems,
--- a/src/collections/Navigations/fields/itemsField.ts
+++ b/src/collections/Navigations/fields/itemsField.ts
@@ -20,6 +20,13 @@ export const itemsField = ({
         description,
       },
       fields: [
+        {
+          name: 'label',
+          type: 'text',
+          admin: {
+            description: 'Label for this nav section (used when item has sub-items but no link)',
+          },
+        },
         navLink,
         {
           name: 'items',
--- a/src/collections/Navigations/fields/itemsField.ts
+++ b/src/collections/Navigations/fields/itemsField.ts
@@ -6,10 +6,6 @@ import { ArrayField, FieldHook } from 'payload'
 const hasSubItems = (_data: unknown, siblingData: Record<string, unknown>) =>
   Array.isArray(siblingData?.items) && siblingData.items.length > 0
 
-// Condition: show field only when item has NO sub-items (direct link mode)
-const hasNoSubItems = (_data: unknown, siblingData: Record<string, unknown>) =>
-  !Array.isArray(siblingData?.items) || siblingData.items.length === 0
-
 // Copy link.label to standalone label when sub-items are added,
 // and clear standalone label when sub-items are removed
 const clearLabelWhenItemHasNoSubItems: FieldHook = ({ siblingData, value }) => {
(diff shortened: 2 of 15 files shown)


__tests__/server/headerUtils.server.test.ts
```diff
@@ -155,6 +155,53 @@ describe('Header Utilities', () => {
...
+    it('excludes URLs from items that have children (accordion items)', () => {
+      // Items with children render as accordion triggers, not clickable links
+      // So their URLs should not be included in canonical URL extraction
+      const navWithAccordionItem: TopLevelNavItem[] = [
+        {
+          label: 'Education',
+          items: [
+            {
+              id: 'classes-section',
+              // This item has children AND a link, but the link is not clickable (accordion trigger)
```
stevekuznetsov, line 168: Why is it possible in our data model to add a link to such a record? That seems like a bug, if it's not click-able and can't be navigated to, no?

busbyk: Building out the navigation data model was a tumultuous experience - the requirements changed during that build out and there was a lot of back and forth on technical design decisions like whether or not a page's location in the navigation should determine the sitemap or not.

An `itemsField` having a `navLink` field and an `items` field (i.e. sub items) is because:
1. Originally, the headings of nested sub items were clickable. I suppose this is a left over from that.
2. We would have had use a `blocks` type for top level nav item children if we wanted to have two different schemas for "just links" and "sub nav items". I considered making that change here but the migration would have been much more difficult so I chose conditional fields which feels like a nice enough UI and makes the data migration easier.

src/collections/Navigations/fields/itemsField.ts
```diff
@@ -20,7 +61,28 @@ export const itemsField = ({
...
+          },
+          hooks: {
+            beforeChange: [syncStandaloneLabelWithSubItems],
+          },
+        },
+        {
+          ...navLink,
+          admin: {
+            ...navLink.admin,
+            condition: hasNoSubItems,
```
rchlfryn: ⛏️ I see you're being very explicit about this but isn't this just `!hasSubItems`?

src/collections/Navigations/fields/itemsField.ts
```diff
@@ -20,7 +61,28 @@ export const itemsField = ({
         description,
       },
       fields: [
-        navLink,
+        {
+          name: 'label',
+          type: 'text',
+          admin: {
```
rchlfryn, line 57: This should be required

__tests__/server/headerUtils.server.test.ts
stevekuznetsov: In my opinion, whatever the cost is of the migration, it's worth it long-term. I don't think it makes sense to allow a data model with fields that are nonfunctional - especially since it makes logic like the file we're chatting on more difficult to write. Very subtle that there are non-empty link fields that should be ignored even though they have content.

src/components/Header/utils-pure.ts
```diff
@@ -10,7 +10,11 @@ export function extractAllInternalUrls(navItems: TopLevelNavItem[]): string[] {
   const urls: string[] = []
 
   function extractFromNavItem(item: NavItem | TopLevelNavItem): void {
-    if (item.link && item.link.type === 'internal') {
+    const hasChildren = item.items && item.items.length > 0
+
+    // Only include link if item has NO children
+    // Items with children render as accordion triggers (expand/collapse), not clickable links
+    if (item.link && item.link.type === 'internal' && !hasChildren) {
```
rchlfryn: Should we just make this (`item.link && item.link.type === 'internal' && !hasChildren`) a thoughtfully named const so we can remove the comments and use below?

src/collections/Navigations/fields/itemsField.ts
```diff
@@ -1,6 +1,47 @@
...
+const hasSubItems = (_data: unknown, siblingData: Record<string, unknown>) =>
+  Array.isArray(siblingData?.items) && siblingData.items.length > 0
+
+// Condition: show field only when item has NO sub-items (direct link mode)
+const hasNoSubItems = (_data: unknown, siblingData: Record<string, unknown>) =>
+  !Array.isArray(siblingData?.items) || siblingData.items.length === 0
+
+// Copy link.label to standalone label when sub-items are added,
+// and clear standalone label when sub-items are removed
+const syncStandaloneLabelWithSubItems: FieldHook = ({ siblingData, value }) => {
```
rchlfryn: I don't think this actually does anything.
I added a nav item -> added sub nav items -> published and didn't see any change.
I also did not see the label get cleared when I removed the subitems. Was the intention to clear the UI?

src/fields/navLink/index.ts
```diff
@@ -10,12 +10,36 @@ const validateLabel: TextFieldSingleValidation = (val, { siblingData }) => {
   return Boolean(val) || 'You must define a label for an external link.'
 }
 
-export const navLink: GroupField = {
+const clearIrrelevantLinkValues: FieldHook = ({ value }) => {
```
rchlfryn, line 13: We should clear external/internal values when the selection changes, similar to how it's already done here for the `defaultLexical` and `linkToPageOrPost` fields. Since this file is still nav related, it totally reasonable to create a follow-up PR/issue to clean this up

rchlfryn (changes requested):
Good stuff on the navigation PR. When I was first looking at the issue, I made a side note to clean up the "non-clickable" links but didn't know it was the actual reason for the bug.  I'll remove this from #825

src/collections/Navigations/fields/itemsField.ts
busbyk: Oh you're totally right. Because this is a `beforeChange` hook this isn't behaving as I intended. I would have needed to create a custom client component to manage updating the various field's values using `useField` hooks.

I'm just going to change this to still be a `beforeChange` hook but to only clear the label value when there are no subitems. You may not have seen that particular behavior if you didn't save it first or if you didn't select draft when viewing the API tab?

src/collections/Navigations/fields/itemsField.ts
busbyk: Yea cool, can skip the extra function and just use `condition: (data: unknown, siblingData: Record<string, unknown>) => !hasSubItems(data, siblingData)`

src/components/Header/utils-pure.ts
busbyk: How's `isClickableInternalLink()`?

src/fields/navLink/index.ts
busbyk: I like it. I added that to https://github.com/NWACus/web/issues/685#issuecomment-3741201788

__tests__/server/headerUtils.server.test.ts
busbyk: I definitely consider the cost of tasks in decision making. But as far as this being a viable long-term solution: I think that it is. We use conditional field rendering in a few places in the admin panel and I think it's a perfectly acceptable pattern. The fields are not nonfunctional. They are just only relevant depending on the value of another field. 

Using blocks might be a slightly more intuitive UI but it results in more tables in our db and a larger payload config. We could certainly re-evaluate whether conditional field rendering is intuitive to users once we've had a full season of usage.

src/collections/Navigations/fields/itemsField.ts
rchlfryn: It looks like this was never updated. I think it's fine either way but leaving a comment in case you meant to change this.

rchlfryn (approved):
LG: a final thought, do we want to leverage `filterOptions` on `navLink.reference` to only show pages/built-in pages that are not previously selected in the navigation by filtering on the reference id? Just a thought as we only want pages in the nav to appear once.

busbyk:
> LG: a final thought, do we want to leverage `filterOptions` on `navLink.reference` to only show pages/built-in pages that are not previously selected in the navigation by filtering on the reference id? Just a thought as we only want pages in the nav to appear once.

Per our Slack discussion we decided to leave it up to the super admin to not add pages more than twice. We need to allow pages being added more than once because of the donate page. We could definitely add something in another PR to warn the user if they've added a page too many times / in a problematic way (or stop them from doing that in a beforeValidate hook).

src/collections/Navigations/fields/itemsField.ts
busbyk: A CLASSIC "forgot to push" moment 🤦‍♂️

rchlfryn (approved):

PR #849, opened 2026-01-11
Add international as a state for courses and providers

### LINKED ISSUE #815: Add international providers
We could temporarily hack international providers by adding a state, but may want to do a bit more discovery and figure out if it's worth listing countries independently. What does this look like for A3 in 1-3 years?

Currently the International based Providers are:

[Aprendica Snow Safety](https://avalanche.org/course_provider/aprendica-snow-safety/)
[Argentine Higher Institute of Mountain Guides (ISAGM)](https://avalanche.org/course_provider/argentine-higher-institute-of-mountain-guides-a-090-isagm/)
[Chamonix Experience](https://avalanche.org/course_provider/chamex/)
[Rimaya](https://avalanche.org/course_provider/rimaya/) 

And US providers offering courses internationally (that were also on that list): 
American Avalanche Institute
Avalanche Science, LLC

I currently have 3 applications in the queue as well from China, Chile, and Argentina

rchlfryn:
## Description
A3 recently brought to our attention they have a few international providers. These providers are either strictly international or have courses in both the US and internationally. The easiest way for us to deal with this is to add international as a state option to both courses and providers. 

## Related Issues
Fixes #815 

## Key Changes
- Add international as state option
- Update provider preview to show international label
- Update course filters to put international option on the bottom of the state list

## Screenshots / Demo
Recorded locally since embeds were not working in preview
https://www.loom.com/share/1687db9d24034cdda6ac86260c252e38

- We will want to decide if we are supporting international addresses or not.
- When discussing this, there was mention of updating permission. Do we need to limit the international option to select user or alter the provider invite flow? Nothing was documented around this.
- The admin UI is not great. I am leaving it since we are treating this PR as a workaround but could use improvement

### DIFF
--- a/src/fields/location/states.ts
+++ b/src/fields/location/states.ts
@@ -50,11 +50,11 @@ export const stateOptions = [
   { label: 'Wisconsin', value: 'WI' },
   { label: 'Wyoming', value: 'WY' },
   { label: 'District of Columbia', value: 'DC' },
-  { label: 'International', value: 'INTL' },
 ]
+export const stateOptionsWIntl = [...stateOptions, { label: 'International', value: 'INTL' }]
 
 // Get full state label from value
 export const getStateLabel = (stateValue: string): string => {
-  const state = stateOptions.find((option) => option.value === stateValue)
+  const state = stateOptionsWIntl.find((option) => option.value === stateValue)
   return state?.label || stateValue
 }
--- a/src/fields/location/states.ts
+++ b/src/fields/location/states.ts
@@ -50,6 +50,7 @@ export const stateOptions = [
   { label: 'Wisconsin', value: 'WI' },
   { label: 'Wyoming', value: 'WY' },
   { label: 'District of Columbia', value: 'DC' },
+  { label: 'International', value: 'INTL' },
 ]
 
 // Get full state label from value
--- a/src/utilities/queries/getCoursesStates.ts
+++ b/src/utilities/queries/getCoursesStates.ts
@@ -31,6 +31,13 @@ export async function getCoursesStates(): Promise<GetCoursesStatesResults> {
       }
     })
 
+    // Move international to the bottom of the list
+    const intlIndex = states.findIndex((s) => s.value === 'INTL')
+    if (intlIndex !== -1) {
+      const [intlItem] = states.splice(intlIndex, 1)
+      states.push(intlItem)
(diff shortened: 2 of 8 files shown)


src/utilities/queries/getCoursesStates.ts
```diff
@@ -31,6 +31,13 @@ export async function getCoursesStates(): Promise<GetCoursesStatesResults> {
       }
     })
 
+    // Move international to the bottom of the list
+    const intlIndex = states.findIndex((s) => s.value === 'INTL')
+    if (intlIndex !== -1) {
+      const [intlItem] = states.splice(intlIndex, 1)
+      states.push(intlItem)
+    }
```
busbyk, line 39: Nice

src/fields/location/states.ts
```diff
@@ -50,6 +50,7 @@ export const stateOptions = [
   { label: 'Wisconsin', value: 'WI' },
   { label: 'Wyoming', value: 'WY' },
   { label: 'District of Columbia', value: 'DC' },
+  { label: 'International', value: 'INTL' },
```
busbyk: Since this `stateOptions` array is used by the `locationField` too this will show up in the events collection as well:
Image

Do we care? You can't filter events by state so maybe this is acceptable until we decide it's worth adding other countries.

busbyk (commented):
> We will want to decide if we are supporting international addresses or not.

I wonder if we could do the simple version of that right now. We could have two countries: US and International. If you select international we'd show a "full address" field and if US we'd show our current location fields (i.e. street address, city, state, zip). 

I suppose we'd still need to have international as a state for the `statesServiced` field though 🤔

> When discussing this, there was mention of updating permission. Do we need to limit the international option to select user or alter the provider invite flow? Nothing was documented around this.

We decided that we would just not invite users for international providers thereby disallowing them from creating their own courses.

src/fields/location/states.ts
rchlfryn: I went adhead and remove international from events in [20cf7fa](https://github.com/NWACus/web/pull/849/commits/20cf7fadaad5698b201099fa082de86ffc7cbf68)

rchlfryn:
For now - we will just keep international as an option for state in courses and providers

busbyk (approved):
Sweet

PR #850, opened 2026-01-12
CLAUDE.md file

busbyk:
## Description

I've been using a CLAUDE.md file locally for a while now and recently refreshed it for recent changes to the codebase. I figured I might as well share it with the team. This will also be useful for when we set up (or at least experiment with) allowing Claude to create PRs from issues. 

[What is the CLAUDE.md file?](https://www.anthropic.com/engineering/claude-code-best-practices)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -180,7 +180,7 @@ See `.env.example` for all available flags. Key ones:
 
 - Only add code comments when necessary - i.e. the code is not easy to understand and needs more thorough explanation
 - Always add a code comment for regex expressions or string replacements
-- Only use TypeScript casting when absolutely necessary. Prefer using correct types and type guards (see `/docs/coding-guide.md`)
+- Never use TypeScript type assertions / casting like `const someVar = val as SomeType`. Write code so that TypeScript can infer the correct type. Type guards can be helpful when you might otherwise use a type assertion.
 - Prefer Payload's logger over `console.log`
 - Prefer Tailwind utility classes over adding `.css` or `.scss` files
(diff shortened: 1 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,242 @@
...
+
+---
+
+## Coding Rules
+
+### Style Preferences
+
+- Only add code comments when necessary - i.e. the code is not easy to understand and needs more thorough explanation
+- Always add a code comment for regex expressions or string replacements
+- Only use TypeScript casting when absolutely necessary. Prefer using correct types and type guards (see `/docs/coding-guide.md`)
```
stevekuznetsov: What happens if you tell Claude to never do it? :P

CLAUDE.md
```diff
@@ -0,0 +1,242 @@
...
+- Prefer Payload's logger over `console.log`
```
stevekuznetsov, line 184: Why 'prefer' over 'you must always use'?

stevekuznetsov (commented):
Looks reasonable! Couple places where the direction is "squishy" and I wonder if we could make it absolute.

CLAUDE.md
busbyk: Just because there are places where we can't currently use the Payload logger. See https://github.com/NWACus/web/issues/398#issuecomment-3407712769 for details.

CLAUDE.md
busbyk: Yea that's probably a good call - we know there are times when casting still makes sense but might as well be more strict with the LLM.

PR #851, opened 2026-01-12
Bugfix: Fixing event date hydration error

### LINKED ISSUE #751: Hydration error on /events due to dates
**Describe the bug**
We get a hydration error (observed in the browser console as a minified react error) on /events because the large date in the upper left of event previews does not respect the startDate field's timezone. 

A simple fix to use the same formatDateTime utility function that uses the timezone as the rest of the preview card. 

**To Reproduce**

1. Visit https://www.avy-fx-demo.org/events

You'll notice that you see a flash of different dates due to server vs. client rendering. 

**Expected behavior**
No hydration error

**Screenshots**

Image

busbyk:
## Description

Fix date hydration errors on event cards by using timezone-aware date formatting.

## Related Issues

Fixes #751

## Key Changes

- Replace raw `new Date()` methods with `formatDateTime()` in EventPreview for the large date sidebar display
- Replace `isSameDay()` from date-fns with timezone-aware comparison in StartAndEndDateDisplay
- Add test event seed data with midnight UTC start time to reproduce the issue

## Screenshots / Demo

No hydration error :)

CleanShot 2026-01-11 at 21 17 47@2x

## Rule of thumb for hydration mismatches for dates

**Triggering date hydration errors:** Hydration mismatches occur when a date falls on different calendar days in UTC vs the user's timezone:

| User Timezone | Problematic UTC Times |
|--------------|----------------------|
| US (behind UTC) | Early morning UTC (00:00-08:00) |
| Ahead of UTC | Late evening UTC (16:00-23:59) |

To test locally: `TZ=UTC pnpm dev`

### DIFF
--- a/src/components/StartAndEndDateDisplay.tsx
+++ b/src/components/StartAndEndDateDisplay.tsx
@@ -1,6 +1,18 @@
 import type { Event } from '@/payload-types'
 import { formatDateTime } from '@/utilities/formatDateTime'
-import { isSameDay } from 'date-fns'
+
+// Timezone-aware same-day comparison to avoid hydration mismatches
+// Compares the formatted date strings in the event's timezone rather than local timezone
+const isSameDayInTimezone = (
+  date1: string,
+  tz1: string | null | undefined,
+  date2: string,
+  tz2: string | null | undefined,
+): boolean => {
+  const day1 = formatDateTime(date1, tz1, 'yyyy-MM-dd')
+  const day2 = formatDateTime(date2, tz2, 'yyyy-MM-dd')
+  return day1 === day2
+}
 
 export function StartAndEndDateDisplay({
   startDate,
@@ -10,7 +22,7 @@ export function StartAndEndDateDisplay({
 }: Pick<Event, 'startDate' | 'startDate_tz' | 'endDate' | 'endDate_tz'>) {
   return (
     <>
-      {endDate && isSameDay(startDate, endDate) ? (
+      {endDate && isSameDayInTimezone(startDate, startDate_tz, endDate, endDate_tz) ? (
         <>
           <span>
             {formatDateTime(startDate, startDate_tz, 'MMM d, yyyy, p')}
--- a/src/components/StartAndEndDateDisplay.tsx
+++ b/src/components/StartAndEndDateDisplay.tsx
@@ -1,18 +1,7 @@
 import type { Event } from '@/payload-types'
 import { formatDateTime } from '@/utilities/formatDateTime'
-
-// Timezone-aware same-day comparison to avoid hydration mismatches
-// Compares the formatted date strings in the event's timezone rather than local timezone
-const isSameDayInTimezone = (
-  date1: string,
-  tz1: string | null | undefined,
-  date2: string,
-  tz2: string | null | undefined,
-): boolean => {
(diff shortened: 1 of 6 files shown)


src/components/StartAndEndDateDisplay.tsx
```diff
@@ -1,6 +1,18 @@
...
+
+// Timezone-aware same-day comparison to avoid hydration mismatches
+// Compares the formatted date strings in the event's timezone rather than local timezone
+const isSameDayInTimezone = (
+  date1: string,
+  tz1: string | null | undefined,
+  date2: string,
+  tz2: string | null | undefined,
+): boolean => {
+  const day1 = formatDateTime(date1, tz1, 'yyyy-MM-dd')
```
stevekuznetsov: Instead of doing stringly-typed comparisons, create a timezone-aware Date object and do the comparison itself.

src/components/StartAndEndDateDisplay.tsx
```diff
@@ -1,6 +1,18 @@
 import type { Event } from '@/payload-types'
 import { formatDateTime } from '@/utilities/formatDateTime'
-import { isSameDay } from 'date-fns'
+
+// Timezone-aware same-day comparison to avoid hydration mismatches
+// Compares the formatted date strings in the event's timezone rather than local timezone
+const isSameDayInTimezone = (
```
rchlfryn: This function is fine but `isSameDay` had options to account for timezones if you wanted to use that 
https://date-fns.org/v4.1.0/docs/isSameDay#

So something like this would work.
```ts
const result = isSameDay(date1, date2, {
  in: (date) => new TZDate(date, "UTC") // Normalize to a specific timezone
});
```

src/endpoints/seed/events.ts
```diff
@@ -258,5 +278,39 @@ export const getEventsData = (
         'Your observations make a difference! Community data is essential for accurate avalanche forecasting.',
       ),
     },
+
+    // HYDRATION ERROR TEST EVENT
```
rchlfryn, line 282: This would be nice to convert into an actual test once we have our test suite up.

rchlfryn (approved):

src/endpoints/seed/events.ts
busbyk: Yea an integration test via Playwright or similar, totally. A more generic check for hydration errors on various pages seems like a good idea.

src/components/StartAndEndDateDisplay.tsx
```diff
@@ -10,7 +11,7 @@ export function StartAndEndDateDisplay({
 }: Pick<Event, 'startDate' | 'startDate_tz' | 'endDate' | 'endDate_tz'>) {
   return (
     <>
-      {endDate && isSameDay(startDate, endDate) ? (
+      {endDate && isSameDay(startDate, endDate, { in: tz(startDate_tz ?? undefined) }) ? (
```
rchlfryn: why is `undefined` here?

rchlfryn (approved):
LG - just one minor comment

src/components/StartAndEndDateDisplay.tsx
busbyk: Yep, we could remove it

PR #856, opened 2026-01-13
Add vercel cleanup to GHA

rchlfryn:
### Description
Since we ran out of space, we need to properly clean up Vercel preview items. This includes two main aread:
* Environment variables
  * This script works great! Added a branch catch all just in case (will need to remove for `main`)
  * We need to remove 4 variables associated with each branch:
      - `DATABASE_URI`
      - `DATABASE_AUTH_TOKEN`
      - `PAYLOAD_SEED_PASSWORD`
      - `NEXT_PUBLIC_ROOT_DOMAIN`

* Deployments and aliases
  * We need to remove the alias first since it is associated with a deployment id
  * Each branch/PR has 2 aliases that switch to the most recently deployed deployment
    * Wild card alias: `BRANCH_NAME.preview.avy-fx.org`
    * ID alias: `avy-HASH-nwac.vercel.app`
    
 TODO: 
 - Clean up deployments for `main` from `github-merge-queue[bot]`
 - Clean up deployments for `release`

### DIFF
--- a/.github/workflows/cleanup.yaml
+++ b/.github/workflows/cleanup.yaml
@@ -21,9 +21,6 @@ jobs:
           branch="${ref##refs/heads/}"
           clean_branch="${branch//[^a-z0-9\-]/x}"
 
-          # Truncate ref to a prefix + hash suffix to avoid collisions while maintaining readability
-          # and ensuring total database name stays within 58 character limit
-          # "payloadcms-preview-" is 19 characters, leaving 39 for the branch
           if [ ${#clean_branch} -gt 39 ]; then
             prefix="${clean_branch:0:31}"
             hash=$(echo -n "${clean_branch}" | sha256sum | cut -c1-7)
@@ -33,10 +30,8 @@ jobs:
           name="payloadcms-preview-${clean_branch}"
           echo "[INFO] Attempting to delete database ${name}"
 
-          # Switch to the correct organization
           /home/runner/.turso/turso org switch nwac
 
-          # Check if database exists before attempting to delete
           if /home/runner/.turso/turso db show "${name}" >/dev/null 2>&1; then
             echo "[INFO] Database ${name} exists, deleting..."
             /home/runner/.turso/turso db destroy --yes "${name}"
@@ -60,7 +55,6 @@ jobs:
           ref="${{ github.event.ref }}"
           branch="${ref##refs/heads/}"
 
-          # Safety check: never delete from production
           if [ "$branch" = "main" ] || [ "$branch" = "release" ]; then
             echo "[ERROR] Refusing to delete environment variables from production branch: ${branch}"
             exit 1
@@ -82,6 +76,7 @@ jobs:
         env:
           VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
           VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
 
       - name: Remove Vercel deployments and aliases
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/cleanup.yaml
```diff
@@ -76,46 +83,44 @@ jobs:
...
-            for DEPLOYMENT_ID in $DEPLOYMENTS; do
-              echo "[INFO] Removing deployment: $DEPLOYMENT_ID"
-              vercel remove "$DEPLOYMENT_ID" --token=${{ secrets.VERCEL_TOKEN }} --yes || echo "[WARN] Failed to remove deployment: $DEPLOYMENT_ID"
-            done
-          fi
+          BRANCH_NAME="${{ github.head_ref }}"
+          echo "Cleaning up deployments and aliases for branch: $BRANCH_NAME"
+
+          # Get all deployments for the project
+          DEPLOYMENTS=$(curl -s -H "Authorization: Bearer $VERCEL_TOKEN" \
```
busbyk: Can we not do this with the CLI? Not that there's necessarily anything wrong with using the REST API but the CLI feels a little bit more readable and it would be consistent with the rest of the workflows.

.github/workflows/cleanup.yaml
```diff
@@ -76,46 +83,44 @@ jobs:
           VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
           VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
 
-      - name: Remove Vercel Deployments and Aliases
+      - name: Remove Vercel deployments and aliases
```
busbyk, line 86: I'm actually having a hard time determining what changed with this job. What's the major change here besides using the REST API instead of the CLI?

.github/workflows/cleanup.yaml
```diff
@@ -59,6 +60,12 @@ jobs:
           ref="${{ github.event.ref }}"
           branch="${ref##refs/heads/}"
 
+          # Safety check: never delete from production
+          if [ "$branch" = "main" ] || [ "$branch" = "release" ]; then
+            echo "[ERROR] Refusing to delete environment variables from production branch: ${branch}"
+            exit 1
+          fi
```
busbyk: 👏

busbyk (commented):
> TODO:
> Clean up deployments for main from github-merge-queue[bot]
> Clean up deployments for release

I'm for deleting all old deployments. We could restore the old deployments manually like you mentioned.

.github/workflows/cleanup.yaml
rchlfryn: ~That's really it - I was having a hard time with the CLI. It seems like the CLI is more for action (deploy, install, etc) and the API allows for better query and search which is what we need.~

I also think the branch name was set incorrectly

.github/workflows/cleanup.yaml
rchlfryn: As I mentioned below, I think the CLI is limited in what it allows.
```bash
➜  web git:(gha-cleanup) vercel alias ls
Vercel CLI 48.12.0
> aliases found under nwac [832ms]
  source    url    age
```

stevekuznetsov: (unsolicited review, feel free to ignore)

This is pretty much the point at which point writing this in a real programming language that has error handling, etc, is well worth the effort - Claude is pretty good to transpose logic. Python, Node, whatever.

rchlfryn: @stevekuznetsov happy to have another reviewer!

> This is pretty much the point at which point writing this in a real programming language that has error handling, etc, is well worth the effort - Claude is pretty good to transpose logic. Python, Node, whatever.

I am a bit confused on the ask, so say more..

stevekuznetsov: The logic is currently written as a shell script. I'd personally prefer to see a Node script, or a Python script, etc, that can implement the logic to interop with the APIs and manage the data, etc

rchlfryn: @stevekuznetsov I'm not familiar with scripts, so I just followed the pattern of our other scripts by writing it as a shell script. I'm trying to learn—are you suggesting using node/pythons scripts because I am using APIs here instead of the CLI?

.github/workflows/cleanup.yaml
rchlfryn: @busbyk after digging into the documentation more, looking at the deployments and aliases before and after & running the command , `vercel remove` does remove the deployment and its relative aliases even though it is not explicitly mentioned in the docs. Switching back to CLI

.github/workflows/cleanup.yaml
busbyk: I think @stevekuznetsov is just saying that once there starts to be conditional logic and things you'd want fine-grained error handling for it makes sense to use something besides a shell script to make that easier. 

The db sanitization, blob syncing, and media prefix updates that happen in `prod-to-dev-sync.yaml` are good examples of this: https://github.com/NWACus/web/blob/8b2259d0915bbf8311de4bbbc6041510f23e8c99/.github/workflows/sync-prod-to-dev.yml#L86-L103

.github/workflows/cleanup.yaml
```diff
@@ -89,11 +89,16 @@ jobs:
           set -o nounset
           set -o pipefail
 
-          BRANCH_NAME="${{ github.head_ref }}"
-          echo "Cleaning up deployments and aliases for branch: $BRANCH_NAME"
+          branch="${{ github.head_ref }}"
+          echo "Cleaning up deployments and aliases for branch: $branch"
+
+          if [ "$branch" = "main" ] || [ "$branch" = "release" ]; then
+            echo "[ERROR] Refusing to delete environment variables from production branch: ${branch}"
```
busbyk: Should this be "Refusing to delete deployments and aliases?"

busbyk (approved):
Looks good aside from one comment change

Issue #280, opened 2025-06-20
Create dashboard widget that allows embedding a PostHog widget


rustynwac:
A simple solution for showing PostHog analytics for tenants is creating a dashboard widget that let's us embed a PostHog dashboard that is built by the tenant admin or us.

busbyk:
There's a Request for Comments (RFC) open for a customizable dashboard in Payload: https://github.com/payloadcms/payload/discussions/11862

I think right now we'd have to completely replace the dashboard view (https://payloadcms.com/docs/custom-components/custom-views#root-views) which might be fine / preferable in some ways actually.

rchlfryn:
It looks like modular dashboards were released in https://github.com/payloadcms/payload/releases/tag/v3.69.0

rustynwac:
https://www.loom.com/share/a4bc998e7c944f548a5f5558ea7bf5c6

busbyk:
Custom fields added to the Payload modular dashboards feature: https://github.com/payloadcms/payload/pull/15700

PR #859, opened 2026-01-14
Dashboard widgets: Getting started widget

busbyk:
## Description

With #857 I hadn't realized that the modular dashboard widgets feature is always enabled now. Since we can't turn it off after version [3.69.0](https://github.com/payloadcms/payload/releases/tag/v3.69.0) we might as well convert our `beforeDashboard` custom component into a getting started widget. 

## Key Changes

- Removes the BeforeDashboard custom component completely. Previously this had a seed button for local development. I don't believe any of us are using this anymore so I removed it. Please let me know if you were still using this and want it added back to the GettingStartedWidget.
- Adds a GettingStartedWidget to the default dashboard widgets

## How to test

Load up the admin panel and test out the modular dashboard functionality. See [here](https://github.com/payloadcms/payload/pull/13683) for a demo video and discussion on the feature.

## Screenshots / Demo video

https://www.loom.com/share/11d40f3d264a4945bfcab3e8b47c0f57

We should add additional widgets! Tracking that on #280. Got ideas? Comment there.

Some change in the modular dashboard feature changed the behavior of the collections widget (i.e. the collection cards with "create" quicklinks (i.e. the plus buttons in the bottom right hand corner if the user has create access). Previously, if you selected a tenant, this would allow create access (due to https://github.com/NWACus/web/blob/main/src/access/byTenantRole.ts#L65) but now the collection cards are not updated unless you refresh the page manually. I believe we'll need to open a bug with Payload for this so keep an eye out for that.


rustynwac:
I think it looks great. My big picture thought is that modding the dashboard is a bit of a feature/nice to have that I was hoping to save for a little later in the spring. I have a bunch of ideas for #280 (90% of it being Posthog insights) but I don't think we need to prioritize til the spring. I need to get Posthog dialed first then can spend some time on the layout.

busbyk:
> I think it looks great. My big picture thought is that modding the dashboard is a bit of a feature/nice to have that I was hoping to save for a little later in the spring. I have a bunch of ideas for #280 (90% of it being Posthog insights) but I don't think we need to prioritize til the spring. I need to get Posthog dialed first then can spend some time on the layout.

Yea we're totally on the same page there. It's just that to use the latest Payload version we can't disable this new dashboard widgets feature so I figured we might as well make this quick cosmetic change. And it's a good example in the code of how to add a widget for when we get there.

rchlfryn (approved):
> I believe we'll need to open a bug with Payload for this so keep an eye out for that.

Not the biggest of issues since it is only for super admins. 

---
Get started widget looks great! That's exactly what I would have added so no notes. I think it is fine to not include a paragraph on how to remove it. We can include a message when we address #280

PR #862, opened 2026-01-14
Improve column selector for content blocks

### LINKED ISSUE #701: Blocks - content column picker is annoying
Whenever I add a new column in a content block, I need to update the column layout, but forget and it throws an annoying error. 

We should have a hook select the first available column layout for us.. Or change this UI to work better.

rchlfryn:
## Description
The column selector experience for content blocks is a bad UI & UX. This PR updates the column selector to automatically pick valid layout based on number of columns added. It will also hide any irrelevant column selections.

## Related Issues
Fixes #701 

## Key Changes
- Adds `useEffect` to listen for change in number of columns to automatically selects the first valid column layout when a column is added
- Hides any irrelevant/not valid column layouts that were previously displayed

## How to test
- Add a content block to a page
- Add/ remove columns and watch the column picker update

## Screenshots / Demo video
https://www.loom.com/share/9576fb618d7045ada752affafe0f00e9

### DIFF
--- a/src/components/ColumnLayoutPicker/index.tsx
+++ b/src/components/ColumnLayoutPicker/index.tsx
@@ -29,7 +29,7 @@ const ColumnLayoutPicker = (props: ColumnLayoutPickerProps) => {
 
   const { path, field } = props
   const { value, setValue } = useField<string>({ path })
-  const [currentValue, setCurrentValue] = useState(value || '1_1')
+  const [layoutSelection, setLayoutSelection] = useState(value || '1_1')
 
   const parentPath = path.split(`.${field.name}`)[0]
   const columnsPath = `${parentPath}.columns`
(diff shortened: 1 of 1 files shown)


src/components/ColumnLayoutPicker/index.tsx
```diff
@@ -4,33 +4,61 @@ import { ToggleGroup, ToggleGroupItem } from '@/components/ui/toggle-group'
...
   const parentPath = path.split(`.${field.name}`)[0]
   const columnsPath = `${parentPath}.columns`
 
   const { value: numColumns } = useField<number>({ path: columnsPath })
 
+  useEffect(() => {
+    if (value) {
+      setCurrentValue(value)
+    }
+  }, [value])
```
busbyk, line 43: ⛏️ Could be nice for readability to name `value` here as like `layoutValue` and the `useState` var as `setCurrentLayoutValue`. Not necessary but I had to think about what value was here haha.

busbyk (approved):
I love it! I really like the UX

PR #864, opened 2026-01-15
Add payload workaround issue

rchlfryn:
## Description
I wanted to make a space for us to keep track of the payload workaround we implement. Hopefully we wont use this issue that much, but I am just trying to make our lift easier to keep track of issues as payload fixes bugs.

~Add [Payload workarounds](https://github.com/NWACus/web/milestone/18) milestone to keep these in~

Thoughts on this? and the milestone?

### DIFF
--- a/.github/ISSUE_TEMPLATE/payload-workaround.md
+++ b/.github/ISSUE_TEMPLATE/payload-workaround.md
@@ -2,8 +2,7 @@
 name: Payload workaround
 about: Keep track of Payload workaround bugs
 title: ''
-labels: payload
-milestone: 'Payload workarounds'
+labels: payload-workaround
 ---
 
 **Describe the bug**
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/payload-workaround.md
busbyk, line 1: The idea is that we'd create one of these for each Payload workaround we introduce?

I think I'd prefer using a tag instead of a milestone to keep track of these. Don't want to start using milestones for everything. The key difference between using a milestone for bugs and tech debt and using one for this is that bugs and tech debt we're actively trying to burn down vs. these will be able to be addressed when / if Payload changes things.

.github/ISSUE_TEMPLATE/payload-workaround.md
```diff
@@ -0,0 +1,16 @@
+---
+name: Payload workaround
+about: Keep track of Payload workaround bugs
+title: ''
+labels: payload
```
busbyk: Yea I think let's make this `payload-workaround` and remove the milestone?

.github/ISSUE_TEMPLATE/payload-workaround.md
rchlfryn: The workaround was just a place to keep all the issues, but we can easily find them by using tags. I removed the milestone.

busbyk (approved):

PR #867, opened 2026-01-15
Appropriate image sizing

### LINKED ISSUE #563: Generate appropriate sizes/srcset to utilize Payload-generated image sizes
The current sizes generation is arbitrary - it's setting the suggested size to twice the size of the breakpoint. Next.js is using this sizes hint to generate the srcset used which is using Next Image's built-in image optimizations. 

But we're actually not even utilizing the different imageSizes generated by the Media collection 🤯

So there's two issues: our sizes hint is inaccurate and we're not even using the different image sizes generated by Payload in our srcsets. What we need to do is set up our sizes hint to tell Next Image to use our different image sizes which are actually different image urls instead of relying on Next.js image optimization.

busbyk:
## Description

This PR removes unused sizes generated by Payload, only keeping the thumbnail size for a more performant Payload admin panel experience. It then adds appropriate sizes (i.e. srcset) attributes for every usage of next/image in our codebase (let me know if I missed any!).

## Related Issues

Fixes #563 

## Key Changes

- Added context-aware `sizes` prop to all ImageMedia usages across the codebase
- Centralized breakpoints and container widths in `cssVariables.js` as single source of truth
- Reduced image quality from 100 to 80 (reduces bandwidth with minimal visual difference)
- Removed unused Payload image sizes (square, small, medium, large, xlarge, og)
- Fixed `sizes` prop formula in ImageMedia/AvatarImageMedia (was using `w` units incorrectly)
- Added `getImageWidthFromMaxHeight` utility for height-constrained images (Header, Footer, MobileNav, Carousel)

## How to test

- Upload a new image and see that only a thumbnail size is generated (along with the original)
- Inspect the `` `srcset` attributes for various components to see the size the browser loads

## Migration Explanation

Removed image sizes from Media collection. These were already unused - we were already only using next.js image optimization.


src/app/(frontend)/page.tsx
```diff
@@ -63,7 +63,7 @@ export default async function LandingPage() {
                   className="p-6 border rounded-lg hover:bg-gray-100 dark:hover:bg-gray-800 transition no-underline text-center"
                 >
                   <div className="aspect-square flex items-center justify-center w-48 mx-auto">
-                    <ImageMedia resource={logo} imgClassName="w-48" />
+                    <ImageMedia resource={logo} imgClassName="w-48" size="192px" />
```
rchlfryn: I am confused what size is actually doing here.. I would have expected it to define one size for next to return, but it still looks like it returns a srcset of image sizes

Image

I guess I am learning no matter the size defined (single or an array), next image will always produce `srcset`.

I wonder if it is worth it to update `ImageMedia` to use a regular `img` tag if we define a `size`.

src/blocks/ImageLinkGrid/Component.tsx
```diff
@@ -3,14 +3,37 @@ import { cn } from '@/utilities/ui'
...
+/**
+ * Generates the sizes attribute for responsive images based on column count.
+ * The sizes attribute tells the browser how wide the image will display at each viewport,
+ * allowing it to select the optimal image size from srcset before downloading.
+ *
+ * Layout behavior:
+ * - Below sm breakpoint: single column layout, image takes full viewport width
+ * - sm and above: multi-column grid, image width = container width / numOfCols
+ */
+const getImageSizes = (numOfCols: number): string => {
```
rchlfryn, line 26: is this worth making into a utility function? It seems like there are 3 different variations with some overlap, so I will leave it up to you.

src/collections/Media/index.ts
```diff
@@ -63,33 +64,6 @@ export const Media: CollectionConfig = {
         name: 'thumbnail',
         width: 300,
       },
-      {
```
rchlfryn, line 66: 👏🏼

rchlfryn (approved):
Overall looks good! This will be wonderful to have cleaned up. Approving but have a few comments that I am curious what you think

src/blocks/ImageLinkGrid/Component.tsx
busbyk: They're very similar but I felt like abstracting this into a common utility wasn't worth it. I wanted these to be specific for how each component is actually rendered. So I can't reason about how to make a flexible-enough common utility function.

src/app/(frontend)/page.tsx
busbyk: Yea great question. And I had to double check myself on this. 

Firstly, `size` is confusing so I just changed that to reflect the actual attribute name: `sizes`. [6fba3b0](https://github.com/NWACus/web/pull/867/commits/6fba3b06d6ba5f7b56d5c38791affda06a8f4920)

Next.js always generates the srcset which is basically a list of possible images for the browser to download. The sizes attribute indicates which one of those to pick from.

Some good reading:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Responsive_images#resolution_switching_different_sizes
- https://css-tricks.com/a-guide-to-the-responsive-images-syntax-in-html/#aa-using-srcset-w-sizes
- https://nextjs.org/docs/app/api-reference/components/image#sizes

Re: using a regular img tag - we still want to use NextImage to get optimized images.

PR #868, opened 2026-01-15
Sort International (INTL) state option to bottom of providers embed list

busbyk:
## Description

Previously, states were sorted alphabetically which placed INTL between IL (Illinois) and IN (Indiana). This change uses a custom sort comparator to ensure INTL always appears last while maintaining alphabetical order for all other states.

## Screenshots / Demo video

Check out: https://international-to-bottom.preview.avy-fx.org/embeds/providers


rchlfryn (approved):
LG

Bonus points: Can you update my mistake https://github.com/NWACus/web/blob/main/src/utilities/queries/getCoursesStates.ts#L35-L39 to use the sort function as well? At first I thought `findByIndex` was more efficient but I was wrong since there are multiple array mutations.

PR #873, opened 2026-01-20
Unify block naming

rchlfryn:
## Description
This PR unifies how we name our blocks per the [updated coding guide](/docs/coding-guide.md) and removes unused blocks

## Related Issues
None

## Key Changes

- Block folder name  ➡️ `SingleButton` 
- Block config name  ➡️ `SingleButtonBlock` 
- Block interface name  ➡️ `SingleButtonBlock` 
- Associated UI component(s) for the block ➡️ `SingleButtonBlockComponent` 
- Config with `wrapInContainer` ➡️  `SingleBlogPostLexicalBlock` 
- Removed `Banner`, `Biography`,`ImageQuote`, `ImageTextList` blocks

**Did not update**
- Block slug  ➡️ `singleButton` 
There are 8 (`button`, `callout`, `document`, `form`, `header`, `media`, `nacMedia`, `sponsors`) blocks that use `singleButtonBlock` naming convention. I do not think it is worth risking the migration to fix this.

## Migration Explanation
Drops blocks that were removed (`Banner`, `Biography`,`ImageQuote`, `ImageTextList`)


src/migrations/20260120_194629_unify_block_naming.ts
```diff
@@ -0,0 +1,395 @@
...
+  await db.run(sql`DROP TABLE \`pages_blocks_image_quote\`;`)
+  await db.run(sql`DROP TABLE \`pages_blocks_image_text_list_columns\`;`)
+  await db.run(sql`DROP TABLE \`pages_blocks_image_text_list\`;`)
+  await db.run(sql`DROP TABLE \`_pages_v_blocks_biography\`;`)
+  await db.run(sql`DROP TABLE \`_pages_v_blocks_image_quote\`;`)
+  await db.run(sql`DROP TABLE \`_pages_v_blocks_image_text_list_columns\`;`)
+  await db.run(sql`DROP TABLE \`_pages_v_blocks_image_text_list\`;`)
+}
+
+export async function down({ db, payload, req }: MigrateDownArgs): Promise<void> {
```
stevekuznetsov, line 22: This doesn't seem right, I don't see an actual data migration in here, so this is just data loss?

I wonder if there's a Drizzle config override we could use somewhere to force the new type to map to the old names

rchlfryn: @stevekuznetsov I think this is right - I removed 4 blocks so it makes sense to drop the tables and re add them in `up` if we need to revert the migration. What were you expecting that you are not seeing?

stevekuznetsov: Ah, sorry, misunderstood - I thought this was for the ones that were renamed.

docs/coding-guide.md
```diff
@@ -136,40 +136,46 @@ Use the `cn()` utility for conditional class names.
...
-Follow this example when creating new blocks:
+Follow this example when creating new block `SingleButton`:
+| | |
+|---|---|
+| Block folder name | `SingleButton` |
+| Block config name | `SingleButtonBlock` |
+| Block slug | `singleButton` |
+| Block interface name | `SingleButtonBlock` |
+| Associated UI component(s) for the block | `SingleButtonBlockComponent` |
+| Config with `wrapInContainer` | `SingleBlogPostLexicalBlock` |
```
busbyk, line 149: I think this was meant to be SingleButton not SingleBlogPost. But also wouldn't it make sense to be `SingleButtonLexicalBlock`?

docs/coding-guide.md
```diff
@@ -136,40 +136,46 @@ Use the `cn()` utility for conditional class names.
...
+### Blocks
 
-Follow this example when creating new blocks:
+Follow this example when creating new block `SingleButton`:
+| | |
+|---|---|
+| Block folder name | `SingleButton` |
+| Block config name | `SingleButtonBlock` |
+| Block slug | `singleButton` |
+| Block interface name | `SingleButtonBlock` |
```
busbyk, line 147: This is just so the generated type in payload-types.ts has "Block" at the end of it right?

busbyk (approved):
Love it!

Just one change suggested.

docs/coding-guide.md
rchlfryn: That is correct

docs/coding-guide.md
rchlfryn: Oops - dont know how that happened. Updated.

Issue #878, opened 2026-01-20
publishedAt field issues


busbyk:
1. If you change the publishedAt date and save a draft, the publishedAt field is cleared
2. If you unpublish, the publishedAt field should be cleared and is cleared in the database but it still shows the old value in the UI. Maybe a regression of https://github.com/payloadcms/payload/pull/10268?
3. The Pages collection's publishedAt field shows description "Set when this page was or should be published. This affects the page's visibility and can be used for scheduling future publications." but this is not true, we do not respect this for whether or not a page will render or show a 404. 
4. We're using three slightly different field configs for publishedAt in homePages, posts, and pages. 

homePages, posts, and pages have custom publishedAt fields. The only things we use publishedAt fields for in the (frontend) is sorting [[example](https://github.com/search?q=repo%3ANWACus%2Fweb%20-publishedAt&type=code)] and the displaying of dates (optionally based on showDates) for posts. 

**Why don't we just use updatedAt for those things instead and do away with the publishedAt field completely?**

updatedAt is a payload-managed field and in combination with status could easily serve the purpose of publishedAt.

busbyk:
## Investigation update

### Items 1, 3, 4 — Fixed in #1091
- Hook bug (clobbering user dates) fixed
- Misleading description removed from Pages
- Field configs standardized to `dayAndTime` across Posts, Pages, HomePages

### Item 2 — Stale UI after unpublish (Payload bug)
Root cause identified in Payload's `UnpublishButton` component (`@payloadcms/ui/dist/elements/UnpublishButton/index.js` lines 112-116):

```js
if (res.status === 200) {
  void resetForm({
    ...(dataFromProps || {}),  // stale data from initial page load
    _status: 'draft'           // only overrides status
  });
```

After unpublish, Payload resets the form with the **original page data** and only overrides `_status`. It never reads the server response (`res.json()` is not called on success), so hook-computed changes like `publishedAt: null` are discarded. The DB is correct — this is purely a client-side form state issue. A page reload shows the correct values.

This cannot be fixed with server-side hooks. It requires a Payload core fix to use the server response data when resetting the form.

busbyk:
Filed upstream: [payloadcms/payload/#16755](https://github.com/payloadcms/payload/issues/16755) — UnpublishButton does not reflect beforeChange hook modifications in form state. Reproduction repo: https://github.com/busbyk/unpublish-hook-stale-form-reproduction

PR #880, opened 2026-01-20
Remove vercel deployments for `main` branch

### LINKED ISSUE #790: Clean up Vercel

rchlfryn:
## Description
Updates `production.yml` `development.yml` and `sync-prod-to-dev.yml` to remove deployments from Vercel for `main` and `release` branch

## Related Issues
Fixes #790 

## Key Changes
Debugged in #876 #877 #878
Uses script that works from #856 #870
Added --safe to not delete any deployments with an alias. I did not add this to `cleanup,yml` since we will want to remove the deployments that have assigned preview aliases. This is just a guard rail for dev and prod
```md
➜  web git:(update-dev-gha) vercel remove "DEPLOYMENT_URL" --safe
Vercel CLI 48.12.0
> Could not find unaliased deployments or projects matching "DEPLOYMENT_URL". Run `vercel projects ls` to list.
```

## How to test
Look at previously run [development GHA](https://github.com/NWACus/web/actions/runs/21189866332/job/60953651279). This run had `vercel remove` commented out with debugging statements enabled to test

### DIFF
--- a/.github/workflows/production.yaml
+++ b/.github/workflows/production.yaml
@@ -88,3 +88,36 @@ jobs:
             --meta="githubRepo=${fullName##*/}" \
             --meta='githubCommitAuthorLogin=${{ github.event.pusher.name }}' \
             --token=${{ secrets.VERCEL_TOKEN }}
+      - name: Remove old Vercel deployments (keep latest 2)
+        run: |
+          set -o errexit
+          set -o nounset
+          set -o pipefail
+
+          ref="${{ github.event.ref }}"
+          branch="${ref##refs/heads/}"
+
+          echo "Cleaning up deployments and aliases for branch: $branch"
+
+          # All deployments for branch
+          deployments=$(vercel list --scope=nwac --meta githubCommitRef=$branch --token=${{ secrets.VERCEL_TOKEN }} 2>/dev/null | tail -n +2)
+
+          # Count total deployments
+          total=$(echo "$deployments" | grep -c . || true)
+
+          # Keep the last 1, delete the rest
+          if [ "$total" -gt 1 ]; then
+            echo "Keeping last 1 deployment, deleting $(($total - 1))"
+            echo "$deployments" | tail -n +2 | while read -r DEPLOYMENT_URL; do
+              if [ ! -z "$DEPLOYMENT_URL" ]; then
+                echo "Removing deployment: $DEPLOYMENT_URL"
+                vercel remove --scope=nwac "$DEPLOYMENT_URL" --safe --yes --token=${{ secrets.VERCEL_TOKEN }} || echo "[WARN] Failed to remove deployment: $DEPLOYMENT_URL"
+              fi
+            done
+          else
+            echo "Only 1 deployment found, nothing to delete"
+          fi
+        env:
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
--- a/.github/workflows/sync-prod-to-dev.yml
+++ b/.github/workflows/sync-prod-to-dev.yml
@@ -128,7 +128,7 @@ jobs:
             --meta="githubRepo=${fullName##*/}" \
             --meta='githubCommitAuthorLogin=${{ github.workflow }}' \
             --token=${{ secrets.VERCEL_TOKEN }}
-      - name: Remove old Vercel deployments (keep last 2)
+      - name: Remove old Vercel deployments (keep latest 2)
         run: |
           set -o errexit
           set -o nounset
@@ -156,7 +156,7 @@ jobs:
             echo "$deployments" | tail -n +2 | while read -r DEPLOYMENT_URL; do
               if [ ! -z "$DEPLOYMENT_URL" ]; then
                 echo "Removing deployment: $DEPLOYMENT_URL"
-                vercel remove --scope=nwac "$DEPLOYMENT_URL" --yes --token=${{ secrets.VERCEL_TOKEN }} || echo "[WARN] Failed to remove deployment: $DEPLOYMENT_URL"
+                vercel remove --scope=nwac "$DEPLOYMENT_URL" --safe --yes --token=${{ secrets.VERCEL_TOKEN }} || echo "[WARN] Failed to remove deployment: $DEPLOYMENT_URL"
               fi
             done
           else
--- a/.github/workflows/sync-prod-to-dev.yml
+++ b/.github/workflows/sync-prod-to-dev.yml
@@ -128,3 +128,41 @@ jobs:
             --meta="githubRepo=${fullName##*/}" \
             --meta='githubCommitAuthorLogin=${{ github.workflow }}' \
             --token=${{ secrets.VERCEL_TOKEN }}
+      - name: Remove old Vercel deployments (keep last 2)
+        run: |
+          set -o errexit
+          set -o nounset
+          set -o pipefail
+
+          ref="${{ github.event.ref }}"
+          branch="${ref##refs/heads/}"
+
+          echo "Cleaning up deployments and aliases for branch: $branch"
+
+          if [ "$branch" = "release" ]; then
+            echo "[ERROR] Refusing to delete deployments and aliases from production branch: ${branch}"
+            exit 1
+          fi
+
+          # All deployments for branch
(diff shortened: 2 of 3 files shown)


.github/workflows/sync-prod-to-dev.yml
```diff
@@ -128,3 +128,41 @@ jobs:
...
+          set -o errexit
+          set -o nounset
+          set -o pipefail
+
+          ref="${{ github.event.ref }}"
+          branch="${ref##refs/heads/}"
+
+          echo "Cleaning up deployments and aliases for branch: $branch"
+
+          if [ "$branch" = "release" ]; then
```
stevekuznetsov, line 142: If you target Bash, please prefer `[[` over `[`: https://stackoverflow.com/a/3427931

.github/workflows/sync-prod-to-dev.yml
```diff
@@ -128,3 +128,41 @@ jobs:
...
+            echo "[ERROR] Refusing to delete deployments and aliases from production branch: ${branch}"
+            exit 1
+          fi
+
+          # All deployments for branch
+          deployments=$(vercel list --scope=nwac --meta githubCommitRef=$branch --token=${{ secrets.VERCEL_TOKEN }} 2>/dev/null | tail -n +2)
```
stevekuznetsov, line 148: No way to get structured data out of `vercel list`? That's a shame. `tail -n +2` is fragile, if they decide that some other header format is more human-readable

.github/workflows/sync-prod-to-dev.yml
```diff
@@ -128,3 +128,41 @@ jobs:
...
+
+          # Count total deployments
+          total=$(echo "$deployments" | grep -c . || true)
```
stevekuznetsov, line 151: Are you counting lines? Prefer `wc`:

```
total=$( wc -l <<<"${deployments}")
```

.github/workflows/sync-prod-to-dev.yml
```diff
@@ -128,3 +128,41 @@ jobs:
...
+          # Keep the last 1, delete the rest
+          if [ "$total" -gt 1 ]; then
+            echo "Keeping last 1 deployment, deleting $(($total - 1))"
+            echo "$deployments" | tail -n +2 | while read -r DEPLOYMENT_URL; do
+              if [ ! -z "$DEPLOYMENT_URL" ]; then
+                echo "Removing deployment: $DEPLOYMENT_URL"
+                vercel remove --scope=nwac "$DEPLOYMENT_URL" --safe --yes --token=${{ secrets.VERCEL_TOKEN }} || echo "[WARN] Failed to remove deployment: $DEPLOYMENT_URL"
+              fi
+            done
+          else
```
stevekuznetsov, line 162: Negation of `-gt 1` is not 1 - what about 0?

.github/workflows/sync-prod-to-dev.yml
```diff
@@ -128,3 +128,41 @@ jobs:
...
+
+          # All deployments for branch
+          deployments=$(vercel list --scope=nwac --meta githubCommitRef=$branch --token=${{ secrets.VERCEL_TOKEN }} 2>/dev/null | tail -n +2)
+
+          # Count total deployments
+          total=$(echo "$deployments" | grep -c . || true)
+
+          # Keep the last 1, delete the rest
+          if [ "$total" -gt 1 ]; then
+            echo "Keeping last 1 deployment, deleting $(($total - 1))"
```
stevekuznetsov, line 155: Documentation and the `tail -n +2` below say you are keeping two, but here you say you're keeping one? Which is it?

.github/workflows/sync-prod-to-dev.yml
```diff
@@ -128,3 +128,41 @@ jobs:
...
+            echo "$deployments" | tail -n +2 | while read -r DEPLOYMENT_URL; do
```
stevekuznetsov, line 156: This `tail -n +2` implicitly assumes sort order for the output of `vercel list` - is that safe?

.github/workflows/sync-prod-to-dev.yml
```diff
@@ -128,3 +128,41 @@ jobs:
...
+              if [ ! -z "$DEPLOYMENT_URL" ]; then
+                echo "Removing deployment: $DEPLOYMENT_URL"
+                vercel remove --scope=nwac "$DEPLOYMENT_URL" --safe --yes --token=${{ secrets.VERCEL_TOKEN }} || echo "[WARN] Failed to remove deployment: $DEPLOYMENT_URL"
```
stevekuznetsov, line 159: Prefer 

```bash
if ! vercel remove ... ; then
    echo "[WARN] ..."
fi
```

.github/workflows/sync-prod-to-dev.yml
busbyk: TIL

.github/workflows/sync-prod-to-dev.yml
```diff
@@ -128,3 +128,41 @@ jobs:
...
             --meta='githubCommitAuthorLogin=${{ github.workflow }}' \
             --token=${{ secrets.VERCEL_TOKEN }}
+      - name: Remove old Vercel deployments (keep latest 2)
+        run: |
+          set -o errexit
+          set -o nounset
+          set -o pipefail
+
+          ref="${{ github.event.ref }}"
+          branch="${ref##refs/heads/}"
```
busbyk, line 138: I'm actually not sure the `github.event.ref` will be set here because this is cron-triggered (or manually triggered) - I think it will be undefined/empty. 

I think you would just want to omit the setting of ref at all and set branch explicitly to `main`. Then you could skip the release branch check too.

.github/workflows/production.yaml
```diff
@@ -88,3 +88,36 @@ jobs:
...
             --meta='githubCommitAuthorLogin=${{ github.event.pusher.name }}' \
             --token=${{ secrets.VERCEL_TOKEN }}
+      - name: Remove old Vercel deployments (keep latest 2)
+        run: |
+          set -o errexit
+          set -o nounset
+          set -o pipefail
+
+          ref="${{ github.event.ref }}"
+          branch="${ref##refs/heads/}"
```
busbyk, line 98: Same here I think it would be safe to set this explicitly to `release` since this workflow will only be triggered by the `release` branch.

busbyk (commented):
Going to defer to @stevekuznetsov on his evaluation of bash logic enhancements. 

Just a couple comments.

.github/workflows/sync-prod-to-dev.yml
rchlfryn: Nice - ended up removing this but also learned something new

.github/workflows/sync-prod-to-dev.yml
rchlfryn: Yes - updated

.github/workflows/sync-prod-to-dev.yml
rchlfryn: Keeping 2 - the one that is most recently deployed in the step before is not included in `vercel list` for some reason. So we will keep the current prod version and 1 older version. It might make sense to leave a more helpful comment.

.github/workflows/sync-prod-to-dev.yml
rchlfryn: Great question - and probably not but I don't know of another way. It does consistently return the deployments in order by date, but it is not great to assume in case they change it.  This is outside of my expertise. Is there a better way to do this??

.github/workflows/sync-prod-to-dev.yml
rchlfryn: Not that I saw. Their docs are not the most helpful. Very open to sugestions!

.github/workflows/sync-prod-to-dev.yml
rchlfryn: From what I understood, 0 would return false

.github/workflows/sync-prod-to-dev.yml
busbyk: Were you going to change these? Saw you re-requested reviews.

rchlfryn:
Merging for now. Since vercel cli is not the best we will keep this as a bash script for now, but should probably move it to a js script and use the API

PR #881, opened 2026-01-20
Update prefixFilenameWithTenant to work with Local API too

busbyk:
## Description

We depend on the `prefixFilenameWithTenant` hook to prefix our upload collections' file names with the tenant slug. We do this to avoid file name conflicts across tenants. 

I noticed that this wasn't happening for media created in the seed script. This was because this hook wasn't working as expected when using the [Local API](https://payloadcms.com/docs/local-api/overview). This could be problematic if we ever write custom media uploading logic using the Local API. Mostly a preventative measure. 

## Related Issues

None - noticed and fixed.

## Key Changes

- `req.data` did not exist for Local API requests
- Simplifies the hook using utility functions we've written since this hook was originally written
- Skips adding the prefix if there is already the same prefix at the start of the filename or the file starts with the tenant slug (without a dash) to avoid weird file names like `dvac-dvac-icon.png` or `dvac-dvac.png`

## How to test

- Verify that the tenant slug has been prepended to filenames created during seed in this PR's preview environment
- Upload an image or video in the Media collection - observe the tenant slug prefix in the file name
- Upload a document in the Documents collection - observe the tenant slug prefix in the file name

## Screenshots / Demo video

https://www.loom.com/share/d0f569da73eb4121a67ec9dd2f057f37

I'm not in love with the filename prefixing approach but that's a whole other conversation.


src/access/byGlobalRole.ts
```diff
@@ -11,8 +11,6 @@ export const byGlobalRole: (method: ruleMethod, collection: ruleCollection) => F
       return false
     }
 
-    payload.logger.debug(`evaluating access by ${user.id} for ${method} on ${collection}`)
```
rchlfryn, line 14: did you mean to remove this?

rchlfryn (approved):
LG - one comment

src/collections/Media/hooks/prefixFilenameWithTenant.ts
```diff
@@ -1,46 +1,41 @@
...
+    const tenant = await resolveTenant(media.tenant)
+
+    if (tenant) {
+      tenantSlug = tenant.slug
     }
   }
+
+  // Add tenant prefix if we have a slug and the filename doesn't already
+  // start with the prefix or start with the slug (i.e. the file name is the slug)
+  // Avoids file names like 'dvac-dvac-icon.png' and 'dvac-DVAC.webp'
```
stevekuznetsov, line 33: I get the reasoning here but I think you will appreciate not having to scratch your head in wonder when some seed script attempts to upload some dummy data with a center's ID as a name to more than one center and has it fail - "all uploads are prefixed to forbid collisions" is easier to grok than "some uploads can still collide". Do we show the filename to users, and/or if we do, do we remove the prefix for display? Would be another way to handle the stutter case.

src/endpoints/seed/upsert.ts
```diff
@@ -216,10 +216,13 @@ export async function upsert<TSlug extends TenantScopedCollectionWithHash>(
     payload.logger.info(
       `Creating ${collection}['${tenant}']['${key}'] with hash ${item.contentHash.slice(0, 8)}...`,
     )
+    // Clone the file object to prevent mutation issues when the same file is reused
+    // (e.g., the prefixFilenameWithTenant hook modifies file.name)
+    const fileToUse = file ? { ...file } : undefined
```
stevekuznetsov, line 221: Does this clone the underlying buffers, and, if so, did it have any appreciable impact to the RSS requirement for the seed?

stevekuznetsov (approved):

src/access/byGlobalRole.ts
busbyk: Yea - sorry should have explained my reasoning: I don't see the value in this statement. It doesn't tell us the outcome for one thing. And we would need to have logging set to debug mode for this to be logged at all in the first place. 

Do you or @stevekuznetsov have any thoughts on this?

src/collections/Media/hooks/prefixFilenameWithTenant.ts
busbyk: OK so in my own words you're saying:

> I get the reasoning here but I think you will appreciate not having to scratch your head in wonder when some seed script attempts to upload some dummy data with a center's ID as a name to more than one center and has it fail - "all uploads are prefixed to forbid collisions" is easier to grok than "some uploads can still collide".

Basically an argument that we should just be consistent with the logic rather than having this one exception to avoid file names with double tenant slugs just because they don't look great when we're already making the file names "not look great" anyways (haha). Yea I see the logic there. 

This logic should mean that if you upload a file "dvac.png" to: 
- dvac you would get "dvac.png"
- nwac you would get "nwac-dvac.png"

So I don't think we're running into the failures you think we would?

But your point about consistent logic still stands. Hmm. 

> Do we show the filename to users, and/or if we do, do we remove the prefix for display? Would be another way to handle the stutter case.

Yea this is my small attempt to make the filenames "look better" so I thought about stripping off the prefix on read too. But this would only work in the admin panel (which might be good enough). The prefixed filename is what's used in Vercel and so you would still need the prefixed filename to be displayed in the image's url like: `/api/media/file/dvac-some-filename.png`. I suppose the average user probably isn't going to notice that, though. 



I'm torn because the addressing of "ugly" file names probably makes sense to do as a larger effort where we evaluate if we can move away from filename prefixing and use a different approach (like folder prefixes or random suffixes). But these exceptions seem pretty reasonable to me and seem nice. Decisions! Thoughts?

src/endpoints/seed/upsert.ts
busbyk: What's RSS in this context? Re seed script? I can't imagine you mean really simple syndication 😂

src/access/byGlobalRole.ts
stevekuznetsov: If you haven't looked the outputs recently and it hasn't helped, delete!

src/endpoints/seed/upsert.ts
stevekuznetsov: Resident set! Are we holding on to a bunch of copies of the files we're uploading in memory during the script runtime?

src/collections/Media/hooks/prefixFilenameWithTenant.ts
stevekuznetsov: My personal take is that the value of simple logic without caveats outweighs the "stutter" on the file name, because I don't think that users would know, care, or mind too much, and because I find the added complexity of things like this to add up over time across the project until your brain melts trying to keep it all in. I agree that prefixing with a directory would be better, IIRC it was impossible when I looked into it originally.

Ultimately, it's a value judgement - take whichever path you'd like!

busbyk: OK I'm convinced. Avoiding the stutter case here is a half-baked attempt at making the filenames prettier after all so we might as well just keep it consistent and then consider alternatives for making the filenames prettier as a separate effort. 

And I think you're right that users probably won't care 😄

src/endpoints/seed/upsert.ts
busbyk: Aha! I believe the buffer is not copied. This is a shallow copy of the file object so the copied file object's buffer should be a reference to the original buffer. 

So no, shouldn't have any appreciable impact to the RSS requirement for the seed.

PR #886, opened 2026-01-22
Remove live preview

### LINKED ISSUE #869: Remove live preview

rchlfryn:
## Description
Since we removed autosave, we decided to remove live preview since it is a bad UX

## Related Issues
Fix #869


src/collections/Navigations/index.tsx
```diff
@@ -26,22 +22,6 @@ export const Navigations: CollectionConfig = {
     // the GlobalViewRedirect will never allow a user to visit the list view of this collection but including this list filter as a precaution
     baseListFilter: filterByTenant,
     group: 'Settings',
-    livePreview: {
```
busbyk, line 29: Do we want to add a `preview` attr here so the "open draft in new tab" button shows up?

busbyk (approved):
Looks great! 

Pushed one commit that remove two more references to live preview.

src/collections/Navigations/index.tsx
rchlfryn: Added in [60f3520](https://github.com/NWACus/web/pull/886/commits/60f352012a2ccf908dcd91784cbf0959fc9936b2)

busbyk (approved):

PR #892, opened 2026-01-26
Clean up how we use links

### LINKED ISSUE #685: Guidance on usage of CMSLink vs. Link + Button with clickWithCapture

### LINKED ISSUE #753: Button block doesn't have a thumbnail
The button block doesn't have a thumbnail:

Image

rchlfryn:
## Description
This PR cleans up for we use `Button` and `CMSLink` components. We currently use buttons in 3 different ways:
1. CMSLink - a guided way for a user to choose a link form a collection. Looks like a button but is rendered as an a tag
2. Link + Button - a link to another page or action
3. Button - not a link but performs an action

---
From the linked issue:
> This issue should consider the best approach to using links with button styling that also capture events.

I assume the second part of this is referring to Posthog events. All link clicks are autocaptured by posthog.

## Related Issues
Resolves #685 
Resolves #753 

## Key Changes
- Add `ButtonLink` component and consolidates `CMSLink` to `ButtonLink` since we were repeating the components pattern
- Removed unused button variants `destructive` and `link`
- Updated `EventTable` and `EventPreview` to use proper `Button` component instead of incorrectly used `CMSLink`
- Adds `clearIrrelevantLinkValues` to `linkField`
- Refactors `linkToPageOrPost` to `linkField`
  - Needed to keep `linkToPageOrPost` for `quickLinks` config since the type is `array` instead of `group`
- Clean up `defaultLexical` url validation  
- Updated coding guide

## How to test
There should be no visual differences

## Migration
Rename `button_appearance` to `button_variant`

We could use `linkField` in `navLink`, since the configs are identical. Using `linkField` will change `newTab` to be the second column which will cause a migration that will create a new table and drop the old one which seems invasive for the change. Do we want to do this?


src/fields/button.ts
```diff
@@ -39,17 +39,8 @@ export const appearanceOptions: Record<ButtonAppearances, { label: string; value
 type ButtonType = (appearances: ButtonAppearances[]) => Field
 
 export const button: ButtonType = (appearances) => {
-  const buttonResults: GroupField = {
-    name: 'button',
-    type: 'group',
-    admin: {
-      hideGutter: true,
-      style: {
-        marginBottom: '0',
```
rchlfryn, line 48: I did not copy this over as it is not necessary.

src/components/ui/button.tsx
```diff
@@ -21,9 +21,7 @@ const buttonVariants = cva(
       variant: {
         default: 'bg-primary text-primary-foreground hover:bg-primary-hover',
         secondary: 'bg-secondary text-secondary-foreground hover:bg-secondary-hover',
-        destructive: 'bg-destructive text-destructive-foreground hover:bg-destructive-hover',
         ghost: 'hover:bg-card hover:text-accent-foreground',
-        link: 'text-primary items-start justify-start underline-offset-4 hover:underline',
```
busbyk, line 26: This is one of those things where I could go back and forth on whether it makes sense to remove these or not a lot in my head haha. 

One the one hand I enjoy removing things that aren't used. On the other hand, since this is a shadcn/ui component a dev might expect to have these variants available to them. 

Buts:
- But it's common practice (and intended practice) to modify shadcn/ui components to meet your needs
- TS will tell you what variants are available

So idk where to land bc we could obviously just grab the code from shadcn if we ever want these variants back in the future.

src/fields/defaultLexical.ts
```diff
@@ -29,31 +29,21 @@ export const defaultLexical: Config['editor'] = lexicalEditor({
       BoldFeature(),
       ItalicFeature(),
       LinkFeature({
-        enabledCollections: ['pages', 'posts'],
+        enabledCollections: ['pages', 'posts', 'builtInPages'],
```
busbyk: We need to update https://github.com/NWACus/web/blob/link-cleanup/src/components/RichText/index.tsx#L58 to account for more than just pages and posts now. I _think_ we could use https://github.com/NWACus/web/blob/link-cleanup/src/utilities/handleReferenceURL.ts#L13 there. 

If you select a built-in page for an internal link the url is `/undefined` right now.

src/fields/defaultLexical.ts
```diff
@@ -29,31 +29,21 @@ export const defaultLexical: Config['editor'] = lexicalEditor({
...
-              // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-              validate: ((value, options) => {
-                // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-                if ((options?.siblingData as LinkFields)?.linkType === 'internal') {
-                  return true // no validation needed, as no url should exist for internal links
-                }
-                return value ? true : 'URL is required'
-              }) as TextFieldSingleValidation,
+              validate: validateExternalUrl,
             },
```
busbyk, line 64: We need a https://github.com/NWACus/web/blob/link-cleanup/src/utilities/collectionFilters.ts#L8 here, pages, posts, and builtInPages are showing for all tenants the user has permission to see not just the currently selected tenant. 

Image

src/blocks/Button/config.ts
```diff
@@ -1,8 +1,8 @@
-import { button } from '@/fields/button'
+import { buttonField } from '@/fields/button'
 import type { Block } from 'payload'
 
 export const ButtonBlock: Block = {
```
busbyk, line 4: Can we add a thumbnail for this block in this PR? 

Image

docs/coding-guide.md
```diff
@@ -182,3 +182,55 @@ blocks
     └── config.ts
 ```
 
+## Button vs CMSLink
```
busbyk: I'm not clear on the intent for Button, ButtonLink, and CMSLink from these docs / from usage in general. 

The use of Button is clear - it's just a button that's not a link. Although I'm surprised that PostHog autocaptures clicks on buttons that aren't wrapped in links / have a <a> child. Is that true?

The difference in usage of ButtonLink and CMSLink is fuzzy. They both render a Button component with a Link inside it with pretty much the same props. 

The only things CMSLink does differently is:
- Accepts an internal reference (a Payload document from pages, posts, or builtInPages) and uses handleReferenceURL to determine the internal url
- Uses a custom PostHog event capture which includes the tenant 

I get that ButtonLink is specifically for external URLs but I'm definitely feeling like these could just be a single component. 

I wonder if we wouldn't want that custom PostHog event capture with tenant context on external links too. 

Using CMSLink (the naming specifically being the issue) is confusing for explicitly external links but I think passing a reference (i.e. Payload document) into a ButtonLink component wouldn't be too confusing. Maybe `reference` isn't the best prop name but 🤷. 

I think it would be clearer to just have a single ButtonLink which should be used for all links (internal and external) that should be styled as buttons. If you want a link that's not styled as a button you can use a pattern similar to ImageLinkGridBlock where you use `handleReferenceURL` and `Link`. 

I don't think we even really need to call out the usage of Button here necessarily except maybe to say: don't use an `onClick` handler to programmatically navigate unless there's a use case that makes sense but...that could go without saying idk. 

Thoughts?

busbyk (commented):
Left some comments and had an idea: https://github.com/NWACus/web/pull/900

Psyched to get this consolidated/consistent.

src/components/ui/button.tsx
rchlfryn: Ah good call out. I thought we added these but we did not (https://github.com/shadcn-ui/ui/blob/main/apps/v4/examples/base/ui/button.tsx)

I do not really like link because I think a button that looks like a link should just be a link. Someone needs to decide..

I propose leaving them in here but not making them an option for the user.

src/fields/defaultLexical.ts
rchlfryn: Ooo great catch! Updated in [849808e](https://github.com/NWACus/web/pull/892/commits/849808efd50bfe554c909db07954123f2b9a14c4) and also added a filter to only get the current tenant.
I decided to us type casting instead of changing the types of `handleReferenceURL` - let me know if you think we should change the types instead.

src/fields/defaultLexical.ts
rchlfryn: See comment above

src/blocks/Button/config.ts
rchlfryn: Done in [474ba26](https://github.com/NWACus/web/pull/892/commits/474ba264917fc5613996a1208c073682c05c6ee2). Also update thumbnails for single blog post and single event because it was bothering me. Scope creep!

src/components/ui/button.tsx
busbyk: > I propose leaving them in here but not making them an option for the user.

Cool, I'm good with that.

src/fields/defaultLexical.ts
busbyk: I think we should [use a typeguard function instead of type assertions](https://github.com/NWACus/web/blob/main/docs/coding-guide.md#L36): https://github.com/NWACus/web/pull/901

docs/coding-guide.md
rchlfryn: Great thoughts, I consolidated CMSLink into Button link in [c452956](https://github.com/NWACus/web/pull/892/commits/c4529569f5c223e350a8097ccc6982893aad8f8f)

> The use of Button is clear - it's just a button that's not a link. Although I'm surprised that PostHog autocaptures clicks on buttons that aren't wrapped in links / have a child. Is that true?

Yes that is true - check out this [dashboard](https://us.posthog.com/project/40776/activity/explore#q=%7B%22kind%22%3A%22DataTableNode%22%2C%22full%22%3Atrue%2C%22source%22%3A%7B%22kind%22%3A%22EventsQuery%22%2C%22select%22%3A%5B%22*%22%2C%22event%22%2C%22person_display_name%20--%20Person%22%2C%22coalesce(properties.%24current_url%2C%20properties.%24screen_name)%20--%20Url%20%2F%20Screen%22%2C%22properties.%24lib%22%2C%22timestamp%22%5D%2C%22orderBy%22%3A%5B%22timestamp%20DESC%22%5D%2C%22after%22%3A%22-1h%22%2C%22event%22%3A%22%24autocapture%22%7D%2C%22propertiesViaUrl%22%3Atrue%2C%22showSavedQueries%22%3Atrue%2C%22showPersistentColumnConfigurator%22%3Atrue%7D)

> The only things CMSLink does differently is:
> 
> * Accepts an internal reference (a Payload document from pages, posts, or builtInPages) and uses handleReferenceURL to determine the internal url
> * Uses a custom PostHog event capture which includes the tenant
> I get that ButtonLink is specifically for external URLs but I'm definitely feeling like these could just be a single component.
> 

I was trying for `CMSLink` to handle anything that had an admin interface and `ButtonLink` was anything we linked to in the code, but I forgot `CMSLink` can have an external URL. As I mentioned everything has been consolidated in [c452956](https://github.com/NWACus/web/pull/892/commits/c4529569f5c223e350a8097ccc6982893aad8f8f)


> Using CMSLink (the naming specifically being the issue) is confusing for explicitly external links but I think passing a reference (i.e. Payload document) into a ButtonLink component wouldn't be too confusing. Maybe reference isn't the best prop name but 🤷.

`LinkFeature` uses `doc`. Take a look and let me know if we should rename the prop.


> I don't think we even really need to call out the usage of Button here

Agreed and updated in [c452956](https://github.com/NWACus/web/pull/892/commits/c4529569f5c223e350a8097ccc6982893aad8f8f)

PR #893, opened 2026-01-27
Add production env to preview.yml

### LINKED ISSUE #713: Set NODE_ENV in preview.yaml
We ran into a bug in the preview environments for A3 embeds because our seed script uses the getURL utility which [determines the protocol of the url based on the NODE_ENV](https://github.com/NWACus/web/blob/main/src/utilities/domain.ts#L1).

Looking into that, I noticed that the NODE_ENV is not set for our preview workflow. I initially thought that this wasn't necessary and [tried setting NODE_ENV === 'production' like the rest of our workflows but we get errors during the build](https://github.com/NWACus/web/pull/708). I wonder if this has something to do with the `getEnvironmentFriendlyName` utility or maybe we're missing some environment variable in the preview env that we need when NODE_ENV === 'production'. 

https://github.com/search?q=repo%3ANWACus%2Fweb%20NODE_ENV%20%3D%3D%3D%20%27production%27&type=code

I'm recording this here rather than investigating since this is non-urgent and we have higher priorities.

rchlfryn:
## Description
Set node env to production for `preview.yml`

## Related Issues
Fixes #713

## Key Changes
- Added env to script
- Ran migrations (same as prodctuon.yml)

## How to test
GHA


busbyk (approved):
Nice 👏

PR #894, opened 2026-01-27
Allow providers to edit and delete users they have access to

### LINKED ISSUE #834: Allow A3 Provider Manager to edit/delete Provider Users
There are often delinquent accounts or account changes needed for Providers and A3 would like to be able to make necessary changes.

rchlfryn:
## Description
As mentioned, we want to give a provider manager permission to:
- read any user with a providers relationship
- edit/delete that is not associated with an avy center

## Related Issues
Resolves #834 

## Key Changes
Changed permissions for provider managers on `users` collection and add beforeChange and beforeDelete hooks

## How to test
- Log in as provider manager
- Navigate to users
- Select user with checkmark
- See options to `Edit` and `Delete` 

## Screenshots / Demo video
https://www.loom.com/share/9a79a65e6cec4e35888171620ed448e4

## Migration Explanation
None

### DIFF
--- a/src/collections/Users/access/byGlobalRoleOrTenantRoleAssignmentOrDomain.ts
+++ b/src/collections/Users/access/byGlobalRoleOrTenantRoleAssignmentOrDomain.ts
@@ -1,4 +1,5 @@
 import { byGlobalRole } from '@/access/byGlobalRole'
+import { getDocumentById } from '@/utilities/getDocumentById'
 import { isProviderManager } from '@/utilities/rbac/isProviderManager'
 import { roleAssignmentsForUser } from '@/utilities/rbac/roleAssignmentsForUser'
 import { ruleMatches, ruleMethod } from '@/utilities/rbac/ruleMatches'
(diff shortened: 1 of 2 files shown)


busbyk (dismissed):
This change would allow a provider manager to delete/edit any user with a provider relationship regardless of that user's roles or global roles. 

This is an issue because we currently have some Sierra admin users with provider relationships to the Sierra Avalanche Center _provider_ in production. So this change would allow a provider manager user to delete a Sierra admin user. Which shouldn't be allowed. 

I think what we want is logic that says: a provider manager can:
- read any user with a providers relationship
- edit/delete any user with _only_ a provider relationship and no roles or global roles

_Maybe_ we go to the field level and allow a provider manager to edit a user's providers relationship field if that user already has a provider relationship but that might not be worth it. 

Test by adding a providers relationship to a tenant-admin user in the local seed data. The provider manager user can then delete that user.

rchlfryn:
> I think what we want is logic that says: a provider manager can:
> - read any user with a providers relationship
> - edit/delete any user with only a provider relationship and no roles or global roles

I am not sure if there is a way we can do this at the access level, since the data being deleted/edited is not available to look at. I added 2 hooks to the user collection to check the user being edited or deleted and not let a provider manager delete a user associated with an avy center. 

Is there another way to do this?

PR #895, opened 2026-01-27
Update validate functions to use default field validations

### LINKED ISSUE #860: Ensure all custom field validations are using default field validations
**Describe the bug**
When working on the pages collection, I noticed a user can now create a page without any blocks in the layout field. This is because we added a [custom validation](https://github.com/NWACus/web/blob/main/src/collections/Pages/index.ts#L112) that by default will use what we write in place of the default. Payload has a great solution for this: [Reusing default field validations](https://payloadcms.com/docs/fields/overview#reusing-default-field-validations)

We should double check all custom field validations to ensure they are reusing default field validations

rchlfryn:
## Description
We recently discovered that you can reuse payload default validation function in a custom validation function. This PR does that.

## Related Issues
Fixes #860 

## How to test
Test admin fields


src/collections/Sponsors/index.ts
```diff
@@ -4,22 +4,25 @@ import { contentHashField } from '@/fields/contentHashField'
...
-  if (!value) return true
+const validateDate: DateFieldValidation = (
+  value,
+  args: ValidateOptions<unknown, unknown, DateField, Date> & { siblingData: Partial<Sponsor> },
+) => {
+  if (!value) return date(value, args)
 
-  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-  const sponsorData = data as Partial<Sponsor>
+  const { siblingData: sponsorData } = args
```
busbyk: Heh it's unexpected to me that siblingData and data have the same data here apparently...

busbyk (approved):

src/collections/Sponsors/index.ts
rchlfryn: Woof - good catch. Updating in [c21d519](https://github.com/NWACus/web/pull/895/commits/c21d519043ad61b73a3f42afdaada4e8e228e13c)

PR #896, opened 2026-01-27
Remove some type assertions

rchlfryn:
## Description
I was messing around and realized we had some low hanging fixes to remove type assertions.

## Key Changes
Refactor code to remove `as` and re ran `pnpm verify-type-assertions`

## How to test
App loads and clean build


src/app/(frontend)/[center]/blog/page.tsx
```diff
@@ -69,8 +69,7 @@ export async function generateMetadata(
   parent: Promise<ResolvedMetadata>,
 ): Promise<Metadata> {
   const { center } = await props.params
-  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-  const parentMeta = (await parent) as Metadata
+  const parentMeta = await parent
```
Comment (busbyk) [inline src/app/(frontend)/[center]/blog/page.tsx:72]: Ooo next.js fixed this type?? Sweet!

src/utilities/findDocumentsWithBlockReferences.ts
```diff
@@ -3,6 +3,14 @@ import { getPayload } from 'payload'
 import { getBlocksFromConfig } from './getBlocksFromConfig'
 import { DocumentForRevalidation, RevalidationReference } from './revalidateDocument'
 
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
```
busbyk: The checks this is replacing didn't check for `!Array.isArray(value)`. Did you test this function to make sure it's still finding all expected block references? 

I think it's fine based on how it's used here but that is the one thing that jumped out at me here.

busbyk (approved):

src/utilities/useClickableCard.ts
busbyk, line 1: This hook is only used in `src/components/Card/index.tsx` which isn't used at all. Should we just remove both of those? I've always felt like this hook is confusing. It's also an anti-pattern because it's intent (as I understand it) is to allow you to have a link with a different href nested inside of another link and browsers don't like that.

src/utilities/findDocumentsWithBlockReferences.ts
rchlfryn: I think this was Claudes attempt at the `else` for `if (Array.isArray(fieldValue)) {` I am going to remove it

src/utilities/useClickableCard.ts
rchlfryn: Let's remove it

Issue #579, opened 2025-10-05
Refactor tenant lookup to use slug instead of id


busbyk:
We currently use the tenant id for our multi tenancy functionality (i.e. determining the tenant scoping of the admin panel and for routing in the middleware). 

This is mainly due to the fact that the Payload multi tenant plugin defaulted to using a number ID. 

We built an [Edge Config tenant lookup](https://github.com/NWACus/web/blob/main/docs/decisions/008-edge-config-tenant-lookup.md) to make tenant id lookups really fast in our middleware. 

Since [we're moving off / have moved off of](https://github.com/NWACus/web/issues/552) the Payload multi tenant plugin, we could potentially get rid of that entire Edge Config integration and hard code all of the avalanche center slugs. 

This would simplify our codebase in many ways but would require updating tenant references in many places and doing thorough regression testing.

busbyk:
Closed by #897

Issue #866, opened 2026-01-15
Baseline integration test suite using Playwright or Puppeteer


busbyk:
We should establish a basic but relatively thorough integration test suite using Playwright or Puppeteer. I primarily want this for knowing if I've broken something after making a change so maybe we build this out as a regression test suite initially. 

I'm thinking basic functionality checks like:
- user can log in
- user can visit the posts collection list
- user can create a new post
- user can delete a post
- user can edit a post

We could start with just testing the Payload admin panel and test basics for a couple roles like super admin, single tenant admin, provider manager, provider.

This sounds like a lot but once we establish a pattern we can use Claude to build out the full suite.

We can take a look at Payload's test suite for inspiration on approach here too. [test dir](https://github.com/payloadcms/payload/tree/main/test) [relevant docs from their contributing.md](https://github.com/payloadcms/payload/blob/main/CONTRIBUTING.md#code)

PR #897, opened 2026-01-27
Tenant lookup refactor

busbyk:
## Description

See decision doc `docs/decisions/013-hardcoded-tenant-lookup.md` for full details. 

Refactors tenant lookup to use a hardcoded list and use slug in the `payload-tenant` cookie instead of id. 

## Related Issues

#579 

## Key Changes

See decision doc `docs/decisions/013-hardcoded-tenant-lookup.md`.

## How to test

See `docs/tenant-selector.md` for manual test cases.

## Screenshots / Demo video

Demo: https://www.loom.com/share/4c083b90aedf4df983eab130db9b1783

Not going to lie: the demo is a little boring. I'm just running through the manual test cases. Might be more helpful for you to run through on your own but def feel free to just watch the loom on 2x speed instead. 

I'm starting to think that we shouldn't allow visiting the admin panel on subdomains and custom domains at all. We already direct users to use the root domain in production. Let's just add a redirect to the root domain for `/admin` routes.


src/utilities/tenancy/avalancheCenters.ts
```diff
@@ -0,0 +1,83 @@
...
+  coaa: { name: 'Central Oregon Avalanche Center', customDomain: 'www.coavalanche.org' },
+  dvac: { name: 'Death Valley Avalanche Center', customDomain: 'www.avy-fx-demo.org' }, // The "template tenant" - not a real avalanche center
+  earac: { name: 'Eastern Alaska Range Avalanche Center', customDomain: 'alaskasnow.org' },
+  esac: { name: 'Eastern Sierra Avalanche Center', customDomain: 'www.esavalanche.org' },
+  ewyaix: { name: 'Eastern Wyoming Avalanche Info Exchange', customDomain: 'ewyoavalanche.org' },
+  fac: { name: 'Flathead Avalanche Center', customDomain: 'www.flatheadavalanche.org' },
+  gnfac: { name: 'Gallatin NF Avalanche Center', customDomain: 'www.mtavalanche.com' },
+  hac: { name: 'Haines Avalanche Center', customDomain: 'alaskasnow.org' },
+  hpac: { name: 'Hatcher Pass Avalanche Center', customDomain: 'hpavalanche.org' },
+  ipac: {
```
stevekuznetsov, line 31: Oh prettier, why do you injure me so

src/utilities/tenancy/avalancheCenters.ts
```diff
@@ -0,0 +1,83 @@
...
+export const VALID_TENANT_SLUGS: ValidTenantSlug[] =
+  Object.keys(AVALANCHE_CENTERS).filter(isValidTenantSlug)
+
+/**
+ * Lookup a center by its custom domain
+ */
+export function findCenterByDomain(domain: string): ValidTenantSlug | undefined {
+  // Normalize both input and stored domains by removing www. prefix
+  // Regex matches 'www.' at start of string
+  const normalizedInput = domain.replace(/^www\./, '')
```
stevekuznetsov: Is it possible for a user's casing to flow into here, or are we guaranteed lower-case by the time we are processing here?

src/utilities/useTenantLookup.ts
```diff
@@ -0,0 +1,63 @@
...
+          { credentials: 'include' },
+        )
+        const result = await response.json()
+        if (result.docs && result.docs.length > 0) {
+          const id = result.docs[0].id
+          setTenantIdBySlug((prev) => ({ ...prev, [slug]: id }))
+          return id
+        }
+      } catch (_) {
+        // Silent fail - caller should handle missing tenant
```
stevekuznetsov: Could you expand a little on what this means? I see callers like `src/fields/tenantField/TenantFieldComponent.tsx` only taking action when the return is truthy, but it's not clear why we want to elide errors other than not-found - which I assume are part of what this catch will silently eat?

stevekuznetsov (approved):

src/utilities/tenancy/avalancheCenters.ts
busbyk: Oo good question - I hadn't considered that. Just tested locally and it appears the browser rewrites the domain to lowercase even if you include uppercase characters.

But it can't hurt to just throw a `.toLowerCase()` on here.

src/utilities/useTenantLookup.ts
busbyk: Maybe we should log something here instead. But for the `TenantFieldComponent` caller, the `tenantField` has a `beforeChange` hook that will handle a missing tenant value in the case that this hook fails: https://github.com/NWACus/web/blob/tenant-lookup-refactor/src/fields/tenantField/index.ts#L34

The `DuplicatePageForDrawer` caller doesn't really handle this actually, now that I'm looking at this again. The impact of this failing would mean that `DuplicatePageForDrawer` doesn't filter out the current tenant and a user could duplicate the page to the _same_ tenant. This isn't really a big deal and it would be worse behavior if this failing caused the user not to be able to duplicate at all. Because if this fetch fails, they could still duplicate to another tenant too. Hopefully that's clear. 

I think it probably makes sense to add a Sentry.captureException here and still fail silently to the user.

busbyk: [9706393](https://github.com/NWACus/web/pull/897/commits/9706393499b09ab0eae7cdc466a01d9e6db2bf29)

src/utilities/tenancy/avalancheCenters.ts
busbyk: [dc52aff](https://github.com/NWACus/web/pull/897/commits/dc52aff1e2df1077e2651b58089e81f88298de88)

docs/decisions/013-hardcoded-tenant-lookup.md
```diff
@@ -0,0 +1,80 @@
...
+
+Status: accepted
+
+Supersedes: [008-edge-config-tenant-lookup.md](./008-edge-config-tenant-lookup.md)
+
+## Context
+
+We're essentially migrating back to the approach we used before [007-dynamic-tenants-middleware.md](./007-dynamic-tenants-middleware.md) which is to have a hardcoded tenant list. We moved to a dynamic tenant lookup because we were using the multi-tenant plugin from Payload at that point which needed to use the tenants collection's id for the tenant cookie.
+
+Since we [moved away from the multi-tenant plugin](./010-remove-multi-tenant-plugin.md), we can use a hardcoded list of all possible avalanche centers and use the slug for the `payload-tenant` cookie because we control the code. The biggest issue that [007-dynamic-tenants-middleware.md](./007-dynamic-tenants-middleware.md) accomplished was to avoid issues where ids were different between environments due to order in which they were created / the tenants that were actually created in a given environment. This isn't an issue when using the slug.
```
rchlfryn, line 13: This is a nice win!

docs/tenant-selector.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Test Matrix Dimensions
+
+- **Collection types:** Tenant-required, Global collection, Non-tenant, Payload global
+- **Roles:** Super admin, Multi-center admin, Single-center admin
+- **Domains:** Root domain, Tenant subdomain
+
+---
+
+### Tenant-Required Collection (e.g., Pages, Posts)
```
rchlfryn, line 67: ⛏️ ⛏️ ⛏️ 

```suggestion
### Tenant-Required Collection (e.g. Pages, Posts)
```

docs/tenant-selector.md
```diff
@@ -0,0 +1,203 @@
...
+- [ ] Should be able to switch between their tenants
+- [ ] Should NOT see tenants they don't have access to in the dropdown
+- [ ] Attempting to manually navigate to a document from an unauthorized tenant should result in access denied
+
+### Single-Center Admin (user with role in exactly 1 tenant)
+
+- [ ] The tenant selector should be hidden (only 1 option available)
+- [ ] The tenant cookie should automatically be set to their single tenant's value
+- [ ] All tenant-scoped operations should use their single tenant
+
```
rchlfryn, line 157: Do we want to add provider manage here or make note of it?

src/collections/Tenants/index.ts
```diff
@@ -53,11 +42,14 @@ export const Tenants: CollectionConfig = {
     },
     {
       name: 'slug',
-      type: 'text',
+      type: 'select',
```
rchlfryn, line 45: I like this change!

src/utilities/tenancy/getGlobalViewRedirect.ts
```diff
@@ -31,14 +30,21 @@ export async function getGlobalViewRedirect({
...
-  })
-  let tenant = getTenantFromCookie(headers, idType)
-  let redirectRoute: `/${string}` | void = undefined
+  // Get tenant slug from cookie and look up tenant ID
+  const tenantSlug = getTenantSlugFromCookie(headers)
+  let tenantId: number | string | null = null
 
-  if (!tenant) {
+  if (tenantSlug) {
+    const { docs } = await payload.find({
```
rchlfryn, line 38: This seems to be repeated a few time through out the code. Might be a good chance to make it a utility function
```ts
export const getTenantIdFromSlug = async (slug: string) => {
  const result = await payload.find({
    collection: 'tenants',
    where: { slug: { equals: slug } },
    limit: 1,
  })
  return result.docs[0]
}
```

src/utilities/useTenantLookup.ts
```diff
@@ -0,0 +1,64 @@
...
+  const [tenantSlugById, setTenantSlugById] = useState<Record<number, string>>({})
+
+  const lookupTenantIdBySlug = useCallback(
+    async (slug: string): Promise<number | null> => {
+      if (tenantIdBySlug[slug]) {
+        return tenantIdBySlug[slug]
+      }
+
+      try {
+        const response = await fetch(
```
rchlfryn, line 21: Any reason we are making this as a client component and using fetch?

src/utilities/useTenantLookup.ts
```diff
@@ -0,0 +1,64 @@
...
+  )
+
+  const lookupTenantSlugById = useCallback(
+    async (id: number): Promise<string | null> => {
+      if (tenantSlugById[id]) {
+        return tenantSlugById[id]
+      }
+
+      try {
+        const response = await fetch(`/api/tenants/${id}?depth=0`, { credentials: 'include' })
```
rchlfryn, line 46: Same question about fetch

consistent-type-assertions.txt
rchlfryn, line 1: 🥳

docs/tenant-selector.md
rchlfryn, line 1: 👏🏼

rchlfryn (commented):
PR looks good - minor comments

One major difference I was thinking before reviewing this PR is we were going to migrate away from using ids and use the slug as the primary key/id. I dug into it and we should not do that since Payload expects id and Payload's core functionality relies on the id field being the primary key. Maybe we add this to the decision doc?


A few questions:

What will happen for current logged in users - will the tenant cookie just be updated? or are they forced to re-login

Can we update the [onboarding docs](https://github.com/NWACus/web/blob/main/docs/onboarding.md) where adding the slug to the edge config in production? Are there any other docs we need to update?

src/utilities/useTenantLookup.ts
busbyk: Yea good thought - this is used in the TenantFieldComponent which is a client component so we need to use fetch/the REST API.

docs/tenant-selector.md
busbyk: Yea we def should - although I guess maybe it makes more sense to expand this into RBAC docs not just tenant-selector specific. I'll think about this for my PR for #866

src/utilities/tenancy/getGlobalViewRedirect.ts
busbyk: Good call, I think that would be nice. I'll see about a follow up PR.

PR #900, opened 2026-01-28
Use a builder function to generate all of the three link field variants

busbyk:
An idea for consolidation. Is this more or less complicated?


rchlfryn (approved):
I am all for consolidation so I think this is great. I will mark as blocked until #892 is fully reviewed ready to be merged.

PR #901, opened 2026-01-28
Use type guard instead of type assertions

busbyk:
^^


rchlfryn (approved):
🥳 don't forget `pnpm verify-type-assertions`

PR #936, opened 2026-02-12
Update timezone to use name instead of GMT-

rchlfryn:
## Description
Fix timezone display to show human-readable abbreviations (e.g. "PST") instead of GMT offsets (e.g. "GMT-8"). Also improve date range formatting for multi-day events combining the year.

## Key Changes
- Update `formatDateTime` to use `tzName()` for short timezone names instead of date-fns default GMT offset format
- Use custom `zzz` token in format strings to trigger timezone name rendering
- Simplify `StartAndEndDateDisplay` — omit redundant year for same-year date ranges, early return for same-day events
- Add calendar icon to `EventPreviewSmallRow` date display

## How to test
1. View events with start/end dates in various scenarios:
   - Single-day event (same start/end date) — should show `MMM d, yyyy, time - time TZ`
   - Multi-day event, same year — should omit year from start date
   - Multi-day event, different years — should show year on both dates
2. Verify timezone shows as e.g. "PST" not "GMT-8"

## Screenshots / Demo video
| Before | After |
|--------|--------|
| Screenshot 2026-02-12 at 14 53 03 | Screenshot 2026-02-12 at 14 58 48 |
| Screenshot 2026-02-12 at 14 53 13 | Screenshot 2026-02-12 at 14 59 02|
| Screenshot 2026-02-12 at 14 53 22 | Screenshot 2026-02-12 at 14 59 09 |

### DIFF
--- a/src/utilities/formatDateTime.ts
+++ b/src/utilities/formatDateTime.ts
@@ -1,4 +1,4 @@
-import { TZDate } from '@date-fns/tz'
+import { TZDate, tzName } from '@date-fns/tz'
 import { format } from 'date-fns/format'
 
 export const formatDateTime = (
@@ -6,5 +6,9 @@ export const formatDateTime = (
   tz: string | null | undefined,
   formatStr: string,
 ) => {
-  return tz ? format(new TZDate(dateString, tz), formatStr) : format(dateString, formatStr)
+  const date = tz ? new TZDate(dateString, tz) : new Date(dateString)
+
+  return tz && formatStr.includes('zzz')
+    ? `${format(date, formatStr.replace(/zzz/g, ''))} ${tzName(tz, date, 'short')}`
+    : format(date, formatStr)
 }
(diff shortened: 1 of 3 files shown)


src/utilities/formatDateTime.ts
```diff
@@ -1,10 +1,14 @@
-import { TZDate } from '@date-fns/tz'
+import { TZDate, tzName } from '@date-fns/tz'
 import { format } from 'date-fns/format'
```
stevekuznetsov, line 2: Import format from the tz size and `zzz` does what you're trying to do natively it looks like https://www.npmjs.com/package/date-fns-tz#format

rchlfryn: `@date-fns/tz` and `date-fns-tz` are different packages. We switched with https://github.com/NWACus/web/pull/851

rchlfryn: Linking 2 active issues with the package 
- https://github.com/date-fns/date-fns/issues/3975
- https://github.com/date-fns/tz/issues/54

PR #950, opened 2026-02-18
Exclude /ingest from middleware matcher

rchlfryn:
## Description
Exclude PostHog `/ingest` paths from middleware to reduce unnecessary edge invocations and Vercel usage.

I was able to confirm on Posthog that page views events alone. This might be related to the current avalanche warnings in Tahoe.
Screenshot 2026-02-18 at 10 29 21

## Related Issues
Production usage alerts — Edge Requests up 4.82x, Function Invocations up 4.37x, CPU Duration up 2.77x, Data Transfer up 5.37x.

## Key Changes
- Added `ingest` to the negative lookahead regex in the middleware matcher
- PostHog analytics traffic (`/ingest/s/`, `/ingest/i/v0/e/`, `/ingest/flags/`, etc.) now bypasses middleware entirely
- ~80% of middleware invocations were `/ingest/*` requests that called `getTenants()` only to pass through with no action

## How to test
- Deploy to preview and confirm PostHog analytics still works (events, session recording, feature flags)
- Run `vercel logs <deployment-url>` and confirm `/ingest/*` paths no longer appear
- Monitor Vercel usage dashboard for reduction in edge requests and function invocations

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -8,13 +8,14 @@ export const config = {
     /*
      * Match all paths except for:
      * 1. /api routes
-     * 2. /_next (Next.js internals)
-     * 3. /_static (inside /public)
-     * 4. all root files inside /public (e.g. /favicon.ico)
-     * 5. /media, /thumbnail, /assets (inside /public)
-     * 6. sitemap.xml, robots.txt, pages-sitemap.xml, posts-sitemap.xml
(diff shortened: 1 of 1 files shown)


yuliadub (approved):
so I dont have all the context on all the pieces at play, but after doing some reading this makes sense to me!

PR #951, opened 2026-02-18
Improve production release - manual deploy & display tag

rchlfryn:
## Description

Update the production GHA workflow to trigger on version tag pushes (`v*`) instead of branch pushes, so workflow runs display as "Deploy v1.8.0" instead of "Merge pull request #943...".

## Related Issues

Just from chatting with @busbyk 

## Key Changes

- **Tag-based trigger**: Changed `on: push: branches: ['release']` to `on: push: tags: ['v*']` — deploys now trigger from the tag push in the existing release process
- **Run name**: Added `run-name: 'Deploy ${{ github.ref_name }}'` for clear version labeling in the Actions UI
- **Release branch verification**: Added a check that fails the workflow if the tag commit doesn't match the `release` branch, catching cases where someone pushes a tag without pushing the branch
- **Manual deploy fix**: `github.event.pusher.name` is empty on `workflow_dispatch` — added `|| github.actor` fallback to prevent Vercel meta validation error
- **Cleanup step**: Hardcoded `branch="release"` since the deploy meta already hardcodes `githubCommitRef=release` and deriving it from `github.event.ref` breaks with tag triggers

## How to test

1. Create a tag and push both tag and branch: `git tag -s v1.8.1-test -m 'v1.8.1-test' && git push origin --tags && git push origin`
2. Verify the production workflow triggers and the run shows "Deploy v1.8.1-test"
3. Verify the "Verify release branch is up to date" step passes
4. Trigger a manual `workflow_dispatch` and verify it no longer fails on the author meta

### DIFF
--- a/.github/workflows/production.yaml
+++ b/.github/workflows/production.yaml
@@ -1,11 +1,12 @@
 name: production
+run-name: 'Deploy ${{ github.ref_name }}'
 env:
   VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
   VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
 on:
   push:
-    branches:
-      - 'release'
+    tags:
+      - 'v*'
   # Allow manual triggering
   workflow_dispatch:
     inputs:
@@ -94,8 +95,7 @@ jobs:
           set -o nounset
           set -o pipefail
 
-          ref="${{ github.event.ref }}"
-          branch="${ref##refs/heads/}"
+          branch="release"
 
           echo "Cleaning up deployments and aliases for branch: $branch"
 
--- a/.github/workflows/production.yaml
+++ b/.github/workflows/production.yaml
@@ -78,8 +78,9 @@ jobs:
           set -o pipefail
 
           fullName="${{ github.repository }}"
+          author="${{ github.event.pusher.name || github.actor }}"
           vercel deploy --yes --prebuilt --target=production \
-            --meta='githubCommitAuthorName=${{ github.event.pusher.name }}' \
+            --meta="githubCommitAuthorName=${author}" \
             --meta='githubCommitOrg=${{ github.repository_owner }}' \
             --meta='githubCommitRef=release' \
             --meta="githubCommitRepo=${fullName##*/}" \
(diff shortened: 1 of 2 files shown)


.github/workflows/production.yaml
```diff
@@ -1,11 +1,12 @@
...
+run-name: 'Deploy ${{ github.ref_name }}'
 env:
   VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
   VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
 on:
   push:
-    branches:
-      - 'release'
+    tags:
+      - 'v*'
```
yuliadub, line 9: would it make sense to make it something avyweb-v?

.github/workflows/production.yaml
```diff
@@ -77,25 +87,25 @@ jobs:
...
             --token=${{ secrets.VERCEL_TOKEN }}
       - name: Remove old Vercel deployments (keep latest 2)
         run: |
           set -o errexit
           set -o nounset
           set -o pipefail
 
-          ref="${{ github.event.ref }}"
-          branch="${ref##refs/heads/}"
+          branch="release"
```
yuliadub, line 108: we sure we want to hardcode this? not sure how often of if ever we would want to release out of a different branch, but just checking 🙂

yuliadub: seeing the comment in the PR description, ignore me!

.github/workflows/production.yaml
rchlfryn: Tags are already scoped to the repo, so there's no risk of collision with other projects. v* is the standard semver convention (e.g., used by GitHub's own "Generate release notes" feature and most actions like actions/checkout@v4).

CLAUDE.md
```diff
@@ -390,6 +390,10 @@ When triggered via GitHub Issues (with `@claude` mention):
 3. Follow existing code patterns and styles in the codebase
 4. Run `pnpm prettify`, `pnpm tsc`, and `pnpm lint` before committing
 
+### Pull Request Descriptions
+
+When asked to write a PR description, follow the template in `.github/PULL_REQUEST_TEMPLATE.md` and return the description in markdown.
+
```
busbyk, line 396: 🎉

busbyk (approved):
Love this. 

I see you updated the Notion Dev docs instructions too 🚀🌕

PR #961, opened 2026-03-01
Add ImageText block to post

rchlfryn:
## Description

Adds the ImageText block to posts (via the Lexical rich text editor) and introduces a new "text wrap" layout option that flows text around the image using CSS floats, in addition to the existing grid-based side-by-side layout.

## Related Issues
Noticed there was no option to add to Sierra's daily flow

## Key Changes

- Added ImageText block to the Posts collection and Lexical rich text editor
- New textWrap checkbox field — when enabled, text wraps around a floated image instead of the grid column layout
- Replaced manual background color handling with the shared `BackgroundColorWrapper` component

## How to test

1. In a Post's Lexical editor, add an ImageText block and verify it renders correctly
2. Enable "Wrap text around image" and confirm text flows around the image
3. Toggle image layout between left/right and verify float direction changes
4. Verify existing ImageText blocks on Pages still render correctly

## Migration Explanation
Adds the textWrap checkbox field (default false) to the imageText block schema. Non-destructive — existing blocks are unaffected and will default to the original grid layout.

- Consider making the image size configurable


busbyk (commented):
> Noticed there was no option to add to Sierra's daily flow

Can you explain this use case a little more? 

If the goal is just to have the ability for content editors to include images where the text wraps around them, I wonder if it makes more sense to allow this as an option when embedding MediaBlocks in the ContentBlock/lexical editors rather than expecting them to know they need to use the ImageText component. 

The code looks good - just a UX question.

rchlfryn:
> Can you explain this use case a little more?
> 
> If the goal is just to have the ability for content editors to include images where the text wraps around them, I wonder if it makes more sense to allow this as an option when embedding MediaBlocks in the ContentBlock/lexical editors rather than expecting them to know they need to use the ImageText component.

I was trying to give Sierra the ability to match their old Daily flow posts
Screenshot 2026-03-03 at 09 59 20


We briefly talked about this in stand, but just to recap.. it would be nice to see what users would want as a UX experience. We don't know what makes sense to them. If we do end up adding this functionality to the content or media block, I think it will make our code more complicated, so we want to be sure that is the correct UX for users.

rchlfryn:
@busbyk this is ready for re-review. As we decided, we will add this for now and rethink/add to inline blocks with other PRs

busbyk (approved):

PR #967, opened 2026-03-03
Fix nav colors to use branding colors

rchlfryn:
## Description

Fix navigation link colors to use proper branding/theme colors instead of generic utility colors.

## Key Changes

- **Desktop nav underline hover effect**: Changed the underline color from `bg-secondary` to `bg-callout` so it uses the the same brand color as the donate button
- **External link icon**: Changed the icon color from `text-muted` to `text-header-foreground` so it matches the nav link text instead of appearing faded

## How to test

1. Visit any tenant site (e.g., `nwac.localhost:3000`)
2. Hover over top-level nav links — the underline should match the tenant's branding color
3. Check any nav links that open in a new tab — the external link icon should match the link text color, not appear muted

## Screenshots / Demo video
Before
Screenshot 2026-03-03 at 12 38 33

After
Screenshot 2026-03-03 at 12 45 11

### DIFF
--- a/src/components/Header/DesktopNav.client.tsx
+++ b/src/components/Header/DesktopNav.client.tsx
@@ -16,7 +16,7 @@ import { RenderNavLink } from './RenderNavLink'
 import { TopLevelNavItem } from './utils'
 
 const underlineHoverClassName =
-  "relative w-fit after:content-[''] after:absolute after:left-2 after:bottom-0 after:h-[1px] after:w-0 after:bg-secondary after:transition-all after:duration-300 hover:after:w-[calc(100%-1rem)] hover:text-header-foreground-highlight"
+  "relative w-fit after:content-[''] after:absolute after:left-2 after:bottom-0 after:h-[1px] after:w-0 after:bg-callout after:transition-all after:duration-300 hover:after:w-[calc(100%-1rem)] hover:text-header-foreground-highlight"
 
 export const DesktopNav = ({
   topLevelNavItems,
--- a/src/components/Header/DesktopNav.client.tsx
+++ b/src/components/Header/DesktopNav.client.tsx
@@ -16,7 +16,7 @@ import { RenderNavLink } from './RenderNavLink'
 import { TopLevelNavItem } from './utils'
 
 const underlineHoverClassName =
-  "relative w-fit after:content-[''] after:absolute after:left-2 after:bottom-0 after:h-[1px] after:w-0 after:bg-callout after:transition-all after:duration-300 hover:after:w-[calc(100%-1rem)] hover:text-header-foreground-highlight"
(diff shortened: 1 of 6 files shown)


src/components/Header/RenderNavLink.tsx
```diff
@@ -37,7 +37,7 @@ export const RenderNavLink = ({ link, className = '', onClick, children }: Rende
       >
         {children || link.label}
         {link.newTab && (
-          <ExternalLink className="w-4 h-4 flex-shrink-0 ml-2 -mt-1.5 lg:-mt-0.5 text-muted" />
+          <ExternalLink className="w-4 h-4 flex-shrink-0 ml-2 -mt-1.5 lg:-mt-0.5 text-callout" />
```
busbyk, line 40: 👏

src/components/Header/DesktopNav.client.tsx
```diff
@@ -16,7 +16,7 @@ import { RenderNavLink } from './RenderNavLink'
 import { TopLevelNavItem } from './utils'
 
 const underlineHoverClassName =
-  "relative w-fit after:content-[''] after:absolute after:left-2 after:bottom-0 after:h-[1px] after:w-0 after:bg-secondary after:transition-all after:duration-300 hover:after:w-[calc(100%-1rem)] hover:text-header-foreground-highlight"
+  "relative w-fit after:content-[''] after:absolute after:left-2 after:bottom-0 after:h-[1px] after:w-0 after:bg-callout after:transition-all after:duration-300 hover:after:w-[calc(100%-1rem)] hover:text-header-foreground-highlight"
```
busbyk: Hm I think this was based on the original designs - intentionally using the secondary color for hover states. 

But you're right, it does look weird to have a different hovered underline color on these nav items compared to the donate button when they're such similar colors like with Sierra and Sawtooth. 

But for NWAC the callout color is very different than their secondary color:
Image

Pretty different to what's shown for hover underline color now (not that they're in production but the theme was chosen to be this way): 
Image

I think I'm for the change but, at least for NWAC, this is different than their chosen theme colors originally.

busbyk: The MobileNav component also sets an underline. Even though this only shows up when you've constrained the width of a web browser on desktop, we might as well update there too, to be consistent.

rchlfryn: I think since NWAC isnt live yet, we can cross that bridge when they launch. Thoughts?

busbyk: @rustynwac any thoughts on this color change? I.e. using a theme's `callout` color instead of the theme's `secondary` color for the hover underline color. Makes sense to both @rchlfryn and I - at least for Sierra and Sawtooth. See above for more context.

rchlfryn: As we talked about in slack, I added a new variable and added it to the onboarding docs. I still think we need to revisit how we help centers build their "brand"

busbyk (approved):
Great!

Issue #757, opened 2025-12-04
Media block should use responsive image sizes based on it's container


busbyk:
Media blocks in richText fields use viewport width based responsive sizing which doesn't work when embedded in a smaller container. Plus we should have an additional, smaller size. 

https://www.loom.com/share/d49426c1009d44879abce2f3f55c447b

Applies to inline media blocks as well.

rchlfryn:
@busbyk can you confirm this done in https://github.com/NWACus/web/pull/867?

busbyk:
> [@busbyk](https://github.com/busbyk) can you confirm this done in [#867](https://github.com/NWACus/web/pull/867)?

No, it's not. The MediaBlock still uses viewport based max width classNames to handle it's responsive sizing. We need to update those to be based on it's parent element instead of the viewport because they don't work as expected when the MediaBlock is embedded in a smaller container - the posts layout for example.

rchlfryn:
@claude  implement this

PR #968, opened 2026-03-03
Inline media block

busbyk:
## Description                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                               
Adds a new `InlineMedia` inline block for the Lexical rich text editor, allowing editors to embed images directly within text flow. Images can be positioned inline with configurable vertical alignment, or floated left/right with text wrapping.
                                                                                                                                                                                                                                                                               
## Related Issues

None, alternative to #961 

## Key Changes

- **New `InlineMedia` block** (`src/blocks/InlineMedia/`) — config and component for inline media in rich text
- **Position options**: Inline (within text flow) or float left/right (text wraps around image)
- **Size options**: Original (natural size), percentage widths (25%/50%/75%/100%), or fixed pixel height
- **Vertical alignment**: Top, middle, bottom, or baseline (only for inline position)
- **Optional caption**: Displayed below the image as small gray text
- **Registered in Posts collection** and **default Lexical editor config** as an `inlineBlock`
- **RichText renderer** updated with JSX converter for frontend rendering

## How to test

1. Run `pnpm dev` and navigate to the admin panel
2. Create or edit a Post (or any content using the default Lexical editor)
3. In the rich text editor, use the inline block inserter to add an "Inline Media" block
4. Upload or select a media item and configure position, size, and alignment
5. Verify:
   - **Inline position**: Image renders within the text line with correct vertical alignment
   - **Float left/right**: Image floats to the side with text wrapping around it
   - **Size options**: Percentage widths scale relative to the container; fixed height respects the pixel value
   - **Caption**: Appears as small text below the image when provided
6. View the published content on the frontend to confirm the rich text JSX converter renders correctly

## Screenshots / Demo video

https://www.loom.com/share/a167f8e500824d60b88914fe6b66014b

Hmm I wonder if floats are too confusing to content editors when they can use them, unconstrained in the ContentBlock?
<img width="1394" height="1128" alt="image" src="https://github.com/user-attachments/assets/c6b5096a-e2c1-419a-a30a-1e1906cd7287" />


## Migration Explanation

No migration needed — this adds a new inline block type to existing Lexical rich text fields. Existing content is unaffected.

- Could extend to support inline video embeds
- May want to add margin/padding controls for finer spacing adjustments around floated images
- Thumbnails for inline blocks don't seem to work. We likely need to open a PR in Payload for this. I've left `imageUrl` in the block config using the same MediaThumbnail.jpg for when that gets fixed.
- Images are still not exactly responsive - if we set a percentage width that width gets used at all screen sizes. Same problem as https://github.com/NWACus/web/issues/757. Will fix this block in that issue as well.


src/blocks/InlineMedia/config.ts
```diff
@@ -0,0 +1,77 @@
...
+      relationTo: 'media',
+      required: true,
+    },
+    {
+      name: 'position',
+      type: 'select',
+      defaultValue: 'inline',
+      options: [
+        { label: 'Inline', value: 'inline' },
+        { label: 'Float left', value: 'float-left' },
```
rchlfryn: Interesting - for some reason I thought we were using camel case for values, but every other value is one word.

src/blocks/InlineMedia/Component.tsx
```diff
@@ -0,0 +1,89 @@
...
+    : cn('inline-block', verticalAlignClasses[verticalAlign ?? 'middle'])
+
+  // For fixed height, wrap Media in a span with explicit height.
+  // Descendant selectors propagate height through Media's intermediate span and picture elements.
+  const mediaElement = isFixedHeight ? (
+    <span
+      className="block [&>span]:h-full [&_picture]:h-full [&_picture]:my-0"
+      style={{ height: `${fixedHeight}px` }}
+    >
+      <Media htmlElement="span" resource={media} imgClassName={imgSizeClass} sizes={sizes} />
```
rchlfryn: Any reason to not include `pictureClassName="my-0"` here?

src/blocks/InlineMedia/config.ts
```diff
@@ -0,0 +1,77 @@
...
+    {
+      name: 'size',
+      type: 'select',
+      defaultValue: 'original',
+      options: [
+        { label: 'Original (natural size)', value: 'original' },
+        { label: '25% width', value: '25' },
+        { label: '50% width', value: '50' },
+        { label: '75% width', value: '75' },
+        { label: '100% width', value: '100' },
```
rchlfryn, line 52: I don't think it matters, but maybe we add a `w` in front of of each value to ensure it's a string. Doesn't seem to be an issue.

src/blocks/InlineMedia/config.ts
rchlfryn: Maybe update the labels to just be `Left` and `Right`? I am not sure float is needed here.

src/blocks/InlineMedia/Component.tsx
```diff
@@ -0,0 +1,89 @@
...
+  if (resolvedSize === 'original') {
+    sizeClass = 'max-w-fit'
+  } else if (isWidthSize(resolvedSize)) {
+    sizeClass = widthClasses[resolvedSize]
+    imgSizeClass = 'w-full h-auto'
+    // Approximate sizes hint for responsive images
+    sizes = `${resolvedSize}vw`
+  } else if (isFixedHeight) {
+    imgSizeClass = 'h-full w-auto'
+    sizes = '96px'
```
rchlfryn, line 57: How is this being calculated/ is it needed?

src/blocks/InlineMedia/Component.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const positionClasses = isFloat
+    ? cn(position === 'float-left' ? 'float-left mr-2' : 'float-right ml-2', 'mb-1')
+    : cn('inline-block', verticalAlignClasses[verticalAlign ?? 'middle'])
+
+  // For fixed height, wrap Media in a span with explicit height.
+  // Descendant selectors propagate height through Media's intermediate span and picture elements.
+  const mediaElement = isFixedHeight ? (
+    <span
+      className="block [&>span]:h-full [&_picture]:h-full [&_picture]:my-0"
+      style={{ height: `${fixedHeight}px` }}
```
rchlfryn, line 69: ⛏️ You could make this a class name like you are above and remove the inline style.

rchlfryn (changes requested):
A lot of thinking out loud as review notes in the PR. Take em or leave em.

To answer your question about having both this and the ImageBlock, I think we should have both and get feedback from users on which makes more sense. I could see users wanting more inline blocks.

src/blocks/InlineMedia/Component.tsx
busbyk: This is the  attribute — a hint that tells the browser which source to pick from a responsive srcset before layout is known. For fixed-height images the rendered width depends on the image's aspect ratio, which we don't have at render time without reading the media dimensions. 96px is a conservative fallback so the browser doesn't load the full-resolution source. It's imperfect but errs on the side of a smaller download; the browser will still render at the correct size.

src/blocks/InlineMedia/Component.tsx
busbyk: `fixedHeight` is dynamic so we can't use a classname here. See https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names:~:text=Don%27t%20use%20props%20to%20build%20class%20names%20dynamically

src/blocks/InlineMedia/config.ts
busbyk: I like the idea, to make it clear that these are strings but we actually use these for the sizes attribute in [the component](https://github.com/NWACus/web/blob/inline-media-block/src/blocks/InlineMedia/Component.tsx#L54) so we'd have to strip off the 'w' prefix which I think would be more confusing.

src/blocks/InlineMedia/config.ts
busbyk: Cool, I'm down for the label change - updated the description too: [61ba101](https://github.com/NWACus/web/pull/968/commits/61ba10180d31715e51ba0ffbe6ed6ffa17d65fc6)

> Interesting - for some reason I thought we were using camel case for values, but every other value is one word.

I almost always default to using snake case for unique values like keys like this or slugs. Just my preference 🤷‍♂️

src/blocks/InlineMedia/Component.tsx
busbyk: Ah no, good call: [4b352f3](https://github.com/NWACus/web/pull/968/commits/4b352f390ed8b90914a935211f9e61bf592b7bc2)

src/blocks/InlineMedia/Component.tsx
rchlfryn: My mistake. I was thinking of the pattern in `BackgroundColorWrapper` where we do `const bgColorClass = bg-${backgroundColor}` and use that in `className`. But you're right — according to the Tailwind docs, dynamically constructed class names like that shouldn't be written like this. Thanks for the link.

Worth noting we may want to revisit `BackgroundColorWrapper` as well, and stop the anti-pattern

busbyk:
@rchlfryn are you waiting on anything from me for this re-review?

rchlfryn (approved):
Looks good

PR #969, opened 2026-03-04
Add devMode for NAC widgets

### LINKED ISSUE #717: NAC Widgets devMode
Add an environment variable that controls the value passed to the NAC widget config's `devMode` attribute. 

Currently, we always set this to false. Could add an environment variable like `NAC_WIDGET_DEV_MODE` or similar that would set `devMode` to true or false in the widget configs. 

Another option would be to add a field to the `nacWidgetsConfig` global that controls the use of `devMode` which would be nice because we wouldn't need to redeploy. I would just want to verify that the widget configs aren't cached too aggressively. The revalidation hooks on `nacWidgetsConfig` should be sufficient though. 

Background: 

Setting `devMode` to true in the widget configs changes the baseUrl used by the widgets, not the actual widgets themselves. If we want to use different widgets we can change the `nacWidgetsConfig` global values in the admin panel without a redeploy.

Having a dedicated environment pointing to staging would be helpful for `nwacus/app` (AvyApp) development when testing custom data we've entered and to cross reference functionality of the existing widgets compared to views we're recreating in the future. So adding a dedicated environment

rchlfryn:
## Description

Add a `devMode` setting to the NAC Widgets Config admin panel and pass it through to all NAC widget components.

Previously, `devMode` was always hardcoded to `false`. This makes it impossible for staging/preview deployments to point widgets at the NAC staging base URL — needed for AvyApp development and testing custom data entered in staging. In production, `devMode` is always forced to `false` regardless of the admin panel setting.

## Related Issues

Fixes #717

## Key Changes

- Add `devMode` checkbox to NAC Widgets Config global
- Respect the admin panel `devMode` setting in non-production environments, force `false` in production
- Pass `devMode` through to all NAC widget components (was hardcoded `false`)
- Add tests for `getNACWidgetsConfig` and `getAvalancheCenterPlatforms` using MSW to intercept HTTP requests
- Add MSW as a dev dependency for realistic API mocking in tests
- Migration to add `devMode` column

## How to test

1. **Local/staging**
    2.  with admin panel devMode off→ widgets use production URL (`devMode: false`)
    3.  with admin panel devMode on → widgets use staging URL (`devMode: true`)
4. **Production** → widgets always use production URL (`devMode: false`) regardless of admin setting
5. Check `/api/[center]/nac-config` returns the correct `devMode` value
6. Run `pnpm test` — all tests pass

## Migration Explanation

Adds a `devMode` boolean column (default `false`) to the NAC widgets config table. Non-destructive — existing rows get the default value.


rchlfryn:
@busbyk ~Leaving this as a draft.~ It is ready fro review but I am curious if you think this is the correct approach based on the issue you wrote.

rchlfryn:
@busbyk Ready for review

__tests__/server/getAvalancheCenterPlatforms.server.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// Mock the entire nac module, re-implementing getAvalancheCenterPlatforms
+// with the real logic but using mocked data instead of fetching from the API
+jest.mock('../../src/services/nac/nac', () => ({
+  getAvalancheCenterPlatforms: jest.fn(async (centerSlug: string) => {
+    const centerSlugToUse = centerSlug === 'dvac' ? 'nwac' : centerSlug
+
+    return mockCenters[centerSlugToUse] ?? allFalsePlatforms
+  }),
+}))
```
busbyk: Replacing the function we're testing here with a mocked function that's only used in this test doesn't actually test the getAvalancheCenterPlatforms function. 

Ideally, we would intercept the API requests and return mocked data using [MSW](https://mswjs.io/) or something similar so that we would be testing the actual implementation of `getAvalancheCenterPlatforms`. 

If you don't want to set up MSW, I suggest removing this test file completely.

__tests__/server/getNACWidgetsConfig.server.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    expect(result.baseUrl).toBe('https://du6amfiq9m9h7.cloudfront.net/public/v2')
+  })
+
+  it('forces devMode to false in production even when global config has it enabled', async () => {
+    Object.defineProperty(process.env, 'NODE_ENV', { value: 'production', configurable: true })
+    mockConfig.devMode = true
+
+    const result = await getNACWidgetsConfig()
+    expect(result.devMode).toBe(false)
+  })
```
busbyk, line 51: The PR description says that `nacWidgetsConfig.devMode` should be respected when `NAC_WIDGET_DEV_MODE` is not set.

src/utilities/getNACWidgetsConfig.ts
```diff
@@ -16,5 +16,9 @@ export async function getNACWidgetsConfig() {
     'Could not determine NAC widgets base url. Ensure this is set in the NAC Widgets Config global.',
   )
 
-  return { version, baseUrl }
+  // In production, always force devMode off regardless of the admin setting
+  const devMode =
+    process.env.NODE_ENV === 'production' ? false : (nacWidgetsConfig?.devMode ?? false)
```
busbyk: Yea I don't see why we wouldn't respect this in production.

rchlfryn: The intention was we would never want this to be accidentally `true` in production, but I guess we might want that? I can't think of a use case.

rchlfryn: from slack convo: we will check our environment friendly name via https://github.com/NWACus/web/blob/main/src/utilities/getEnvironmentFriendlyName.ts and disable in prod but allow in other environments like dev and preview.

__tests__/server/getNACWidgetsConfig.server.test.ts
rchlfryn: Ah good shout - this was implemented incorrectly and overengineered. I somehow got confused. The priority is:
- if prod -> always `false`
- `nacWidgetsConfig.devMode` from admin panel (if env var not set)
- `false` (if neither is set)

__tests__/server/getAvalancheCenterPlatforms.server.test.ts
rchlfryn: Good callout - set up MSW you linked and implemented this in [251372b](https://github.com/NWACus/web/pull/969/commits/251372b68477177bea15a0b145894aa551a31d22)

busbyk: Did you forget to push? I'm not seeing this update in that commit or in this PR?

rchlfryn: This was so long ago I have no idea what I meant but this was not done in that commit. I remember coding this but don't know what happened. Done in [9ff7a91](https://github.com/NWACus/web/pull/969/commits/9ff7a91179a1f28908b85c5c49db31eac8ee55d5)

rchlfryn:
@busbyk this is ready for re-review

busbyk (approved):
Re-reviewing this I had the thought: I wonder if we could use a provider and hook to provide the NAC Widgets Config to the NACWidget client component instead of all of this prop passing. This PR had to touch so many files just to add an option to that component.

rchlfryn:
@busbyk Made this update. Good suggestion

PR #972, opened 2026-03-04
Remove custom domain from tenants collection

busbyk:
## Description

Removes the `customDomain` field from the `tenants` collection in the database, consolidating custom domain configuration into the hardcoded `AVALANCHE_CENTERS` constant in `src/utilities/tenancy/avalancheCenters.ts`. This eliminates a dual-source-of-truth problem where custom domains existed in both the database and the hardcoded constant, which could lead to mismatches.

## Related Issues

Follow-up to #947.

## Key Changes

- **Removed `customDomain` field from the Tenants collection schema** (`src/collections/Tenants/index.ts`) and generated types
- **Added `getEmailDomain()` utility** to `avalancheCenters.ts` — derives an email-safe domain from the custom domain by stripping `www.` prefix and port numbers
- **Refactored `isTenantDomainScoped()`** — now uses `findCenterByDomain()` from `AVALANCHE_CENTERS` instead of querying the database for custom domains, with a subdomain fallback
- **Updated domain-scoped access control** — `validateDomainAccessBeforeLogin` and `byGlobalRoleOrTenantRoleAssignmentOrDomain` now use the new `getEmailDomain()` utility
- **Simplified seed and bootstrap scripts** — removed all `customDomain` references from seed data
- **Migration** — `20260303_233752_remove_custom_domain_from_tenants.ts` drops the `custom_domain` column from the `tenants` table
- **Fixed e2e tests** — updated tenant cookie assertions to use slugs instead of numeric IDs (the `payload-tenant` cookie now stores slugs)

## How to test

1. Run `pnpm seed:standalone` — verify it completes without errors
2. Run `pnpm dev` and visit `nwac.localhost:3000` — verify the site resolves correctly via subdomain
3. Visit `localhost:3000/admin` — verify the admin panel loads and tenant selector works
4. Run `pnpm tsc` and `pnpm lint` — verify no type or lint errors
5. Run `pnpm test` — verify all tests pass

## Screenshots / Demo video

N/A — no visual changes

## Migration Explanation

**Migration**: `20260303_233752_remove_custom_domain_from_tenants`

- **Up**: Drops the `custom_domain` column from the `tenants` table (`ALTER TABLE tenants DROP COLUMN custom_domain`)
- **Down**: Re-adds the column as nullable text (`ALTER TABLE tenants ADD custom_domain text`)
- **Data loss**: The column values are not preserved on rollback — they would need to be re-seeded. This is acceptable because the custom domain data now lives in the `AVALANCHE_CENTERS` constant.

### DIFF
--- a/__tests__/server/findCenterByDomain.server.test.ts
+++ b/__tests__/server/findCenterByDomain.server.test.ts
@@ -1,0 +1,50 @@
+import { findCenterByDomain } from '../../src/utilities/tenancy/avalancheCenters'
+
+describe('findCenterByDomain', () => {
+  it('finds a center by exact domain match (no www)', () => {
+    expect(findCenterByDomain('nwac.us')).toBe('nwac')
+    expect(findCenterByDomain('alaskasnow.org')).toBe('aaic')
+    expect(findCenterByDomain('avalanche.state.co.us')).toBe('caic')
+  })
+
+  it('finds a center when domain has www. prefix', () => {
+    expect(findCenterByDomain('www.nwac.us')).toBe('nwac')
+    expect(findCenterByDomain('www.cnfaic.org')).toBe('cnfaic')
+    expect(findCenterByDomain('www.sierraavalanchecenter.org')).toBe('sac')
+  })
+
+  it('finds a center when stored domain has www. but input does not', () => {
+    expect(findCenterByDomain('cnfaic.org')).toBe('cnfaic')
+    expect(findCenterByDomain('sierraavalanchecenter.org')).toBe('sac')
+  })
+
+  it('is case insensitive', () => {
+    expect(findCenterByDomain('NWAC.US')).toBe('nwac')
+    expect(findCenterByDomain('NwAc.Us')).toBe('nwac')
+    expect(findCenterByDomain('WWW.CNFAIC.ORG')).toBe('cnfaic')
+    expect(findCenterByDomain('www.SierrAAvAlAnchECentEr.Org')).toBe('sac')
+  })
+
+  it('handles domains shared by multiple centers (returns first match)', () => {
+    // Multiple centers use 'alaskasnow.org': aaic, cac, earac, hac, vac
+    // Should return the first one found (aaic)
+    const result = findCenterByDomain('alaskasnow.org')
+    expect(result).toBe('aaic')
+  })
+
+  it('returns undefined for non-existent domains', () => {
+    expect(findCenterByDomain('example.com')).toBeUndefined()
+    expect(findCenterByDomain('notarealcenter.org')).toBeUndefined()
+    expect(findCenterByDomain('www.doesnotexist.com')).toBeUndefined()
+  })
+
+  it('returns undefined for empty string', () => {
+    expect(findCenterByDomain('')).toBeUndefined()
+  })
+
+  it('handles domains with trailing/leading whitespace', () => {
+    expect(findCenterByDomain(' nwac.us')).toBe('nwac')
+    expect(findCenterByDomain('nwac.us ')).toBe('nwac')
+    expect(findCenterByDomain('  nwac.us  ')).toBe('nwac')
+  })
+})
--- a/__tests__/server/getEmailDomain.server.test.ts
+++ b/__tests__/server/getEmailDomain.server.test.ts
@@ -1,0 +1,15 @@
+import { getEmailDomain } from '../../src/utilities/tenancy/avalancheCenters'
+
+describe('getEmailDomain', () => {
+  it('returns the domain as-is when there is no www. prefix', () => {
+    expect(getEmailDomain('nwac')).toBe('nwac.us')
+  })
+
+  it('strips only the www. prefix, not other subdomains', () => {
+    // cnfaic has customDomain 'www.cnfaic.org' — www. should be stripped
+    expect(getEmailDomain('cnfaic')).toBe('cnfaic.org')
+
+    // aaic has customDomain 'alaskasnow.org' — no www., returned as-is
+    expect(getEmailDomain('aaic')).toBe('alaskasnow.org')
+  })
+})
--- a/__tests__/server/getEmailDomain.server.test.ts
+++ b/__tests__/server/getEmailDomain.server.test.ts
@@ -6,10 +6,8 @@ describe('getEmailDomain', () => {
   })
 
   it('strips only the www. prefix, not other subdomains', () => {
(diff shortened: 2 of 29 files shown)


rchlfryn (changes requested):
@busbyk it looks like this contains changes from other PRs. do you mind taking a look and cleaning it up?

busbyk:
> @busbyk it looks like this contains changes from other PRs. do you mind taking a look and cleaning it up?

Fixed!

src/endpoints/seed/index.ts
```diff
@@ -540,26 +537,27 @@ export const seed = async ({
         password: password,
       },
       ...Object.values(tenants)
-        .map((tenant): RequiredDataFromCollectionSlug<'users'>[] => [
-          {
-            name: tenant.slug.toUpperCase() + ' Admin',
-            // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
rchlfryn, line 546: don't forget to update and commit `consistent-type-assertions.txt`

src/utilities/generatePreviewPath.ts
```diff
@@ -28,8 +29,8 @@ export const generatePreviewPath = async ({ collection, slug, tenant, req }: Pro
   }
   // Check if current host is the tenant's domain
   const currentHost = req.headers.get('host') || req.host
-  const isTenantCustomDomain =
-    resolvedTenant?.customDomain && currentHost === resolvedTenant.customDomain
+  const matchedSlug = findCenterByDomain(currentHost)
+  const isTenantCustomDomain = !!matchedSlug && matchedSlug === resolvedTenant?.slug
```
rchlfryn, line 33: `isTenantCustomDomain` is no longer accurate. Maybe update to `matchedTenantSlug` or something else

__tests__/server/getEmailDomain.server.test.ts
```diff
@@ -0,0 +1,15 @@
+import { getEmailDomain } from '../../src/utilities/tenancy/avalancheCenters'
+
+describe('getEmailDomain', () => {
+  it('returns the domain as-is when there is no www. prefix', () => {
+    expect(getEmailDomain('nwac')).toBe('nwac.us')
+  })
+
+  it('strips only the www. prefix, not other subdomains', () => {
+    // cnfaic has customDomain 'www.cnfaic.org' — www. should be stripped
```
rchlfryn: ⛏️  do we need these comments?

src/utilities/tenancy/avalancheCenters.ts
```diff
@@ -65,6 +65,17 @@ export function isValidTenantSlug(slug: string): slug is ValidTenantSlug {
...
 
+/**
+ * Lookup a center by its custom domain
+ */
+/** Derive an email-safe domain from the custom domain by stripping www. prefix and port */
+export function getEmailDomain(slug: ValidTenantSlug): string {
+  return AVALANCHE_CENTERS[slug].customDomain
+    .toLowerCase()
+    .replace(/^www\./, '') // strip www. prefix
+    .replace(/:\d+$/, '') // strip port number (e.g., :3000 for local testing)
```
rchlfryn, line 76: do we need this? I don't think this will ever be needed and we do not have a test for this

rchlfryn (changes requested):
Overall looks good - couple of comments

src/endpoints/seed/index.ts
busbyk: Good call! I'm used to this failing in CI to alert me 😅 but since this PR isn't against main, CI didn't run. Thanks for the reminder 🙏

src/utilities/tenancy/avalancheCenters.ts
busbyk: When we test custom domains locally we need to include the port. The likelihood that we're testing custom domains and testing the ability to log into the admin panel on that custom domain is pretty low of course. But we would need this for that scenario. 

In the future enhancements section of https://github.com/NWACus/web/pull/947 I suggest that we stop allowing users to log into the admin panel on custom domains/subdomains at all, fyi. But I still think we should keep this port stripping since it's possible that we have a port specifified in a custom domain in this file. 

I can add a test?

__tests__/server/getEmailDomain.server.test.ts
busbyk: Removed ✅

src/utilities/tenancy/avalancheCenters.ts
busbyk: I added a test :)

src/utilities/generatePreviewPath.ts
busbyk: Hmm, I think `isTenantCustomDomain` is still accurate. Matched slug will only be truthy if `findCenterByDomain` finds a matching custom domain (i.e. the same domain as the current host). The extra check `matchedSlug === resolvedTenant?.slug` is just to be thorough. 

Although this is another check that could be removed if we decide we want to disallow logging into the admin panel on tenant domains (custom and subdomains) and only allow on the `NEXT_PUBLIC_ROOT_DOMAIN`

__tests__/server/findCenterByDomain.server.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(findCenterByDomain('WWW.CNFAIC.ORG')).toBe('cnfaic')
+    expect(findCenterByDomain('www.SierrAAvAlAnchECentEr.Org')).toBe('sac')
+  })
+
+  it('handles domains shared by multiple centers (returns first match)', () => {
+    // Multiple centers use 'alaskasnow.org': aaic, cac, earac, hac, vac
+    // Should return the first one found (aaic)
+    const result = findCenterByDomain('alaskasnow.org')
+    expect(result).toBe('aaic')
+  })
```
busbyk, line 33: So I just realized how problematic this is. I don't think we should hold up this PR or #947 based on this because this is the same behavior we had before and we only need to solve for this when we onboard any of the alaskasnow.org centers. But I'll start a thread in Slack to discuss.

src/utilities/generatePreviewPath.ts
rchlfryn: Oh I missed `matchedSlug` was calling `findCenterByDomain`. Ignore me

__tests__/server/findCenterByDomain.server.test.ts
rchlfryn: I think this is fine to side step for now.

rchlfryn (approved):

PR #974, opened 2026-03-05
Automate onboarding for new tenant

### LINKED ISSUE #533: Automate onboarding new avy centers
### Description
The way we create built-in pages is currently not great. We rely on a super admin user manually creating them during onboarding. 
We need a create these pages when adding a new avalanche center.
> [!NOTE]
> 3rd column in table below refers to avalanche center with a `single` or `multi` zones. 
> If `both`, it should be added regardless of the number of zones.

| Title | URL | for AC w/single or multi zone*  | Notes |
|--------|--------|--------|--------|
| All Forecasts | /forecasts/avalanche| mutli | |
| _ZONE NAME_ | /forecasts/avalanche/_ZONE_ | multi |  possibly use `getActiveForecastZones` to get zones |
| Avalanche Forecast | /forecasts/avalanche/_ZONE_ | single | |
| Weather Stations | /weather/stations/map| both | |
| Recent Observations | /observations | both | |
| Submit Observations | /observations/submit | both | |
| Blog | /blog | both | |
| Events | /events | both | | 

### Possible solutions

We need to figure out when/where to run a script to ensure that all of the built-in pages have documents in the built-in pages collection.
* We vetoed the [init hook](https://payloadcms.com/docs/configuration/overview#config-options:~:text=More%20details.-,onInit,-A%20function%20that) but I think that runs every time a new serverless function is executed since Payload is re-initialized. we dont want that performance hit.
* Running something after deployment to ensure the appropriate built-in pages collection documents exist is what we're looking for so a custom endpoint that our GHA workflows hit after deployment might be a good solution.

- There should be a single source of truth for AFP/built-in pages in the navigations collection
	- Weather stations special scenario should be removed at some point
	- Forecasts and observations should display a read-only representation of how they'll render on the frontend
- We'd need to make Payload the source of truth for forecast zones and navigation items which would require keeping our built-in pages collection up to date
	- Likely use a cron job to keep these in sync. Could try Payload jobs. 

### v1
- Automate creation of website settings, navigation, built-in pages, home pages, pages (possibly posts)
- Idempotent - allow to run and fail. If not allow to rerun

### v2
- Use AFP and NAC APIs to keep navigation in sync

### v3
- Use an onboarding guide

rchlfryn:
## Description

Automates the entire tenant onboarding workflow. Creating a new tenant in the admin panel now automatically provisions all default data (settings, pages, navigation, etc.) and displays an interactive onboarding checklist on the tenant edit page to track progress.

## Related Issues

Fixes #533 

## Key Changes

- **Provision endpoint** (`POST /api/tenants/provision`) — creates a tenant and provisions all default data in one call. Can also be triggered via the `provisionAfterChange` hook on tenant creation.
- **Provisioning logic** — creates website settings with placeholder brand assets, 4 built-in pages, copies template pages from DVAC (filtering tenant-scoped blocks, converting static lists to dynamic), creates a home page with welcome content, and builds navigation linking everything together.
- **Onboarding Checklist UI** — a `ChecklistItem`-based component on the tenant edit page showing status of all automated and manual steps, with a "Rerun Provisioning" button if anything is incomplete.
- **Server actions** (`onboardingActions.ts`) — checks provisioning status by querying all relevant collections and inspecting `colors.css` and `centerColorMap` for theme setup.
- **Placeholder brand assets** — three PNG images (logo, icon, banner) used as required upload placeholders for website settings during provisioning.
- **`removeIdKey` utility** — strips `id` keys from deeply nested objects when copying template pages to avoid ID collisions.
- **Seed fix** — added `skipProvision: true` context to `upsertGlobals` to prevent auto-provisioning during seeding.
- **Updated `docs/onboarding.md`** — rewrote to reflect automated workflow, added CSS variable reference table for theme setup.

## How to test

1. Run `pnpm seed` to get a fresh database with the template tenant (DVAC)
2. Go to admin panel → Avalanche Centers → Create
3. Enter a name and slug, save
4. Verify the onboarding checklist appears showing all automated steps as complete
5. Click "Update Brand Assets" to confirm the settings link works
6. Verify "Add brand colors" and "Add OG image colors" show as incomplete (since no theme exists for the new tenant)

## Screenshots / Demo video
if nothing runs correctly
Screenshot 2026-03-05 at 17 50 40

Need to do manual actions
Screenshot 2026-03-05 at 16 38 17

Full set up
Screenshot 2026-03-05 at 16 38 54

https://www.loom.com/share/7ba0486274454cff97b273f3880c358d

## Migration Explanation

No schema changes — no migration needed.

- #976 — Tenant deletion currently leaves orphaned data. Need a `beforeDelete` cleanup hook and confirmation prompt.
- #998 Theme creation could potentially be partially automated (generate a colors.css entry from brand asset colors).
- Custom domain setup remains manual (Vercel, DNS, PostHog configuration).
- #999 Automatically check form the AFP if this is a single or multi-zone center and provision accordingly
- #1000 How do we want to define completed for each of the content steps? Right now the step will be incomplete if something is deleted. See https://github.com/NWACus/web/pull/974#discussion_r2943535563


src/collections/Tenants/hooks/provisionAfterChange.ts
busbyk, line 1: Do you remember the behavior of the seed button that came with the website template? It would use a toast with a loading state during the seeding so the UI would indicate that things were running. You called out the latency in saving a tenant in the demo. We could do something like that to address that latency where it seems like something has gone wrong in the UI. 

Here is the seed button code: https://github.com/payloadcms/payload/blob/da212fd849666f99b726d7fa6e54f3f0d733e29a/templates/website/src/components/BeforeDashboard/SeedButton/index.tsx

So instead of a `CollectionAfterChangeHook` we could use a component in `beforeDocumentControls` like you did in https://github.com/NWACus/web/pull/970/changes#diff-3511e1d9635960105835044816f4df520c312154691f9c03c42ca97dce359ad3

This would also mean we could remove the skipProvision check here and the setting in the seed script since that component would only trigger in the UI.

src/collections/Tenants/index.ts
```diff
@@ -66,5 +73,15 @@ export const Tenants: CollectionConfig = {
       },
     },
     contentHashField(),
+    {
+      type: 'ui',
+      name: 'onboardingChecklist',
+      admin: {
+        components: {
```
busbyk, line 75: Might be nice to add a Cell component here so we could see at a glance if the checklist had been completed in the list view: 

Image

docs/onboarding.md
```diff
@@ -1,84 +1,34 @@
...
+## Automated
+
+Create the tenant in the admin panel → provisioning runs automatically via `provisionAfterChange`. An **Onboarding Checklist** on the tenant edit page tracks progress. This requires super admin permissions. Creates the tenant record and runs all steps below.
+
+Provisioning is idempotent and can be rerun safely.
+
+| Step | Details |
+|------|---------|
+| Website Settings | Created with placeholder brand assets (logo, icon, banner). Replace with real assets via the checklist link. |
+| Built-in pages | Creates 4 standard pages: All Forecasts, Weather Stations, Recent Observations, Submit Observations. |
```
busbyk: Based on the table in https://github.com/NWACus/web/issues/533, this is missing a few built-in pages:
1. If a center is a single zone center they should get an "Avalanche Forecast" built-in page which points to their only zone.
2. Blog
3. Events

docs/onboarding.md
```diff
@@ -101,7 +51,7 @@ Add a new CSS class using the tenant's slug (e.g. `.dvac`) in `src/app/(frontend
 | `--brand-50` through `--brand-950` | Brand color range used for block backgrounds |
 | `--brand-foreground-50` through `--brand-foreground-950` | Text colors for each brand background |
 
-## Configuring a custom domain in production
+### Custom domain
```
busbyk: Can we change this back to `## Configuring a custom domain in production`?

src/collections/Tenants/endpoints/provisionTenant.ts
```diff
@@ -0,0 +1,563 @@
...
+    return Response.json({ error: 'Super admin access required' }, { status: 403 })
+  }
+
+  const body = await req.json?.()
+  if (!body?.name || !body?.slug) {
+    return Response.json({ error: 'name and slug are required' }, { status: 400 })
+  }
+
+  try {
+    // Create the tenant (allowDirectCreate bypasses the beforeChange guard)
```
busbyk: What is allowDirectCreate?

src/collections/Tenants/components/onboardingActions.ts
```diff
@@ -0,0 +1,158 @@
...
+            return false
+          }
+          return true
+        })
+        .map((p) => ({ slug: p.slug, title: p.title }))
+    }
+
+    const tenantPagesBySlug = new Map(pages.docs.map((p) => [p.slug, p]))
+    const copiedPages = templatePageSlugs.filter((p) => tenantPagesBySlug.has(p.slug))
+    const missing = templatePageSlugs.filter((p) => !tenantPagesBySlug.has(p.slug))
```
busbyk, line 117: If I successfully provision a new tenant and then I go and delete one of the template pages (or just change the slug which Sierra did for their donate page as an example) then the onboarding checklist shows as incomplete:

Image

Do we care about that? My immediate reaction was that we could instead store the ids of the pages that were actually created or just store on the tenants collection that all pages were indeed successfully created at one point. I suppose it depends how we want to define "completed" in this case.

src/collections/Tenants/components/onboardingActions.ts
```diff
@@ -0,0 +1,158 @@
...
+      payload.findByID({
+        collection: 'tenants',
+        id: tenantId,
+      }),
+      fs
+        .readFile(path.join(process.cwd(), 'src/app/(frontend)/colors.css'), 'utf-8')
+        .catch(() => ''),
+      fs
+        .readFile(path.join(process.cwd(), 'src/app/api/[center]/og/route.tsx'), 'utf-8')
+        .catch(() => ''),
```
busbyk: Did you test that these files are included in the bundle on Vercel? using `process.cwd()` is correct for serverless environments and these files get imported so they might be available but I'm not sure they would be included in this specific serverless function. 

Did you check that adding a new tenant's slug to the colors.css file results results in the checkbox being marked in the preview env?

Also should we log something in these catch functions?

src/collections/Tenants/components/onboardingActions.ts
```diff
@@ -0,0 +1,158 @@
...
+        .map((p) => ({ slug: p.slug, title: p.title }))
+    }
+
+    const tenantPagesBySlug = new Map(pages.docs.map((p) => [p.slug, p]))
+    const copiedPages = templatePageSlugs.filter((p) => tenantPagesBySlug.has(p.slug))
+    const missing = templatePageSlugs.filter((p) => !tenantPagesBySlug.has(p.slug))
+
+    return {
+      status: {
+        builtInPages: { count: builtInPages.totalDocs, expected: 4 },
```
busbyk: I think we want this to be variable based on whether it's a single zone or a multi zone center.

src/collections/Tenants/components/OnboardingChecklist.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+      <div className="mt-3 border-0 border-t border-solid border-t-[var(--theme-border-color)] pt-3">
+        <h4 className="mb-2">Needs action</h4>
+
+        <ChecklistItem done={theme.brandColors} label="Add brand colors">
+          {loaded && !theme.brandColors && (
+            <span>
+              Add slug to <code>colors.css</code> — see{' '}
+              <Link
+                href="https://github.com/NWACus/web/blob/main/docs/onboarding.md"
```
busbyk: Could use https://github.com/NWACus/web/blob/main/docs/onboarding.md#theme to deep link directly to this section

src/collections/Tenants/endpoints/provisionTenant.ts
```diff
@@ -0,0 +1,563 @@
...
+import { getSeedImageByFilename } from '@/endpoints/seed/utilities'
+import type { BuiltInPage, Page, Tenant } from '@/payload-types'
+import { removeIdKey } from '@/utilities/removeIdKey'
+import type { Payload, PayloadHandler } from 'payload'
+
+const TEMPLATE_TENANT_SLUG = 'dvac'
+
+// Block types that reference tenant-scoped data (teams, sponsors, events) and can't
+// be copied to a new tenant without the corresponding records existing.
+export const TENANT_SCOPED_BLOCK_TYPES = new Set([
```
busbyk, line 11: What about Media, blocks that reference Media (like LinkPreview, ImageText, ImageLinkGrid), Document, and ButtonBlock even has tenant specific fields since it references pages, posts, and built-in pages in the link fields.

busbyk (commented):
When I created a new avalanche center in preview with a valid slug in the AFP, I got an a 404 when visiting their subdomain site: https://btac.endpoint-new-tenant.preview.avy-fx.org/

Image

It's possible that this is related to the `[...segmentsNotFound]` matching issues that will be resolved by #947 but this happens locally too: 

Image

src/collections/Tenants/endpoints/provisionTenant.ts
rchlfryn: Old code - will remove the comment

docs/onboarding.md
rchlfryn: Updated the onboarding doc to include the pages you mentioned. We will also want to automatically check form the AFP if this is a single or multi-zone center, but I will make that in another PR.

src/collections/Tenants/index.ts
rchlfryn: Great call - updated in [2d98a50](https://github.com/NWACus/web/pull/974/commits/2d98a50f2cf6a2936f26f78c07c89d07e4dd42d8)

src/collections/Tenants/components/onboardingActions.ts
rchlfryn: > I suppose it depends how we want to define "completed" in this case.

I think this is a great question. For now it is fine, since we as super admins are the only one using it. I will make a note in the future enhancement section to make an issue later.

src/collections/Tenants/components/onboardingActions.ts
rchlfryn: I made a note of this in future enhancements. I will likely update this in a separate PR.

src/collections/Tenants/endpoints/provisionTenant.ts
rchlfryn: Greatly improved this in #985. Because of all the changes, I will also punt making the pages blank to that PR as well. Is it fair to address this comment in #985?

src/collections/Tenants/hooks/provisionAfterChange.ts
rchlfryn: This was a fun little side quest. Turns out payload's `LoadingOverlayToggle` is just what we needed and has built in state functions like `useFormProcessing` and `useOperation` to help us show the overlay on tenant creation. Check out [51c7df2](https://github.com/NWACus/web/pull/974/commits/51c7df248c7c6fbf0832543703ebb8d93d38d76f)

Kapture 2026-03-20 at 08 24 26

> This would also mean we could remove the skipProvision check here and the setting in the seed script since that component would only trigger in the UI.

I added a `ProvisioningIndicator` component in `beforeDocumentControls` that shows an overlay while the `afterChange` hook is running, which addresses the UX latency concern. But you're right that moving the trigger entirely to a UI component would let us remove the hook and `skipProvision` flag altogether.

rchlfryn: **Update:** I moved provisioning from the `afterChange` hook to the UI using `toast.promise` (following the `SeedButton` pattern), removing the hook and `skipProvision` flag in [3f10c4b](https://github.com/NWACus/web/pull/974/commits/3f10c4b3bcef5b3ae7eea3f4f2010a6e76a025da). The checklist auto-triggers provisioning on mount if needed, shows spinners during, and disables the save button until complete. This means provisioning only runs from the admin UI — not via API or seed script.

Kapture 2026-03-20 at 10 40 30

src/collections/Tenants/components/onboardingActions.ts
rchlfryn: This does work on preview. I added logging as suggested to help if needed when we move to prod, but I did have to move to `centersColorMap` its own file [62856ce](https://github.com/NWACus/web/pull/974/commits/62856ceaa7105c7ae905ddefc4ed86bc5d321072). Scenarios I tested were:
- New tenant: nothing checked
- Tenant with both: css checked, og checked

rchlfryn:
> When I created a new avalanche center in preview with a valid slug in the AFP, I got an a 404 when visiting their subdomain site: https://btac.endpoint-new-tenant.preview.avy-fx.org/

I believe this was fixed by #947. But I did find another 404 issue with the AFP returning incomplete schemas. This is fixed in #996

src/collections/Tenants/endpoints/provisionTenant.ts
busbyk: Yea I think that makes sense. Let's just wait to release this PR until #985 is closed too.

src/collections/Tenants/hooks/provisionAfterChange.ts
busbyk: I like the UI component pattern but also whoa good to know about `LoadingOverlayToggle`!

__tests__/e2e/admin/tenant-selector/sync-on-save.e2e.spec.ts
```diff
@@ -124,13 +152,14 @@ test.describe('Tenant selector syncs on save', () => {
         page.waitForURL('**/admin/collections/pages'),
         page.locator('nav a[href="/admin/collections/pages"]').click(),
       ])
-      await page.waitForLoadState('networkidle')
+      await page.waitForLoadState('domcontentloaded')
 
       const options = await getTenantOptions(page)
       expect(options).toContain(UPDATED_TENANT_NAME)
-      expect(options).not.toContain(TEMP_TENANT_NAME)
+      // FIXME
```
rchlfryn: We currently can't delete any tenant so this test is flaky until #976 is fixed

busbyk (approved):
🎉

PR #979, opened 2026-03-07
Fix ViewPostButton to show for all published posts

### LINKED ISSUE #709: Add live page button to Posts
We currently have a live page button on Pages (when clicked, opens the published page in a new tab). It would be nice to add this to Posts. 

Image

https://github.com/NWACus/web/blob/main/src/collections/Pages/components/ViewPageButton.tsx
https://github.com/NWACus/web/blob/main/src/collections/Pages/index.ts#L88

busbyk:
## Description

The ViewPostButton (live page button on Posts) was using `getCanonicalUrlForSlug()` which searches the **navigation menu** for the post's slug. Since most blog posts aren't linked in navigation, the function returned `null` and the button never rendered. This fix constructs the URL directly as `/{center}/blog/{slug}`, matching the actual route pattern.

## Related Issues

Closes #709

## Key Changes

- Removed `getCanonicalUrlForSlug` dependency — posts don't need navigation lookup since they always live at `/blog/{slug}`
- Reduced query `depth` from 2 to 0 (only `slug`, `_status`, and `tenant` are needed)
- Moved `_status` check before `resolveTenant()` to avoid unnecessary work for draft posts

## How to test

1. Run `pnpm dev` and navigate to `/admin`
2. Open any **published** post in the admin editor
3. Verify the globe button appears in the document controls (top right)
4. Click the globe button — it should open `/{center}/blog/{slug}` in a new tab
5. Open a **draft** post — the globe button should NOT appear

## Migration Explanation

No migration needed — this is a UI-only change.


rchlfryn (changes requested):
Looks good - can we add a test?

busbyk:
> Looks good - can we add a test?

Tests added 🕺

__tests__/e2e/admin/view-document-button.e2e.spec.ts
rchlfryn, line 1: ⛏️ There are a lot of comments in this file. Do we want that?

I am fine either way but I think test variable should help explain what is happening which they do.

rchlfryn (approved):
Nit pick - but LG

__tests__/e2e/admin/view-document-button.e2e.spec.ts
busbyk: Cleaned up the comments a bit. A lot of them felt helpful to me but I see your point. Hopefully this is a nice middle ground.

Issue #540, opened 2025-09-24
linkToPageOrPost field should use canonical urls


busbyk:
Right now when the linkToPageOrPost field is used for an internal link, the url used is the non-navigation-nested url. i.e. /about-us vs. /about/about-us. 

The src/app/(frontend)/[center]/[slug]/page.tsx route handles redirecting to the navigation-nested url (i.e. canonical url) if the page is nested in the navigation. But ideally we would use the navigation-nested urls. 

This is not critical by any means but if we've gone to the trouble of nesting pages under their navigation nesting, we should respect that everywhere. Doing so would mean that hovering over a button that is powered by the linkToPageOrPost field would display the navigation nested url and a user could copy that and have the correct link.

PR #981, opened 2026-03-07
Fix Safari 404 on donate button by using canonical URL

### LINKED ISSUE #746: 404 on SAC Donate page in Safari
**Describe the bug**
The Donate button sends the user to a different URL (+ 404) on Safari than Chrome

**To Reproduce**

1. Go to SAC homepage on Safari
2. Click on Support > Donate
3. Click on Donate (in nav)

**Screenshots**
Chrome working as expected:

Image

Safari:

Image

**Environment details**

- Desktop
- Chrome

**Existential musings**
What else you got for us?

busbyk:
## Description

Fixes a Safari-only bug where clicking the donate button results in a 404. The donate nav item was linking to `/{slug}` (e.g., `/donate`), relying on a server-side `redirect()` to the canonical nested URL (`/support/donate`). Safari mishandles this redirect — likely due to a duplicate `Location` header bug in Next.js ([vercel/next.js#82117](https://github.com/vercel/next.js/issues/82117)) — producing a mangled URL like `/support/donate,%20/support/donate`.

The fix resolves the canonical URL at nav-build time by looking up the donate page in the already-built navigation tree using `findNavigationItemBySlug`, avoiding the redirect entirely.

## Related Issues

Fixes #746

Related upstream issues:
- [vercel/next.js#82117](https://github.com/vercel/next.js/issues/82117) — Duplicate Location header with redirect/permanentRedirect
- [vercel/next.js#48309](https://github.com/vercel/next.js/issues/48309) — Safari redirect fails with App Router
- [WebKit Bug #77538](https://bugs.webkit.org/show_bug.cgi?id=77538) — Safari aggressive 302 redirect caching

## Key Changes

- `src/components/Header/utils.ts`: After building the donate `NavLink`, if it's an internal page link, look up the page's slug in the already-built `topLevelNavItems` to get the canonical (navigation-nested) URL. This means the donate button links directly to e.g. `/support/donate` instead of `/donate`, bypassing the problematic redirect.

## How to test

1. On a tenant with a donate page nested under Support (e.g., SAC, Sierra), verify the donate button in the nav links to the canonical nested URL (e.g., `/support/donate`) rather than the bare slug (`/donate`)
2. Test in Safari specifically — clicking the donate button should navigate correctly without a 404
3. Verify other nav items still work correctly

## Screenshots / Demo video

Demo: https://www.loom.com/share/3acc89bf68a9405480312972784bfe4d

## Migration Explanation

No migration needed — this is a runtime behavior change only.

- The underlying redirect from `/{slug}` → `/{parent}/{slug}` still exists for direct URL visits and should continue working. If the Next.js duplicate `Location` header bug is confirmed fixed in a future version, the redirect path will also be safe for Safari.
- Related: #540 (canonical URLs) and #752 (better nav page handling) track broader improvements to how page URLs are resolved.

### DIFF
--- a/src/components/Header/utils.ts
+++ b/src/components/Header/utils.ts
@@ -387,6 +387,7 @@ export const getTopLevelNavItems = async ({
     if (link) {
       // For internal page links, resolve the canonical (navigation-nested) URL
       // to avoid a redirect from e.g. /donate -> /support/donate which Safari mishandles
+      // see https://github.com/NWACus/web/pull/981 for more context
       if (link.type === 'internal') {
         const slug = link.url.split('/').filter(Boolean).pop()
         if (slug) {
--- a/src/components/Header/utils.ts
+++ b/src/components/Header/utils.ts
@@ -385,6 +385,18 @@ export const getTopLevelNavItems = async ({
     const link = convertToNavLink(navigation.donate.link)
 
     if (link) {
+      // For internal page links, resolve the canonical (navigation-nested) URL
+      // to avoid a redirect from e.g. /donate -> /support/donate which Safari mishandles
+      if (link.type === 'internal') {
+        const slug = link.url.split('/').filter(Boolean).pop()
+        if (slug) {
+          const navItem = findNavigationItemBySlug(topLevelNavItems, slug)
+          if (navItem?.link?.type === 'internal') {
+            link.url = navItem.link.url
+          }
+        }
+      }
+
       donateNavItem = {
         label: link.label,
         link,
--- a/redirects.js
+++ b/redirects.js
@@ -12,7 +12,14 @@ const redirects = async () => {
     source: '/:path((?!ie-incompatible.html$).*)', // all pages except the incompatibility page
   }
 
-  const redirects = [internetExplorerRedirect]
+  // Strip duplicate path segments after URL-encoded space (e.g. "/foo,%20foo" -> "/foo")
+  const trailingDuplicateRedirect = {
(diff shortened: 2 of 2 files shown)


busbyk:
We may also want to set this in our next config: https://github.com/vercel/next.js/issues/82117#issuecomment-3714254364

src/components/Header/utils.ts
```diff
@@ -385,6 +385,18 @@ export const getTopLevelNavItems = async ({
     const link = convertToNavLink(navigation.donate.link)
 
     if (link) {
+      // For internal page links, resolve the canonical (navigation-nested) URL
+      // to avoid a redirect from e.g. /donate -> /support/donate which Safari mishandles
```
rchlfryn, line 389: ⛏️  maybe add a note about the pr or issue for more info, or include a note about `%20`

rchlfryn (commented):
LG - did you test the preview link on Safari?

Also I don't see why we wouldn't include the workaround in our next config.

src/components/Header/utils.ts
busbyk: Good call 👍

busbyk:
> LG - did you test the preview link on Safari?

I added the redirect logic (which was definitely still necessary! I had just been focused on making sure we were using the navigation-nested (canonical) url for internal links for the donate button) and recorded a demo (see the PR description). 

Ready for re-review.

rchlfryn (approved):

PR #985, opened 2026-03-12
Enhance duplicate page action

### LINKED ISSUE #625: Duplicate page to... enhancements

rchlfryn:
## Description

Improves the page duplication and provisioning logic introduced in #974. The `clearLayoutRelationships` utility properly clears relationship and upload fields from any block — keeping the block structure intact and letting editors repopulate the tenant-specific references.

Provisioning is now driven by **DVAC's navigation structure**. It queries the template tenant's published navigation to determine which pages and built-in pages to create — only pages referenced in navigation are provisioned. Pages are created as **blank published pages** (title, slug, empty content block) with no copied content. Built-in pages are also filtered to those referenced in the navigation. Navigation and all pages are published immediately. The `duplicatePageToTenant` action still uses `clearLayoutRelationships` to copy layout structure while clearing tenant-specific references, and now only appends " - Copy" to the title/slug when a page with the same slug already exists for the target tenant.

The `clearLayoutRelationships` utility currently handles all field types in Payload's schema (`group`, `array`, `tabs`, `row`, `collapsible`, `richText` with Lexical blocks), including Lexical-embedded blocks like `ButtonBlock` and `CalloutBlock`. If new field container types are added to Payload in the future, the utility would need updating.

## Related Issues

Part of #533 (follow-up to #974)
Closes #625

## Key Changes

- **Navigation-driven provisioning** — queries DVAC's published navigation to extract referenced page slugs and built-in page URLs via `extractNavReferences`. Only pages and built-in pages that appear in the navigation are created for the new tenant. If DVAC's navigation changes, provisioning automatically adjusts.
- **`clearLayoutRelationships` utility** — recursively clears `relationship` and `upload` fields from any block in a page layout (including nested groups, arrays, tabs, and Lexical rich-text blocks with `ButtonBlock` and `CalloutBlock`), keeping structure intact so editors repopulate tenant-specific references after duplication.
- **Provisioning creates blank published pages** — pages use DVAC's titles and slugs but contain only an empty content block. Navigation and pages are published immediately.
- **Conditional " - Copy" suffix** — `duplicatePageToTenant` now only appends " - Copy" to the title and "-copy" to the slug when a page with the same slug already exists for the target tenant.
- **Bug fix** — `duplicatePageToTenant` was spreading the cleaned layout array into `data` instead of assigning it as `data.layout`.
- **Tests** — unit tests for `clearLayoutRelationships` (relationships, uploads, groups, arrays, tabs, Lexical blocks, id stripping) and integration tests for `duplicatePageToTenant` (slug conflict and no-conflict cases), backed by a shared `fixtures/mockBlocks.ts`.

## How to test

Added tests

1. `pnpm seed` to get a fresh database with the DVAC template tenant
2. **Provisioning**: Create a new avalanche center in the admin panel. Verify the onboarding checklist shows automated steps as complete. Only pages and built-in pages referenced in DVAC's navigation are created. Pages are blank (empty content block) and published.
3. **Duplicate to Tenant**: Open any published page on DVAC → ⋮ menu → "Duplicate to Tenant" → pick another tenant. Verify the new page is created as a draft with the original title/slug (no " - Copy" if no conflict), relationship fields are cleared (empty), and non-relationship fields (text, colors, layout structure) are preserved.

## Screenshots / Demo video
Screenshot 2026-03-11 at 17 48 34

[Duplicate page to... demo](https://www.loom.com/share/8ea9142397be4087aca4c1cef40fb97d)
[Provisioning tenant demo](https://www.loom.com/share/6fd514dddb414109b13f2cb43df45557)

## Migration Explanation

No schema changes — no migration needed.

## Future enhancements

Should we include `ButtonBlock` &  `CalloutBlock` in `DEFAULT_BLOCKS`? They are the only 2 not included. #1001


src/collections/Tenants/components/OnboardingChecklist.tsx
```diff
@@ -132,10 +138,13 @@ export function OnboardingChecklist() {
...
         {pages.missing.length > 0 && <div>Missing: {pages.missing.join(', ')}</div>}
         {pages.skipped.length > 0 && <div>Skipped (demo pages): {pages.skipped.join(', ')}</div>}
+        <div>Copied as drafts — review and publish</div>
       </ChecklistItem>
 
       <ChecklistItem done={homePage} label="Home page" />
-      <ChecklistItem done={navigation} label="Navigation" />
+      <ChecklistItem done={navigation} label="Navigation">
+        {navigation && <div>Created as a draft — review and publish</div>}
+      </ChecklistItem>
```
busbyk: It would be sweet if these were links to the relevant list views but obv that's optional since a super admin should know where to go anyways. 

Image

src/utilities/clearLayoutRelationships.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const allBlocksMap = new Map([...DEFAULT_BLOCKS, NACMediaBlock].map((b) => [b.slug, b]))
+
+function isDataObject(value: unknown): value is DataObject {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
+
+function clearFieldValuesInPlace(data: DataObject, fields: Field[]): void {
+  // Strip 'id' (not fields like 'videoId') so new IDs are assigned on create
+  Reflect.deleteProperty(data, 'id')
```
busbyk, line 21: Whoa, TIL about Reflect

src/collections/Pages/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
```diff
@@ -17,11 +17,8 @@ import { useRouter } from 'next/navigation'
 import { formatAdminURL } from 'payload/shared'
 import { useCallback, useState } from 'react'
 
-// TODOs
-// - Remove photos from blocks or use a global photo?
-
 export const DuplicatePageForDrawer = () => {
-  const { savedDocumentData: pageData } = useDocumentInfo()
+  const { data: pageData } = useDocumentInfo()
```
busbyk, line 22: Oh did this change in a recent version of Payload?

src/utilities/clearLayoutRelationships.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (isDataObject(lexicalData.root)) {
+    walkNode(lexicalData.root)
+  }
+}
+
+/**
+ * Clears relationship and upload fields from a page layout so
+ * tenant-scoped references can be repopulated after duplication.
+ */
+export function clearLayoutRelationships(layout: Page['layout']): Page['layout'] {
```
busbyk, line 97: I feel like this function and this file should be called `clearBlocksFieldRelationships` or `clearBlocksRelationships` because this could be used on any `blocks` type field right? Your passing in the `layout` key so you're not accessing the `layout` key in this function - this could be named more generically. 

Although I guess the trouble with that is that we couldn't use a Payload generic type like `BlocksField` because the `Page['layout']` type is based on the specific block types allowed in the [blocks config](https://github.com/NWACus/web/blob/59d85d37ba90ace47a74fa2f75416e758733638f/src/collections/Pages/index.ts#L68). Hmm I suppose this is specific to the pages collection then 🤔

Still leaving this comment bc I think it's a valuable thinking-out-loud thought 😁

__tests__/server/fixtures/mockBlocks.ts
busbyk, line 1: Same question as [here](https://github.com/NWACus/web/pull/974#discussion_r2943618445) about additional block types

__tests__/server/duplicatePageToTenant.server.test.ts
```diff
@@ -0,0 +1,113 @@
+jest.mock('../../src/constants/defaults', () => ({
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  DEFAULT_BLOCKS: require('./fixtures/mockBlocks').DEFAULT_BLOCKS,
+}))
+
+jest.mock('../../src/blocks/NACMedia/config', () => ({
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  NACMediaBlock: require('./fixtures/mockBlocks').NACMediaBlock,
+}))
```
busbyk, line 9: Why do these need to be mocked?

__tests__/server/duplicatePageToTenant.server.test.ts
```diff
@@ -0,0 +1,113 @@
...
+describe('duplicatePageToTenant', () => {
+  it('creates the page as a draft', async () => {
+    const req = buildRequest('42', {
+      newPage: { title: 'About', slug: 'about', layout: [] },
+    })
+    await duplicatePageToTenant(req)
+    expect(mockCreate).toHaveBeenCalledWith(expect.objectContaining({ draft: true }))
+  })
+
+  it('appends " - Copy" to the title and "-copy" to the slug', async () => {
```
busbyk: This functionality has always bothered me about the duplicate page to functionality 😅 -- ideally, we would only append "-copy" if there's a page with a matching slug right? Maybe it makes sense to update this in a separate PR but we definitely have the ability to see if there's a matching slug already and add "-copy" if so but use the same slug if not.

busbyk (commented):
One relationship removal isn't working (ButtonBlock inside of a richText field in EventList (likely BlogList too)): https://www.loom.com/share/777dddf99b784ec7b60e67ffa792a844

src/utilities/clearLayoutRelationships.ts
rchlfryn: Same...pretty cool eh?

src/collections/Pages/components/DuplicatePageFor/DuplicatePageForDrawer.tsx
rchlfryn: I think you changed this in #929. But yes there was a linting error

src/collections/Tenants/components/OnboardingChecklist.tsx
rchlfryn: This is a good idea but it is more of a headache to code than the benefit of the link. For pages we would need to make sure the tenant selector selects the right tenant. And for navigation, we need to query to make sure we navigate to the corresponding tenant. 

This will also likely be removed with adding blank pages.

rchlfryn: Update: this was removed since we are using blank pages now

__tests__/server/duplicatePageToTenant.server.test.ts
rchlfryn: I was running into issue importing the real `DEFAULT_BLOCKS` since pulls in the full Payload block config chain: `DEFAULT_BLOCKS` → `BlogList/config` → `validateMaxPosts` → `@payloadcms/richtext-lexical` → fails because of module format incompatibility

Jest runs in CommonJS mode mode and can't transpile Payload's ESM modules. The mock fixtures give clearLayoutRelationships just enough field metadata (relationship/upload type info) to work without dragging in the Payload runtime. We should revisit this as a larger decision.

__tests__/server/fixtures/mockBlocks.ts
rchlfryn: Marking as resolved and will continue discussion below.

src/utilities/clearLayoutRelationships.ts
rchlfryn: Thanks for sharing this.

Agreed and I went back and forth. We could rename it to `clearBlocksRelationshipsForLayout`?

__tests__/server/duplicatePageToTenant.server.test.ts
rchlfryn: good call out - implemented in [only append copy to page title and slug if page exists](https://github.com/NWACus/web/pull/985/commits/c367dc418ac7a4d87e1d695153026335e99d77be)

rchlfryn:
> One relationship removal isn't working (ButtonBlock inside of a richText field in EventList (likely BlogList too))

Good catch! `ButtonBlock` & `CalloutBlock` are the only 2 not included in DEFAULT_BLOCKS?  I added a note under Future Enhancement of this PR to make an issue since I do not want to scope creep this PR.

I added `ButtonBlock` & `CalloutBlock` to the `allBlocksMap` in `clearLayoutRelationships` so it will copy as expected now.

src/collections/Tenants/components/OnboardingChecklist.tsx
busbyk: If we ever need something like this in the future (i.e.a navigation where we also change the tenant), I believe we could do this with an onClick handler and the `useTenantSelection` hook's `setTenant`:

```
onClick={() => {
  setTenant({ slug: 'nwac' })
  router.push('/')
}
```

src/utilities/clearLayoutRelationships.ts
busbyk: I think it's fine because it's named appropriately for what it's doing - you just have to know that the layout field is a blocks field which is fine. It would be sweet to have this be generic but we don't need it for any other fields so there's no point haha.

__tests__/server/duplicatePageToTenant.server.test.ts
busbyk: Ahhhh yes. Similar problem to why I had to put https://github.com/NWACus/web/blob/main/src/components/Header/utils-pure.ts in it's own file for testing so I didn't import next.js specific things related to caching.

src/collections/Tenants/endpoints/provisionTenant.ts
```diff
@@ -93,7 +82,7 @@ export const provisionTenant: PayloadHandler = async (req) => {
  * Provisions a tenant with all default data:
  * 1. Website Settings with placeholder brand assets (logo, icon, banner)
  * 2. Built-in pages (forecasts, weather stations, observations)
- * 3. Pages copied from the template tenant (DVAC)
+ * 3. Blank pages matching the template tenant's (DVAC) page structure
```
busbyk: Can we update this to make it clearer that we're creating pages matching the page titles and slugs in DVAC's navigation, not the actual page structures? i.e. we're not copying any content within each page. We're just using the pages in DVAC's navigation to know what slugs to create.

src/collections/Tenants/endpoints/provisionTenant.ts
busbyk, line 1: RE blank page creation: the way this is written now, every single DVAC page will result in a page in the new tenant with blank content. If we create some example pages in DVAC that show things like sidebar layouts or other examples that aren't required for our default navigation structure, those pages would get copied over too. 

Is that what we want? Or do we really just want to create _only_ the pages needed for the navigation structure?

I think it's the latter. I think it would be nice if we only created the pages needed for the navigation and then we could decide (as a super admin) to pull over additional pages from DVAC if we wanted to. 

@rustynwac what do you think about that?

src/collections/Tenants/endpoints/provisionTenant.ts
rchlfryn: That was my intention to leave a note that says we are copying the pages from the navigation, but it is definitely confusing. I will update it.

src/collections/Tenants/endpoints/provisionTenant.ts
rchlfryn: Decision made in slack: Use DVAC navigation to structure which pages to copy, and do not copy content

busbyk (approved):

PR #997, opened 2026-03-22
Tenant improvements on addition and ability to delete

### LINKED ISSUE #976: Support deleting an avalanche center
## Summary

Deleting a tenant (avalanche center) currently has no cleanup step — tenant-scoped data (pages, settings, media, navigation, built-in pages, home pages, etc.) is left orphaned when the tenant record is deleted.

## What's needed

1. **Confirmation step** — since this is very invasive, add a confirmation prompt that requires typing the avalanche center name before allowing deletion.
2. **Cleanup hook** — a `beforeDelete` hook that removes all tenant-scoped data before the tenant record itself is deleted. A prior implementation (`deprovisionBeforeDelete`) was started but removed in commit [f074b15](https://github.com/NWACus/web/pull/974/commits/f074b150a67b0284bc3612fa2fe4495f35303349).

## References

- Commit that removed the hook: [f074b15](https://github.com/NWACus/web/pull/974/commits/f074b150a67b0284bc3612fa2fe4495f35303349)
- Tenant collection config: `src/collections/Tenants/index.ts`

rchlfryn:
## Description

While working on onboarding tickets, I though of some improvements we can make for the tenant (avalanche center) creation and deletion (which was broken) experience in the admin UI.

## Related Issues

Fixes #976

## Key Changes

**Slug field: only show unused slugs on creation**
- The slug field uses a custom `TenantSlugField` component that fetches existing tenants and filters out already-used slugs when creating a new tenant. We can't use Payload's built-in `filterOptions` here because it only works on `relationship` fields, not `select` fields — the slug is a `select` with a static list of valid avalanche center identifiers, so filtering must happen client-side.
- When editing an existing tenant, all options are shown (the field is read-only on update anyway).

**Auto-fill name from slug**
- An invisible `AutoFillNameFromSlug` component syncs the name field to the full avalanche center name when a slug is selected during creation.

**Delete confirmation with type-to-confirm**
- Deleting a tenant is extremely destructive — it cascades across 17 tenant-scoped collections (pages, posts, media, settings, navigation, etc.). Payload's built-in delete confirmation is a simple "Are you sure?" which is not sufficient for an operation this invasive.
- A `DeleteTenantModal` component intercepts Payload's default delete modal and replaces it with a custom one that requires typing the tenant name to confirm. This uses Payload's `Modal` directly rather than `ConfirmationModal` because the latter doesn't support controlling button state based on user input.

**Deprovision hook**
- A `beforeDelete` hook (`deprovisionBeforeDelete`) deletes all tenant-scoped data before the tenant record itself is removed, preventing orphaned data and SQLite foreign key constraint errors. This runs regardless of how deletion is triggered (admin UI, API, etc.).

## How to test
Create a new tenant:
-  Go to Admin > Avalanche Centers > Create New
1.   Slug field
    - Above name field now
    - The slug dropdown should only show slugs not already in use. 
    - Edit an existing tenant — all slugs should be visible (field is read-only).
2. Name field
    - On the create form, select a slug — the name field should auto-populate with the full center name.
    - You should be able to edit this if you want
3. Delete confirmation 
    - Open an existing tenant, click the actions menu > Delete. 
    - The custom modal should appear requiring you to type the tenant name. 
    - Typing the wrong name and clicking Delete should show an error toast. 
    - Typing the correct name should proceed with deletion.
4. Deprovision cleanup
    - After deletion, verify no orphaned data remains for that tenant across collections (pages, settings, media, etc.).

## Screenshots / Demo video

https://www.loom.com/share/af64206841684c02ae1c75c44e546ae9

## Migration Explanation

No migration needed — no schema changes.

How do we feel about the custom slug field?


__tests__/server/deprovisionBeforeDelete.server.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      pages: 5,
+      posts: 3,
+      media: 10,
+    }
+    const mockPayload = buildMockPayload(findResults)
+    const mockReq = { payload: mockPayload }
+
+    await deprovisionBeforeDelete(
+      // @ts-expect-error - partial mock; only id and req.payload are used by the hook
+      { id: 42, req: mockReq },
```
busbyk, line 40: nice

src/collections/Tenants/hooks/deprovisionBeforeDelete.ts
```diff
@@ -0,0 +1,51 @@
+import type { CollectionBeforeDeleteHook } from 'payload'
+
+export const TENANT_SCOPED_COLLECTIONS = [
```
busbyk, line 3: [17:30:01] ERROR (payload): Failed query: delete from "tenants" where "tenants"."id" = ?
params: 2
    err: {
      "type": "DrizzleQueryError",
      "message": "Failed query: delete from \"tenants\" where \"tenants\".\"id\" = ?\nparams: 2: SQLITE_CONSTRAINT_NOTNULL: NOT NULL constraint failed: forms.tenant_id: NOT NULL constraint failed: forms.tenant_id",

src/collections/Tenants/hooks/deprovisionBeforeDelete.ts
```diff
@@ -0,0 +1,51 @@
...
+  'eventTags',
+  'sponsors',
+  'tags',
+  'teams',
+  'biographies',
+  'documents',
+  'media',
+  'redirects',
+  'roleAssignments',
+] as const
```
busbyk, line 22: `consistent-type-assertions.txt` needs to be updated for this.

Too bad we need to record this as a "bad" `as` usage because this is actually type narrowing not casting.

src/collections/Tenants/components/TenantSlugField.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useEffect, useState } from 'react'
+
+export const TenantSlugField = (props: SelectFieldClientProps) => {
+  const { id } = useDocumentInfo()
+  const [usedSlugs, setUsedSlugs] = useState<string[]>([])
+
+  useEffect(() => {
+    if (id) return
+
+    fetch('/api/tenants?limit=0&depth=0&select[slug]=true')
```
busbyk: You could use [`usePayloadAPI`](https://payloadcms.com/docs/admin/react-hooks#usepayloadapi) to avoid the `usedSlugs` state hook. 

I also wonder if this could just be a server component?

src/collections/Tenants/components/TenantSlugField.tsx
```diff
@@ -0,0 +1,29 @@
...
+      .then((res) => res.json())
+      .then((data) => setUsedSlugs(data.docs.map((doc: { slug: string }) => doc.slug)))
+      .catch(() => {})
+  }, [id])
+
+  const options = id
```
busbyk: Why the different behavior on edit vs. create? Just because edit is a read only field so not worth filtering?

__tests__/e2e/admin/collections/tenants.e2e.spec.ts
busbyk, line 1: 👏

busbyk (changes requested):

src/collections/Tenants/hooks/deprovisionBeforeDelete.ts
rchlfryn: how did you get this error?

src/collections/Tenants/components/TenantSlugField.tsx
rchlfryn: Ooo good call about `usePayloadAPI` - updated to a server component

src/collections/Tenants/hooks/deprovisionBeforeDelete.ts
rchlfryn: Actually I learned this with this PR but  `as const` isn't a type assertion — it's a const assertion. The ESLint rule consistent-type-assertions targets type assertions like `x as Foo`, not `as const`.

src/collections/Tenants/components/TenantSlugField.tsx
rchlfryn: Removed this "feature" when converting to server component

src/collections/Tenants/hooks/deprovisionBeforeDelete.ts
busbyk: Oh my bad I had run `pnpm verify-type-assertions` locally and gotten an error.

src/collections/Tenants/hooks/deprovisionBeforeDelete.ts
busbyk: Oh jeez I thought I'd included context here - that's terrible haha. 

What I _thought_ I'd written was:

I believe you need to add the forms collection to this list because you get this NOT NULL constraint failed on the form that gets created in our seed data. https://www.loom.com/share/5b53172aed9048ec83b367fcd4412266

busbyk: I just pushed that change since it was super quick. [0027fa7](https://github.com/NWACus/web/pull/997/commits/0027fa73319607c73188e7324da04e84e041f11d)

busbyk (approved):

src/collections/Tenants/hooks/deprovisionBeforeDelete.ts
rchlfryn: thank you!

PR #1002, opened 2026-03-23
Unify navigation tabs with `displayMode` and use built in pages for autofilled links

### LINKED ISSUE #825: Navigation improvements - donate tab & sort link refs

### LINKED ISSUE #988: Use built-in page references for automated top level nav items
There are a few top level nav items we are not displaying on the admin side.: forecasts, observations, blog, and events. Since adding built-in page, we should change the UI to use a built-in page references for these tabs. These nav items should be readonly.

rchlfryn:
## Description

Restructures the Navigation admin UI and seeds built-in page references for forecasts, observations, blog, and events so their links appear as proper CMS relationships rather than being hardcoded. Also unifies all nav tabs behind a single `navTab` helper with a `displayMode` field, letting admins pick how each tab renders (dropdown, single link, or button) without a code change.

## Related Issues

Fixes #988
Fixes #825

Follow-ups:
- #1038 (AFP → navigation sync system)
- #1042 (warn when deleting documents referenced in navigation)

## Key Changes

- **Unified `navTab` helper** — replaces the old `topLevelNavTab` / `singleLinkNavTab` split. Every tab has the same shape: a `displayMode` radio (Dropdown / Link / Button) alongside an enabled toggle, with `link` / `items` fields conditional on the selected mode. This gives admins flexibility while keeping all tabs rendered through the same code path.
- **`BannerDescription` component** — reusable Payload admin component that wraps `@payloadcms/ui`'s `Banner` with an icon, used for contextual info banners on nav tabs.
- **Tab descriptions** — every tab now has an inline banner explaining what it does, where its data comes from, and whether admin edits may be overwritten by autofill.
- **Stale-data clearing on mode switch** — switching a tab's `displayMode` clears the fields not used by the new mode. A custom `DisplayModeField` client component wipes the in-form values immediately (no save needed), and a collection `beforeChange` hook (`clearUnusedTabData`) provides a server-side safety net so the DB stays clean even if the admin bypasses the UI.
- **Universal button rendering** — any tab with `displayMode: 'button'` renders as a styled callout button in both desktop and mobile nav. Donate is no longer special-cased in `Header/utils.ts` or the nav components; it flows through the same unified array as every other tab.
- **Forecasts rendering** — `Header/utils.ts` now handles nested nav items (items with sub-items render as accordion sections in the dropdown). Multi-zone centers show "All Forecasts" + a "Zones" section with per-zone links.
- **Seed** — fetches active forecast zones per tenant from NAC API; creates zone-specific built-in pages; populates forecasts, observations, blog, and events navigation entries with proper `builtInPage` references.
- **Data migration** — backfills existing tenants' built-in pages and navigation references so no standalone post-deploy script is needed.

**Per-tab defaults (admin can change via `displayMode`):**
- **Forecasts** — dropdown, autofilled from NAC zone data. Single-zone centers (e.g. SAC) show "Avalanche Forecast"; multi-zone centers (e.g. NWAC) show "All Forecasts" + a "Zones" accordion with per-zone items sorted by rank.
- **Observations** — dropdown, autofilled with "Recent Observations" and "Submit Observations".
- **Blog / Events** — single link to landing page.
- **Donate** — button, to donation page.
- **Weather / Education / Accidents / About / Support** — dropdown, admin-configured.
- **`isRecord` type guard** — extracted to `src/utilities/isRecord.ts`; `findDocumentsWithBlockReferences` and the new nav hook both consume it instead of redefining locally.

## How to test

- Run `pnpm seed`
- In admin, open Navigation. For each tab, verify the `displayMode` radio defaults as listed above and the admin UI swaps between link/items appropriately when toggled.
  - Forecasts
    - single-zone center (SAC): single "Avalanche Forecast" item, no zones section
    - multi-zone center (NWAC): "All Forecasts" item + "Zones" accordion with per-zone items
  - Observations: "Recent Observations" and "Submit Observations" as items
  - Blog / Events: link to their landing page
  - Donate: link renders as a button on the site
- Flip a tab's `displayMode` and confirm the visible fields change (items hidden, link shown, etc.) without data loss.

## Screenshots / Demo video
https://www.loom.com/share/2f9df8650db64dd6af417a94570e1bf3

## Migration Explanation

Two migrations (run in order):

**`20260418_120000_convert_auto_nav_items.ts`** — schema migration. Adds the `options.displayMode` column plus `link` and `items` structures to every tab in `navigations` (and `_navigations_v`). Defaults per tab: `dropdown` for forecasts/observations/weather/education/accidents/about/support; `link` for blog/events; `button` for donate. Existing rows inherit defaults via `ALTER TABLE ADD COLUMN DEFAULT`.

**`20260418_120100_backfill_nav_builtin_pages.ts`** — data migration. Iterates all tenants, creates missing built-in pages (forecast zones, blog, events, observations, weather stations, mountain weather), and populates each tenant's navigation record with the new built-in page references. Uses raw SQL to bypass document-level validation, which would otherwise fail on pre-existing invalid fields in unrelated tabs. Clears stale draft versions so Payload regenerates them from the updated main tables on next load.

The data migration is self-contained — it fetches the NAC API directly via `fetch` rather than importing from `@/services/nac`, so it stays stable if service modules change later. If the NAC API is unreachable for a tenant at deploy time, that tenant's forecasts portion is skipped with a warning and the deploy continues; re-saving that tenant's navigation in the admin repopulates it.

Because the schema and data migrations ship together, there is no deploy-gap window between code deploy and data backfill — the moment the deploy completes, navigation is fully populated and the frontend can render real CMS links without hardcoded fallbacks.

- The real AFP → navigation sync system is tracked in #1038. The one-time backfill migration in this PR is not a substitute for sync.
- Pages/built-in pages referenced in navigation can still be deleted silently (pages cascade; built-in pages hit a cryptic FK error). A UI indicator + warning flow is tracked in #1042 — useful follow-up now that nav references are first-class CMS relationships.


src/scripts/migrate-nav-builtin-pages.ts
busbyk, line 1: Would running provisioning after the updates from https://github.com/NWACus/web/pull/1011 not work because we need to clear the old values? I would have thought the migration in this PR would delete the old data and then we could just re-run provisioning manually for the couple production tenants. 

I'm also curious about using a migration to run this logic. That's the approach I chose for #1025

src/collections/BuiltInPages/index.ts
```diff
@@ -37,6 +41,15 @@ export const BuiltInPages: CollectionConfig<'pages'> = {
...
     },
+    {
+      name: 'isInNav',
+      type: 'checkbox',
+      defaultValue: false,
+      admin: {
+        hidden: true,
+        description: 'Pages used in navigation cannot be deleted to prevent broken links.',
+      },
+    },
```
busbyk: What sets this field? Looks like just the one-time data migration script. So how would this field be kept up to date? 

We'd need a hook to keep this up to date. I think that hook would need to live in the navigations collection. 

I'm not convinced we need this though. We have logic to filter out "invalid" links in https://github.com/NWACus/web/blob/nav-updates/src/components/Header/utils.ts#L256. 

Although it does seem like something we would want to warn the user about. And I say warn because I don't think we'd want to restrict delete for a page that's in the nav if a center admin really wants to delete it because they couldn't remove it from the nav in order to delete it because we don't allow them to edit the nav. And they could mark it as a draft page (which is also filtered out in the nav) anyways. 

So maybe a warning. We could do this with an OnOperation (on delete) hook. But honestly it might just be better to have a virtual field or some custom ui component that indicate whether or not this page is in the nav dynamically (not saving it to the db like this isInNav field requiring a hook to keep up to date) and just have it display.

src/components/Header/utils.ts
```diff
@@ -261,122 +255,77 @@ export function convertToNavLink(
 
 export const getTopLevelNavItems = async ({
```
busbyk, line 54: I'm skeptical of these fallbacks. If we want to commit to using built-in page/page/post references for these previously dynamic and hardcoded top level nav items, why have fallbacks? If it's just for the period of time in between deploying this change and running the `scripts/migrate-nav-builtin-pages.ts`, that could be solved by using the migration file to do data migration like the approach I used in https://github.com/NWACus/web/pull/1025

These fallbacks also mean that if all links in a top level nav section were removed, these fallbacks would be shown. Which could be confusing behavior / not wanted at all. Ideally, the section would just be hidden but 🤷

src/collections/Navigations/fields/topLevelNavTab.ts
```diff
@@ -1,31 +1,70 @@
...
   description,
   hasConfigurableNavItems = true,
+  hasReadOnlyLink = false,
+  hasReadOnlyNavItems = false,
   hasEnabledToggle = true,
   enabledToggleDescription = 'If hidden, pages with links in this nav item will not be accessible at their navigation-nested URLs.',
 }: {
   name: string
   description?: string
   hasConfigurableNavItems?: boolean
```
busbyk: I think we'd want to remove this prop now right?

src/collections/Navigations/fields/itemsField.ts
busbyk, line 1: Depending on your reply about the `hasReadOnly___` props / read only nature of the modified top level nav items, this might make sense to revert.

busbyk (changes requested):
I was imagining a solution that used the center's config from the AFP to keep built-in page references in sync in our navigations collections. 

https://github.com/NWACus/web/issues/988 was very light in terms of description here - I never expanded further on what I meant. I probably created quickly. There was a small mention of this approach in https://github.com/NWACus/web/issues/533#:~:text=v2-,Use%20AFP%20and%20NAC%20APIs%20to%20keep%20navigation%20in%20sync,-v3

Anyways - the `hasReadOnly___` props feel counterintuitive to me. Here's why:

We made a technical decision (which I regret not documenting in a design decision - just looked to see if this was documented 🤦 ) that we would build the navigation so that it's extremely flexible under the hood but that we limit how much control center admins have over it using roles. So we currently have the center admin role without edit permission on navigations. 

**For events and blog**: What if we needed to point the events link at a center's blog or a center's blog with specific filters applied (like they wanted to use blog posts for events for some reason). Silly example but I don't see a strong argument for making this read only. 

**For AFP/widget-based built-in pages (that are dynamic based on configuration in the AFP)**: 
- I think we should be able to change the label of these links for centers if they ask for it
- If we're going to make these read only, we should also be building a system to keep them in sync with the AFP. i.e. if a center updates their zones in the AFP, we should be updated the built-in page references in that center's navigations document. A sync system would also eliminate the need for the one-time script. 


What's your plan for keeping the built-in page references in sync with the AFP?

rchlfryn:
@busbyk Thanks for the reminder/context on the AFP sync direction. That makes sense as the end state, and I agree a sync system would be the right long-term answer for forecast zones and AFP driven pages.

For this PR I'd like to keep the scope focused on seeding built-in page references and getting the nav pointed at real CMS links. Building an AFP → navigation sync is still needed, but I think it is a separate issue. I'd rather do the sync as a follow-up so we can land the user-visible improvements and finish provisioning for now. I made #1038 and tried to capture everything you mentioned, but feel free to edit it as needed.

**On the read-only props:**  What I was actually going for was a visual representation in the admin UI that these fields are AFP-driven. Since all avy centers will have built-in pages, I think it makes sense to give admins a visual to the other nav rather than it being blank. `readOnly` was a bad name for the AFP driven pages since it implied admins couldn't edit the field, which isn't what I wanted to communicate. It conflated "this is autofilled from AFP" with "role-gated editing." Two different concerns.

I've stripped the read-only behavior. Update access is still gated to super-admins so we will know not to update the navigations after provisioning until #1038 is done.

src/collections/Navigations/fields/itemsField.ts
rchlfryn: Reverted. Just kept the `Condition` type improvement

src/collections/Navigations/fields/topLevelNavTab.ts
rchlfryn: I think this was in reference to `hasReadOnly` props, but I have updated this. See main comment for the thread.

src/components/Header/utils.ts
rchlfryn: I am also skeptical about the fallbacks, but I thought it was the best stop gap for downtime in between running the one time script, and when we re-provision/ manually update the navigation to point to built-in pages. I had ever intention of removing them after we ran the script.

I will take a look at #1025.

src/scripts/migrate-nav-builtin-pages.ts
rchlfryn: Oh very interesting thought. I was thinking we would just manually add the built-in pages, rather than re-provision. Let me look into this a bit further.

src/collections/BuiltInPages/index.ts
rchlfryn: The intent for this was to stop a user from deleting a built-in page that was in use in the navigation. I guess we don't care about this for pages, so do we care about it here?

rchlfryn: Since we are not doing anything to stop admins from deleting page/posts in use in the nav, I added this as a follow up item https://github.com/NWACus/web/issues/1042

src/scripts/migrate-nav-builtin-pages.ts
rchlfryn: 1025 was super helpful. I added a migration to replace this script.

src/components/Header/utils.ts
rchlfryn: I was able to remove the fallbacks by adding a migration to populate the nav

src/components/Header/MobileNav.client.tsx
```diff
@@ -109,11 +109,19 @@ export const MobileNav = ({
               )}
             </Link>
           )}
-          {donateNavItem && (
-            <RenderNavLink link={donateNavItem.link} onClick={() => setMobileNavOpen(false)}>
-              <Button variant="callout">{donateNavItem.label}</Button>
-            </RenderNavLink>
-          )}
+          {buttonNavItems.map((navItem) => {
```
busbyk: Wouldn't this result in any top level nav item that's rendered as a button showing up out of the default tabs order? They would all appear at the end of the list instead of in the order of the tabs that we've defined, right?

src/components/Header/MobileNavItem.tsx
```diff
@@ -2,25 +2,40 @@
...
 }
 
 export const MobileNavItem = ({
   label,
   navItem,
+  displayMode,
   setMobileNavOpen,
   className,
 }: MobileNavItemProps) => {
+  if (displayMode === 'button' && navItem.link) {
```
busbyk: Couldn't we just handle rendering as a button in <RenderNavLink> rather than having this if block here? Pass displayMode into RenderNavLink.

src/components/Header/DesktopNav.client.tsx
```diff
@@ -33,6 +27,18 @@ export const DesktopNav = ({
 
           if (!label) return null
 
+          if (navItem.displayMode === 'button' && navItem.link) {
```
busbyk: Couldn't we just handle rendering as a button in rather than having this if block here? Pass displayMode into RenderNavLink.

src/components/Header/utils.ts
```diff
@@ -39,392 +40,110 @@ export type NavItem = {
...
   const topLevelNavItems: TopLevelNavItem[] = [
-    ...(avalancheCenterPlatforms.forecasts ? [forecastsNavItem] : []),
-    ...topLevelNavItem({
-      tab: navigation.weather,
-      label: 'Weather',
-    }),
-    ...(avalancheCenterPlatforms.obs ? [observationsNavItem] : []),
+    ...(avalancheCenterPlatforms.forecasts && forecastsNavItem ? [forecastsNavItem] : []),
+    ...topLevelNavItem({ tab: navigation.weather, label: 'Weather' }),
+    ...(avalancheCenterPlatforms.obs && observationsNavItem ? [observationsNavItem] : []),
```
busbyk: Why can't we pass forecasts and obs into `topLevelNavItem()` instead of doing custom processing above this? We can still respect the `avalancheCenterPlatforms` config and pass into `topLevelNavItem()` can't we?

src/endpoints/seed/index.ts
```diff
@@ -951,22 +952,41 @@ export const seed = async ({
       ),
     )
 
+    const forecastZonesByTenant: Record<
+      string,
+      Awaited<ReturnType<typeof getActiveForecastZones>>
+    > = {}
+    for (const tenant of Object.values(tenants)) {
+      forecastZonesByTenant[tenant.slug] = await getActiveForecastZones(tenant.slug)
```
busbyk: This means we need internet connectivity to run the seed script. I don't love that. I think it would be reasonable and better to hardcode this.

src/collections/Navigations/index.tsx
```diff
@@ -37,60 +38,66 @@ export const Navigations: CollectionConfig = {
...
+            'Autofilled from your forecast zones in the AFP. Edits here may be overwritten when zones change in the AFP.',
           hasEnabledToggle: false,
         }),
-        topLevelNavTab({
+        navTab({
           name: 'observations',
-          description: 'This nav dropdown is autofilled with the default observations links.',
-          hasConfigurableNavItems: false,
+          description:
+            'Autofilled with default observation links (Recent Observations, Submit Observation). Edits here may be overwritten.',
```
busbyk: As of this PR, this isn't true. If we're going to do this progressively, I think it makes sense to remove this messaging until the PR that may or may not introduce the scenario where these get overwritten by a job.

src/collections/Navigations/index.tsx
```diff
@@ -37,60 +38,66 @@ export const Navigations: CollectionConfig = {
...
           hasEnabledToggle: false,
         }),
-        topLevelNavTab({
+        navTab({
           name: 'weather',
-          description: 'This nav dropdown will also include your weather stations.',
+          description:
+            'Your weather stations map is automatically included in this dropdown. Add any additional weather-related links below.',
```
busbyk: Do we need this description? They would see the link to the weather stations map here if provisioning has been done.

src/collections/Navigations/index.tsx
```diff
@@ -37,60 +38,66 @@ export const Navigations: CollectionConfig = {
...
         }),
-        topLevelNavTab({ name: 'education' }),
-        topLevelNavTab({ name: 'accidents' }),
-        topLevelNavTab({
+        navTab({
           name: 'blog',
           description:
-            'This nav item navigates to your blog landing page and does not have any dropdown items.',
-          hasConfigurableNavItems: false,
+            'Points to your blog landing page (posts are listed automatically on that page).',
```
busbyk: Do we need the "(posts are listed automatically on that page)"?

src/collections/Navigations/index.tsx
```diff
@@ -37,60 +38,66 @@ export const Navigations: CollectionConfig = {
...
           enabledToggleDescription:
             'If hidden from the nav, the blog landing page will still be accessible to visitors for filtered blog lists.',
         }),
-        topLevelNavTab({
+        navTab({
           name: 'events',
           description:
-            'This nav item navigates to your events landing page and does not have any dropdown items.',
-          hasConfigurableNavItems: false,
+            'Points to your events landing page (upcoming events are listed automatically on that page).',
```
busbyk: Do we need the "(upcoming events are listed automatically on that page)"?

src/collections/Navigations/index.tsx
```diff
@@ -37,60 +38,66 @@ export const Navigations: CollectionConfig = {
...
-                  defaultValue: true,
-                  name: 'enabled',
-                },
-              ],
-            },
-            navLink,
-          ],
-        },
+          description:
+            'Renders as a call-to-action button at the end of the navigation. Usually points to your donation page.',
```
busbyk: Well, technically, they could change this to just be a link. And as I've gotten to the end of this list and left comments on the descriptions, I wonder if we should have these descriptions at all given that the type can be changed. Descriptions that prescribe a specific type (dropdown, button, link) don't really make sense, right?

src/collections/Navigations/fields/navTab.ts
```diff
@@ -0,0 +1,107 @@
...
+}): Tab => {
+  const displayModeField: Field = {
+    name: 'displayMode',
+    type: 'radio',
+    defaultValue: defaultMode,
+    options: MODE_OPTIONS,
+    admin: {
+      layout: 'vertical',
+    },
+  }
```
busbyk, line 52: I think it would look a lot nicer if we used a custom component with a ToggleGroup like the ContentBlock's [ColumnLayoutPicker](https://github.com/NWACus/web/blob/nav-updates/src/components/ColumnLayoutPicker/index.tsx#L67)

busbyk (changes requested):

src/collections/Navigations/index.tsx
rchlfryn: It's not necessary. I will clean up all the descriptions. Done in [bf07961](https://github.com/NWACus/web/pull/1002/commits/bf079614faa579b2b955224579723e2582614032)

src/collections/Navigations/index.tsx
rchlfryn: Nope - I will clean this up. Done in [bf07961](https://github.com/NWACus/web/pull/1002/commits/bf079614faa579b2b955224579723e2582614032)

src/collections/Navigations/index.tsx
rchlfryn: Nope - removing

src/collections/Navigations/index.tsx
rchlfryn: Good point - I think it made sense before `displayMode` was added. Done in [bf07961](https://github.com/NWACus/web/pull/1002/commits/bf079614faa579b2b955224579723e2582614032)

src/collections/Navigations/fields/navTab.ts
rchlfryn: Agreed - I looked for Payload to have a toggle group, but no reason we cannot make our own. Done in [649f391](https://github.com/NWACus/web/pull/1002/commits/649f391f8dbf3494eea914722330233ef284add8)

src/collections/Navigations/index.tsx
rchlfryn: Done in [bf07961](https://github.com/NWACus/web/pull/1002/commits/bf079614faa579b2b955224579723e2582614032)

src/endpoints/seed/index.ts
rchlfryn: Done in [6fc1ed2](https://github.com/NWACus/web/pull/1002/commits/6fc1ed28e606d22432eadb5eed56ff47a272b5f7)

src/components/Header/MobileNav.client.tsx
rchlfryn: Good catch - cleaned up in [c42e3b2](https://github.com/NWACus/web/pull/1002/commits/c42e3b2de88b0cef4a5548feacd9112a03637f87)

src/components/Header/MobileNavItem.tsx
rchlfryn: Updated in [c42e3b2](https://github.com/NWACus/web/pull/1002/commits/c42e3b2de88b0cef4a5548feacd9112a03637f87)

src/components/Header/DesktopNav.client.tsx
rchlfryn: Updated in [c42e3b2](https://github.com/NWACus/web/pull/1002/commits/c42e3b2de88b0cef4a5548feacd9112a03637f87)

src/components/Header/utils.ts
rchlfryn: Yes we can - good improvement [c42e3b2](https://github.com/NWACus/web/pull/1002/commits/c42e3b2de88b0cef4a5548feacd9112a03637f87)

rchlfryn:
Github is struggling today, but ready for re-review

src/components/Header/utils.ts
```diff
@@ -125,6 +77,17 @@ export const getTopLevelNavItems = async ({
...
+  if (center === 'sac') {
+    const observations = topLevelNavItems.find((item) => item.label === 'Observations')
+    if (observations?.items) {
+      observations.items.push({
+        id: 'archive',
+        link: { type: 'internal', label: 'Observations Archive', url: '/observations-archive' },
+      })
+    }
+  }
+
```
busbyk: With our ability to control the items in the observations tab now we should be able to remove this since they can add this manually. They'd just add a built-in page pointing to this URL. Right?

busbyk (approved):
Looks good, just one comment.

src/components/Header/utils.ts
rchlfryn: Great call - I even looked at this and was like "no we still need this". Not sure why I had that thought.

PR #1011, opened 2026-03-26
Auto build provisioning from AFP & track tenant provisioning lifecycle

### LINKED ISSUE #999: Auto-detect single vs multi-zone center from AFP during provisioning
## Summary

When provisioning a new avalanche center, we should automatically check the AFP (Avalanche Forecast Program) data to determine if the center is a single-zone or multi-zone operation and provision accordingly.

Currently, provisioning creates the same 7 built-in pages regardless of zone count. Zone-aware navigation is handled dynamically at runtime (`src/components/Header/utils.ts`), but the built-in pages created during provisioning should also differ.

## Areas to update

- **`src/collections/Tenants/endpoints/provisionTenant.ts`** — Query AFP via `getActiveForecastZones()` from `src/services/nac/nac.ts` to determine zone count, then create the appropriate built-in pages:
  - **Single zone**: Create `/forecasts/avalanche/{zone}` page
  - **Multi-zone**: Create `/forecasts/avalanche` (all forecasts) + one page per zone at `/forecasts/avalanche/{zone}`
- **`docs/onboarding.md`** — Already documents the single vs multi-zone difference (lines 22-34), update if provisioning handles this automatically
- **Filter built-in pages using DVAC navigation** — `extractNavReferences` (in `provisionTenant.ts`) already extracts `builtInPageUrls` from the template navigation. Use this to filter `BUILT_IN_PAGES` so only navigation-referenced built-in pages are created, matching how pages are already filtered. See TODO comments in `provisionTenant.ts` and `onboardingActions.ts` referencing #1002.

## Context

From the onboarding provisioning PR — zone configuration was identified as something that could be automated rather than manually configured. The NAC API already provides zone data via `getActiveForecastZones(centerSlug)`.

## Acceptance Criteria

- [ ] Query AFP data during provisioning to determine zone count
- [x] Create zone-specific built-in pages based on single vs multi-zone
- [ ] Filter built-in pages to those referenced in DVAC's navigation (use `navBuiltInPageUrls` from `extractNavReferences`)
- [ ] Handle cases where AFP data is unavailable (fall back to manual configuration)
- [x] Update onboarding docs to reflect automation

### LINKED ISSUE #1000: Define completion criteria for onboarding checklist content steps
## Summary

The onboarding checklist currently checks completion by querying whether expected content exists (e.g., pages with specific slugs). This means if a tenant admin deletes or renames a provisioned page, the checklist reverts to "Incomplete" even though provisioning originally succeeded.

**Example**: If a tenant renames their donate page slug (as Sierra did), the checklist shows the page as missing.

See discussion: https://github.com/NWACus/web/pull/974#discussion_r2943535563

## Options to consider

1. **Track provisioning history** — Store the IDs of pages created during provisioning, so completion reflects "was successfully created" rather than "currently exists"
2. **Store a completion flag** — Add a field on the Tenants collection (e.g., `provisioningComplete: true`) that gets set once provisioning succeeds, independent of current content state
3. **Keep current behavior** — Checklist reflects the live state, which is useful for detecting accidental deletions but noisy for intentional changes

## Acceptance Criteria

- [ ] Decide on the desired definition of "completed" for each content step
- [ ] Implement the chosen approach
- [ ] Update tests to reflect the new behavior

rchlfryn:
## Description

Improve tenant provisioning to auto-detect forecast zones, conditionally include Mountain Weather, and track the provisioning lifecycle directly on the tenant record so the admin UI always has a single source of truth.

## Related Issues

Fixes #999
Fixes #1000

### Provisioning updates
- **Static page lists as source of truth** — `BUILT_IN_PAGES` and `PAGES_TO_PROVISION` are hardcoded in `provisionTenant.ts` rather than sourced from a template tenant. New built-in/blank pages get added here explicitly, avoiding drift from whatever DVAC happens to have configured.
- **Provisioning lifecycle tracked on tenant** — New `provisioning` group (`status`, `lastRunAt`, `failed`) on the tenants collection. Status moves `not_started → in_progress → complete | partial | manual`. Writing `in_progress` upfront prevents reloads mid-run from triggering a second provision. `failed` captures per-step error messages (`websiteSettings`, `homePage`, `navigation`, `pages: { [title]: error }`) so the UI can show exactly which pages didn't make it.
- **`manual` state for theme/OG items** — `checkProvisioningStatusAction` flips `complete` ↔ `manual` in the DB based on whether the tenant slug appears in `colors.css` and `centerColorMap`. Surfaces the "needs manual code edit" state distinctly from a provisioning failure.
- **Crashed-run recovery** — if a tenant has been `in_progress` for longer than 5 minutes, `checkProvisioningStatusAction` surfaces it as `partial` with a `timedOut` failure so the checklist shows a Rerun button instead of an infinite spinner.
- **Per-tenant DB lock** — `provision()` acquires the `in_progress` state via a conditional update that only succeeds if the tenant isn't already running (or the active run is past the 5-minute stale cutoff). Prevents two admins clicking Rerun simultaneously from racing through provisioning.
- **StrictMode-safe auto-provision** — the mount-time auto-provision is keyed on `tenantId` via a `useRef` so React StrictMode's double-mount (or any remount) doesn't fire two runs.
- **Super-admin auth on server actions** — `checkProvisioningStatusAction`, `runProvisionAction`, and `markProvisioningCompleteAction` all require super-admin. Previously any authenticated Payload admin could invoke them on any tenant.
- **Migration schema-drift fix** — `20260418_120100_backfill_nav_builtin_pages` now reads tenants via raw SQL (`SELECT id, slug FROM tenants`) instead of `payload.find`. Avoids SQLite's double-quoted-identifier gotcha where Drizzle's SELECT of a column added by a later migration silently returns the column name as a string literal and chokes the JSON parser.

### AFP dependent 
- **Zone-aware forecast provisioning** — `provisionTenant.ts` calls `getActiveForecastZones()` during provisioning. Single-zone centers get one "Avalanche Forecast" page; multi-zone centers get "All Forecasts" + per-zone pages sorted by rank. Falls back to a default "All Forecasts" page if AFP is unavailable.
- **Mountain Weather from NAC** — Mountain Weather (`/weather/forecast`) is conditionally included per-tenant based on `platforms.weather` from NAC.
- **Zone-aware navigation** — Provisioning populates the forecasts tab (single zone: direct link; multi-zone: "All Forecasts" + per-zone items) and the observations tab with built-in page links. Also fixes missing `blog` and `events` tab data that was causing nav creation to silently fail.
- **Shared `resolveBuiltInPages`** — The AFP zone query + Mountain Weather resolution is factored out so both `provisionTenant.ts` and the onboarding status check use the same logic.

### UI updates
- **Onboarding checklist UI** — Auto-provisions brand-new tenants on mount, shows placeholder items with spinners during `in_progress`, collapses to a "Complete" / "Complete manual actions" / "Failed" header afterward with per-step failure detail and a Rerun button. An inline link to the tenant's Website Settings sets the tenant cookie via `useTenantSelection` before navigating, so the admin doesn't redirect back to the previously selected tenant. Create-view is handled (no tenant yet → no spinner, placeholder items only).
- **List view pill** — New `OnboardingStatusCell` renders a Complete / Manual actions / Partial / In progress / Not started pill on the tenants list so provisioning state is visible without opening each tenant.

## How to test

1. `pnpm migrate` to apply the new provisioning columns, then `pnpm seed`
2. At `/admin/collections/tenants`, verify the new list-view pill shows "Complete" for all seeded tenants
3. Create a new tenant
    - Checklist shows placeholder items with spinners while provisioning runs
    - Header flips to "Complete" (or "Failed" with per-step detail) when done
    - For multi-zone `btac` (4 zones): "All Forecasts" + 4 zone pages are created as built-in pages
    - For single-zone `msac`: a single "Avalanche Forecast" built-in page is created
4. Verify Mountain Weather is only created for centers with weather forecasts in NAC
5. Reload the tenant page mid-provision — the UI shows "in progress" state and does **not** retrigger a second provisioning run
6. Simulate a partial failure (e.g. direct DB write to `provisioning_status='partial'` with a `provisioning_failed` JSON payload) and verify the checklist lists the failed pages and shows a Rerun button
7. `pnpm test`

## Screenshots / Demo video

| &nbsp; | &nbsp; |
|--------|--------|
| Screenshot 2026-04-22 at 13 12 52 | Screenshot 2026-04-22 at 09 57 19 | 
| Screenshot 2026-04-22 at 13 09 04 | Screenshot 2026-04-22 at 13 12 32 |

## Migration Explanation

Adds three columns to `tenants`:

- `provisioning_status` (text, NOT NULL, default `'not_started'`)
- `provisioning_last_run_at` (text, nullable)
- `provisioning_failed` (text, JSON)

Existing rows are conditionally backfilled to `complete` only if they have matching `settings`, `home_pages`, and `navigations` rows (with `lastRunAt = created_at`). Half-provisioned tenants stay at `not_started` so the checklist shows Rerun.

- The onboarding status check calls AFP on every checklist load — could cache the zone count if this becomes a performance concern (likely not).

### DIFF
--- a/docs/onboarding.md
+++ b/docs/onboarding.md
@@ -11,29 +11,25 @@ Provisioning is idempotent and can be rerun safely.
 | Step | Details |
 |------|---------|
 | Website Settings | Created with placeholder brand assets (logo, icon, banner). Replace with real assets via the checklist link. |
-| Built-in pages | Creates standard pages per the table below. |
+| Forecast pages | Queries AFP via `getActiveForecastZones()` to auto-detect single vs multi-zone. Creates zone-specific built-in pages (see table below). Falls back to a default "All Forecasts" page if AFP is unavailable. |
+| Default built-in pages | Creates non-forecast built-in pages sourced from the template (DVAC) navigation (see table below). |
 | Template pages | Copies all published pages from the template tenant (DVAC). Pages whose blocks all reference tenant-scoped data (teams, sponsors, events, forms) are copied as empty drafts. Demo pages (`blocks`, `lexical-blocks`) are skipped. Static blog/event list blocks are converted to dynamic mode. |
 | Home page | Creates a home page with welcome content and quick links to About Us and Donate. |
-| Navigation | Creates navigation menus linked to all copied pages and built-in pages. |
+| Navigation | Creates navigation menus linked to all copied pages and built-in pages. Forecasts tab is zone-aware (single zone: direct link; multi-zone: "All Forecasts" + per-zone items). |
 | Edge Config | The `updateEdgeConfigAfterChange` hook automatically adds the tenant to Vercel Edge Config. |
 
 #### Built-In Pages
 
-\* If a center has a single forecast zone, it gets an "Avalanche Forecast" page pointing to that zone. Multi-zone centers get an "All Forecasts" page plus individual zone pages.
+Forecast pages are determined by AFP zone data\*. Non-forecast pages are sourced from the template tenant's (DVAC) navigation — adding or removing a built-in page in DVAC's nav automatically changes what new tenants get.
 
-| Title | URL | For AC with single or multi zone* |
-|-------|-----|-----------------------------------|
-| All Forecasts | `/forecasts/avalanche` | multi |
-| _ZONE NAME_ | `/forecasts/avalanche/ZONE` | multi |
-| Avalanche Forecast | `/forecasts/avalanche/ZONE` | single |
-| Mountain Weather** | `/weather/forecast` | both |
-| Weather Stations | `/weather/stations/map` | both |
-| Recent Observations | `/observations` | both |
-| Submit Observations | `/observations/submit` | both |
-| Blog | `/blog` | both |
-| Events | `/events` | both |
+\* If a center has a single forecast zone, it gets an "Avalanche Forecast" page pointing to that zone. Multi-zone centers get an "All Forecasts" page plus individual zone pages. If AFP is unavailable, a default "All Forecasts" page is created.
 
-\*\* Mountain Weather is only available for centers that have a weather forecast configured.
+| Title | URL | Source |
+|-------|-----|--------|
+| All Forecasts | `/forecasts/avalanche` | AFP (multi-zone) |
+| _ZONE NAME_ | `/forecasts/avalanche/ZONE` | AFP (multi-zone) |
+| Avalanche Forecast | `/forecasts/avalanche/ZONE` | AFP (single-zone) |
+| _Non-forecast pages_ | _varies_ | DVAC navigation |
 
 ## Manual steps
 
--- a/docs/onboarding.md
+++ b/docs/onboarding.md
@@ -12,7 +12,7 @@ Provisioning is idempotent and can be rerun safely.
 |------|---------|
 | Website Settings | Created with placeholder brand assets (logo, icon, banner). Replace with real assets via the checklist link. |
 | Forecast pages | Queries AFP via `getActiveForecastZones()` to auto-detect single vs multi-zone. Creates zone-specific built-in pages (see table below). Falls back to a default "All Forecasts" page if AFP is unavailable. |
-| Default built-in pages | Creates non-forecast built-in pages sourced from the template (DVAC) navigation (see table below). |
+| Default built-in pages | Creates non-forecast built-in pages sourced from the template (DVAC) navigation (see table below). Mountain Weather is only included if the center has a weather forecast configured in NAC (`platforms.weather`). |
 | Template pages | Copies all published pages from the template tenant (DVAC). Pages whose blocks all reference tenant-scoped data (teams, sponsors, events, forms) are copied as empty drafts. Demo pages (`blocks`, `lexical-blocks`) are skipped. Static blog/event list blocks are converted to dynamic mode. |
 | Home page | Creates a home page with welcome content and quick links to About Us and Donate. |
 | Navigation | Creates navigation menus linked to all copied pages and built-in pages. Forecasts tab is zone-aware (single zone: direct link; multi-zone: "All Forecasts" + per-zone items). |
@@ -29,6 +29,7 @@ Forecast pages are determined by AFP zone data\*. Non-forecast pages are sourced
 | All Forecasts | `/forecasts/avalanche` | AFP (multi-zone) |
 | _ZONE NAME_ | `/forecasts/avalanche/ZONE` | AFP (multi-zone) |
 | Avalanche Forecast | `/forecasts/avalanche/ZONE` | AFP (single-zone) |
+| Mountain Weather | `/weather/forecast` | NAC `platforms.weather` |
 | _Non-forecast pages_ | _varies_ | DVAC navigation |
 
 ## Manual steps
--- a/docs/onboarding.md
+++ b/docs/onboarding.md
@@ -12,15 +12,15 @@ Provisioning is idempotent and can be rerun safely.
 |------|---------|
 | Website Settings | Created with placeholder brand assets (logo, icon, banner). Replace with real assets via the checklist link. |
 | Forecast pages | Queries AFP via `getActiveForecastZones()` to auto-detect single vs multi-zone. Creates zone-specific built-in pages (see table below). Falls back to a default "All Forecasts" page if AFP is unavailable. |
-| Default built-in pages | Creates non-forecast built-in pages sourced from the template (DVAC) navigation (see table below). Mountain Weather is only included if the center has a weather forecast configured in NAC (`platforms.weather`). |
-| Template pages | Copies all published pages from the template tenant (DVAC). Pages whose blocks all reference tenant-scoped data (teams, sponsors, events, forms) are copied as empty drafts. Demo pages (`blocks`, `lexical-blocks`) are skipped. Static blog/event list blocks are converted to dynamic mode. |
+| Default built-in pages | Creates non-forecast built-in pages from the static `BUILT_IN_PAGES` list in `provisionTenant.ts` (see table below). Mountain Weather is only included if the center has a weather forecast configured in NAC (`platforms.weather`). |
+| Blank pages | Creates empty pages for every slug in the static `PAGES_TO_PROVISION` list in `provisionTenant.ts`. Admins are expected to fill in the content after provisioning. |
 | Home page | Creates a home page with welcome content and quick links to About Us and Donate. |
-| Navigation | Creates navigation menus linked to all copied pages and built-in pages. Forecasts tab is zone-aware (single zone: direct link; multi-zone: "All Forecasts" + per-zone items). |
+| Navigation | Creates navigation menus linked to all provisioned pages and built-in pages. Forecasts tab is zone-aware (single zone: single-item dropdown; multi-zone: "All Forecasts" + a "Zones" accordion with per-zone items). |
 | Edge Config | The `updateEdgeConfigAfterChange` hook automatically adds the tenant to Vercel Edge Config. |
 
 #### Built-In Pages
 
-Forecast pages are determined by AFP zone data\*. Non-forecast pages are sourced from the template tenant's (DVAC) navigation — adding or removing a built-in page in DVAC's nav automatically changes what new tenants get.
+Forecast pages are determined by AFP zone data\*. Non-forecast pages come from the static `BUILT_IN_PAGES` list in `src/collections/Tenants/endpoints/provisionTenant.ts` — edit that list to change what new tenants get.
 
 \* If a center has a single forecast zone, it gets an "Avalanche Forecast" page pointing to that zone. Multi-zone centers get an "All Forecasts" page plus individual zone pages. If AFP is unavailable, a default "All Forecasts" page is created.
 
@@ -30,7 +30,7 @@ Forecast pages are determined by AFP zone data\*. Non-forecast pages are sourced
 | _ZONE NAME_ | `/forecasts/avalanche/ZONE` | AFP (multi-zone) |
 | Avalanche Forecast | `/forecasts/avalanche/ZONE` | AFP (single-zone) |
 | Mountain Weather | `/weather/forecast` | NAC `platforms.weather` |
-| _Non-forecast pages_ | _varies_ | DVAC navigation |
+| _Non-forecast pages_ | _varies_ | `BUILT_IN_PAGES` constant |
 
 ## Manual steps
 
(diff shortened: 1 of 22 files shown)


busbyk:
@rchlfryn thinking I'll wait to review this until some discussion on #1002

src/collections/Tenants/index.ts
```diff
@@ -75,6 +75,61 @@ export const Tenants: CollectionConfig = {
...
+      fields: [
+        {
+          name: 'status',
+          type: 'select',
+          defaultValue: 'not_started',
+          required: true,
+          options: [
+            { label: 'Not started', value: 'not_started' },
+            { label: 'In progress', value: 'in_progress' },
+            { label: 'Complete', value: 'complete' },
```
stevekuznetsov, line 94: I couldn't quite tell from the PR description or a quick review if you expected this to be a terminal phase. If it is, I would recommend instead to never have terminal phases, as they lead to messy broken states when things go awry. Centers can and do change their zones, including re-naming them, so storing the last timestamp of successful processing and re-running the logic on some cadence (along with allowing the admin to force it) would make for a more resilient design.

rchlfryn: This PR is still a WIP since it is in draft mode, but I think you are referring to this issue https://github.com/NWACus/web/issues/1038?

stevekuznetsov: Not sure I fully understand what you mean? The description of this PR talks about bringing zone info from the AFP into our system to populate pages - does this PR attempt to implement #1038 ?

I'm just commenting that any such system would be more resilient without terminal phases in the state diagram.

rchlfryn: > does this PR attempt to implement https://github.com/NWACus/web/issues/1038 ?

No what this PR does is looks at the zones at the AFP and adds them as built in pages for new tenants.

I think we're conflating two things here. This is the provisioning/onboarding status. The question it answers is "did we successfully provision this tenant?", which is a one-shot with a permanent yes/no answer. `complete` is correctly terminal for that question.

What's not here is automatic periodic re-sync of AFP zone changes (rename, rank, add/remove) — that's #1038, which is a bigger scope with its own design decisions (trigger mechanism: scheduled vs. webhook vs. on-read; conflict resolution with admin-edited labels; orphan handling for removed zones). I'd rather scope that PR independently than fold it in here.

stevekuznetsov: I don't think a one-shot process for provisioning is architecturally correct, and shipping that without having a way to capture updates that centers do in the AFP leaves a feature gap. Solving the "reconcile state continually" is a superset of "reconcile it once on provision," so I would recommend doing the work once to solve the problem fully instead of doing the work twice. Can you provide some clarity on what point-in-time benefits we get today by shipping a solution with downsides now and waiting to solve the full problem until later? Are we in dire need to provision many tenants in the extreme short-term such that we don't think we could write a reconciler in the time that they need to be done?

docs/onboarding.md
```diff
@@ -8,32 +8,33 @@ Create the tenant in the admin panel → provisioning runs automatically via `pr
...
+Forecast pages are determined by AFP zone data\*. Non-forecast pages come from the static `BUILT_IN_PAGES` list in `src/collections/Tenants/endpoints/provisionTenant.ts` — edit that list to change what new tenants get.
+
+\* If a center has a single forecast zone, it gets an "Avalanche Forecast" page pointing to that zone. Multi-zone centers get an "All Forecasts" page plus individual zone pages. If AFP is unavailable, a default "All Forecasts" page is created.
+
+| Title | URL | Source |
+|-------|-----|--------|
+| All Forecasts | `/forecasts/avalanche` | AFP (multi-zone) |
+| _ZONE NAME_ | `/forecasts/avalanche/ZONE` | AFP (multi-zone) |
+| Avalanche Forecast | `/forecasts/avalanche/ZONE` | AFP (single-zone) |
+| Mountain Weather | `/weather/forecast` | NAC `platforms.weather` |
```
busbyk, line 36: I just realized that previously this wasn't in anyone's navigations 😳 -- Sierra has weather enabled and you can visit the direct URL https://www.sierraavalanchecenter.org/weather/forecast#/weather but it's not in their nav anywhere. Woops.

src/collections/Tenants/components/onboardingActions.ts
```diff
@@ -1,74 +1,57 @@
...
+import { isRecord } from '@/utilities/isRecord'
 import config from '@payload-config'
 import fs from 'fs/promises'
+import { headers } from 'next/headers'
 import path from 'path'
-import { getPayload } from 'payload'
+import { getPayload, type Payload, type PayloadRequest } from 'payload'
+
+// Server actions are POST endpoints reachable by anyone with the action ID —
+// gate every action behind a super-admin check before touching tenant data.
```
busbyk: I agree with this but do we need this comment? I feel like this guidance would make more sense in the coding-guide.md doc

src/collections/Tenants/components/OnboardingChecklist.tsx
```diff
@@ -135,105 +127,167 @@ export function OnboardingChecklist() {
...
+              <div className="flex items-center gap-3">
+                {status.status === 'complete' ? (
+                  <>
+                    <CheckCircle2 size={20} className="shrink-0 text-success" />
+                    <span>Complete</span>
+                  </>
+                ) : status.status === 'manual' ? (
+                  <>
+                    <AlertTriangle size={20} className="shrink-0 text-warning" />
+                    <span>Complete manual actions</span>
```
busbyk: This makes me think I need to do something and I'm tempted to click re-run haha 

Image

This makes me feel like we should skip the manual status and just have a concept of complete, and then check the manual code changes programmatically each time, since those can't drift like the pages can. 

But I guess we'd still wanna keep the indication that manual steps are needed in the onboarding status cell component. I'm a little torn on this one. If it makes sense to keep the manual status, I'd at least want to change the UI display here to indicate that the automated actions were completed successfully, but that manual actions are still needed.

src/collections/Tenants/components/onboardingActions.ts
rchlfryn: Moved in [c7c0dd1](https://github.com/NWACus/web/pull/1011/commits/c7c0dd170b97e375ac570de612cd37e938a194f3)

src/collections/Tenants/components/OnboardingChecklist.tsx
rchlfryn: I agree with you this UI needed some work. I stopped being lazy and moved the rerun button to be out of the box, and updated the manual actions remaining. What do you think?

Screenshot 2026-04-27 at 15 05 02
Screenshot 2026-04-27 at 15 05 14

src/collections/Tenants/index.ts
rchlfryn: The reconciler is a superset of AFP-zone provisioning, you're right about that. But tenant provisioning establishes initial state across ~10 collections (built-in pages, website settings, home page, blog/events landing pages, sponsors, contact form, navigation). An AFP→nav reconciler operates on a small subset (forecast zones, built-in pages, nav references). The bulk of the provisioning code (`BUILT_IN_PAGES`, `PAGES_TO_PROVISION`, the upsert sequencing) isn't made obsolete by a reconciler because most of it isn't AFP-driven. The reconciler adds to this work. It doesn't replace it.

A reconciler forces design decisions that first-time provisioning avoids entirely: trigger mechanism, conflict resolution, orphan handling, drift detection, sync timestamps, failure observability. None of those need answers for one-shot provisioning, and I'd rather work through them deliberately on #1038, have a proper planning meeting and chat about it than rush them here.

I don't think we are doing the work twice. I think this is iterating. There's a small overlap. This PR inserts AFP-resolved zones into built-in pages and nav as one-shot writes. A future reconciler will need a diff/upsert path over the same zones. The shared logic (`resolveBuiltInPages`) is already factored out. I'd rather pay that cost than block provisioning on a reconciler design with its own non-trivial decisions, especially since most of provisioning is non-AFP work the reconciler doesn't touch.

There is no rush here. All of this is being merged into a feature branch to wait for the #1038 to go along with it.

src/collections/Tenants/components/OnboardingChecklist.tsx
busbyk: I like it! Thanks for updating. I think this is great.

src/migrations/20260421_181446_add_tenant_provisioning_field.ts
```diff
@@ -0,0 +1,30 @@
...
+    sql`ALTER TABLE \`tenants\` ADD \`provisioning_status\` text DEFAULT 'not_started' NOT NULL;`,
+  )
+  // lastRunAt is nullable — brand-new tenants have no run to record yet.
+  await db.run(sql`ALTER TABLE \`tenants\` ADD \`provisioning_last_run_at\` text;`)
+  await db.run(sql`ALTER TABLE \`tenants\` ADD \`provisioning_failed\` text;`)
+
+  // Only backfill tenants that have evidence of a successful provision —
+  // a settings row, a home page, AND a navigation. Half-provisioned tenants
+  // stay at the default 'not_started' so the checklist surfaces a Rerun
+  // affordance instead of claiming they're done.
```
busbyk, line 15: Nice. 

Does provisioning_failed need to be set explicitly? I assume no, it will default to NULL but just making sure.

src/collections/Tenants/components/OnboardingStatusCell.tsx
```diff
@@ -1,38 +1,28 @@
...
+  if (status === 'partial')
+    return (
+      <Pill size="small" pillStyle="warning">
+        Partial
+      </Pill>
+    )
+  if (status === 'manual')
+    return (
+      <Pill size="small" pillStyle="warning">
+        Manual actions
```
busbyk, line 23: ⛏️ Like the OnboardingChecklist, do we want to make it clear that it's completed but manual steps still required?

busbyk (approved):

src/collections/Tenants/components/OnboardingStatusCell.tsx
rchlfryn: I was back and forth on this. The label seemed long with `Complete - Manual actions` so let's leave it for now but I will likely change in when addressing 1038

src/migrations/20260421_181446_add_tenant_provisioning_field.ts
rchlfryn: Correct — `provisioning_failed` has no DEFAULT on the ADD COLUMN, so it's NULL for existing rows, and the UPDATE leaves it untouched. NULL = no failure, which is the right state for completed tenants.

PR #1020, opened 2026-04-02
Improve tenant slug field

rchlfryn:
## Description

The tenant slug drop down was a confusing UX because it displayed names first rather than the slug (which is the actual selection). This PR improves the tenant slug field UX in the admin panel. The slug is now displayed in uppercase throughout (list view and edit view), and is read-only on existing tenants since it should never change after creation.

## Key Changes

- **SlugCell**: New custom cell component that renders the slug in uppercase in the list view
- **TenantSlugField**: On existing tenants, renders a read-only `SelectInput` showing the slug in uppercase. On new tenants, filters out already-used slugs from the dropdown as before. (bug introduced in #997)
- **Slug option labels**: Changed format from `"Name (slug)"` to `"SLUG — Name"` to emphasize the slug value

## How to test

2. Verify slugs display in uppercase in the list view
3. Open an existing tenant — the slug field should be read-only and display the slug in uppercase (e.g., "NWAC")
4. Create a new tenant — the slug dropdown should only show unused slugs, with labels like "NWAC — Northwest Avalanche Center"

## Screenshots / Demo video
**When choosing**
Screenshot 2026-04-15 at 10 32 00

**Before save**
Screenshot 2026-04-15 at 10 32 24

**After save**
Screenshot 2026-04-15 at 10 32 35

## Migration Explanation

No migration needed — these are admin UI display changes only.


busbyk (commented):
Isn't this a little confusing though because the slug value is actually lowercased, not uppercased? 

I suppose it doesn't really matter.

rchlfryn:
@busbyk We can make it lower case. Not a problem. I made it uppercase thinking no avalanche center probably uses the lower case acronym, but the acronym is different than the slug.

busbyk (approved):

PR #1024, opened 2026-04-06
Add Jest ESM transform support and use real Payload configs in tests

busbyk:
## Description

Enables Jest to transform Payload/Lexical ESM packages so tests can import real block configs directly instead of relying on synthetic mocks that can drift from reality.

`next/jest` hardcodes `/node_modules/` in `transformIgnorePatterns`, preventing ESM packages like `payload` and `@payloadcms/*` from being transformed by SWC. This adds `addEsmTransformSupport()` to override that pattern, then refactors existing tests to use real configs.

See: https://github.com/vercel/next.js/issues/35634

## Related Issues

Prerequisite for the revalidation PR which adds tests that import Payload configs.

## Key Changes

- **`jest.config.mjs`** — Add `addEsmTransformSupport()` helper to override `next/jest`'s hardcoded `transformIgnorePatterns` for Payload/Lexical ESM packages
- **`docs/coding-guide.md`** — Add "Testing with Payload Configs" section documenting when to use real configs vs mocks
- **`__tests__/server/duplicatePageToTenant.server.test.ts`** — Remove 4 block config mocks, add lexical stub, import real configs
- **`__tests__/server/provisionTenant.server.test.ts`** — Remove 4 block config mocks, import real configs
- **`__tests__/server/fixtures/mockBlocks.ts`** — Deleted (no longer needed)

## How to test

1. `pnpm test` — all 286 tests should pass
2. Verify the `duplicatePageToTenant` and `provisionTenant` tests now import real block configs instead of mocks

## Screenshots / Demo video

N/A — test infrastructure change

## Migration Explanation

No migration needed.

None


jest.config.mjs
```diff
@@ -4,6 +4,23 @@ const createJestConfig = nextJest({
...
+    transformIgnorePatterns: ['^.+\\.module\\.(css|sass|scss)$'],
+    moduleNameMapper: {
+      ...jestConfig.moduleNameMapper,
+      // file-type is a pure ESM package that Jest/SWC can't transform.
+      // No tests need actual file-type detection, so stub it out.
+      '^file-type$': '<rootDir>/__tests__/server/fixtures/fileTypeStub.js',
+    },
+  }
+}
+
```
rchlfryn, line 23: Right now `addEsmTransformSupport()` removes the `node_modules` ignore entirely, so Jest translates every dependency before running tests. This is slow and is what caused the `file-type` stub to be needed. Instead, you can replace the pattern with one that only translates Payload packages:
```mjs
function addEsmTransformSupport(jestConfig) {
  const esmPackages = ['payload', '@payloadcms', '@lexical'].join('|')
  return {
    ...jestConfig,
    transformIgnorePatterns: jestConfig.transformIgnorePatterns.map((pattern) =>
      pattern === '/node_modules/'
        ? `/node_modules/(?!(${esmPackages})/)`
        : pattern,
    ),
  }
}
```
Then you can delete `fileTypeStub.js`, since `file-type` stays ignored. Thoughts?

__tests__/server/duplicatePageToTenant.server.test.ts
```diff
@@ -1,21 +1,18 @@
...
-  // eslint-disable-next-line @typescript-eslint/no-require-imports
-  ButtonBlock: require('./fixtures/mockBlocks').ButtonBlock,
-}))
-
-jest.mock('../../src/blocks/Callout/config', () => ({
-  // eslint-disable-next-line @typescript-eslint/no-require-imports
-  CalloutBlock: require('./fixtures/mockBlocks').CalloutBlock,
+// Real block configs are used thanks to ESM transform support in jest.config.mjs.
+// lexicalEditor() calls at module initialization need a stub since there's no running Payload instance.
+jest.mock('@payloadcms/richtext-lexical', () => ({
```
rchlfryn, line 3: I wonder if we want to make this dynamic somehow

rchlfryn (changes requested):
Thanks for adding this - real configs over mocks is way better.

jest.config.mjs
busbyk: Did you try this? I'm trying it out and keep getting issues with transitive dependencies of Payload that are ESM. Even if we targeted all of those specifically, file-type was still giving me issues and needed stubbing. 

The time savings was about .6s so yea it would be nice to avoid this but I couldn't get it to work for me. Let me know if you weren't running into errors with this config though and we can see what might be different on my machine?

__tests__/server/duplicatePageToTenant.server.test.ts
busbyk: Do you mean breaking this out into a share mocking function/utility? What do you mean?

rchlfryn: Very vague on my part here. It seems like we have a lot of lists around the richtext editor/ blocks, and I wonder if we can consolidate them or make them dynamic somehow so we don't have to always remember to add them places. Thinking outloud.

jest.config.mjs
rchlfryn: I did not try when I first posted and was in research land and thought that would be better but turns out it doesn't work with pnpm. It symlinks them through `.pnpm/` and jest's resolver can't follow those symlinks. Kindly ignore

__tests__/server/duplicatePageToTenant.server.test.ts
busbyk: Ah I see. Well this kind of accomplishes that goal of reducing the amount of lists around :) At least removes one!

I think as of right now though, there are only the collection configs and the https://github.com/NWACus/web/blob/refactor/esm-test-support/src/constants/defaults.ts#L18 right?

busbyk:
@rchlfryn merging is blocked on your requested changes review - is there anything you need me to do here?

rchlfryn (approved):

__tests__/server/duplicatePageToTenant.server.test.ts
rchlfryn: Fair. And yes that is the only list currently

PR #1025, opened 2026-04-06
Document References Extraction System

busbyk:
## Description

The foundation for unified document reference revalidation — a single documentReferences field on all routable collections that replaces the two parallel revalidation subsystems (block reference tracking in `richText` type fields + relationship reference tracking). This PR adds the field, the extraction system, and a backfill migration. The old system remains in place and continues to drive revalidation; the new system runs alongside it temporarily for comparison.

This is the first phase. #1026 implements the revalidation logic based on querying `documentReferences` and removes the old system. 

## Related Issues

First step for [#455](https://github.com/NWACus/web/issues/455).

## Key Changes

- **extractDocumentReferences** — Core extraction system that recursively walks any document and extracts all relationship/upload references at any nesting depth (relationship, upload, blocks, richText/Lexical, group, array, row, collapsible, tabs).
- **DocumentReference shape** — References are grouped by `collection` + `docId` with an `instances` array that preserves every location where the document is referenced (blockType + fieldPath). This means if the same media appears in a HeroBlock and a ContentBlock, both instances are tracked.
- **Thorough test suite** that uses a combination of synthetic field configs/data and fixtures modeled on production data.
- **documentReferences field** — Added to all 4 routable collections (pages, posts, homePages, events). Hidden in admin UI via `disabled: true`.
- **populateDocumentReferences hook** — Generic beforeChange hook that merges `originalDoc` with incoming `data` before calling extractDocumentReferences, ensuring partial updates (where `data` only contains changed fields) still produce a complete set of references. Populates the field on every save (including drafts).
- **Backfill migration** — Re-saves all published routable documents to trigger the hook and populate documentReferences for existing data. Drafts are not backfilled — they'll be populated naturally on their next save or not at all since they don't appear on pages (we only display published content).
- **Tenant relationship filtering** — extractDocumentReferences now skips tenants references (both non-polymorphic and polymorphic), consistent with getRelationshipsFromConfig. Every tenant-scoped document has a tenant ref, but these aren't useful for revalidation tracking.

## How to test

1. Run pnpm test
2. Start dev server, open a page/post in the admin, save it with deeply nested relationships — documentReferences is populated
3. To view the field value open the API tab

## Screenshots / Demo video

https://www.loom.com/share/1e190bb2960445adb5f80492a070321a

## Migration Explanation

20260404_012604_add_documentReferences_field - Adds the documentReferences field to routable collections.

20260404_015415_backfill_document_references — Data-only migration (no schema change). Iterates all published documents in pages, posts, homePages, and events, performing a no-op update (data: {}) on each. This triggers the populateDocumentReferences beforeChange hook, populating the new documentReferences field. Uses context.disableRevalidate to prevent cascading revalidation during the backfill. Draft documents are not backfilled.

#### The next PR will:

1. Create unified findDocumentsWithReferences query function — Single function that queries documentReferences across all routable collections, replacing findDocumentsWithBlockReferences + findDocumentsWithRelationshipReferences
2. Create unified revalidateDocumentReferences function — Orchestration function that calls the query function then revalidates each result
3. Update reference collection hooks — Switch all 7 hooks from old dual-call pattern to single revalidateDocumentReferences call
4. Remove old system — Delete old fields (blocksInContent, blocksInHighlightedContent), hooks, and utilities; generate migration to drop fields
5. Increase revalidation intervals — Once validated, increase revalidate from 600s to 3600s+

#### Performance monitoring

The populateDocumentReferences hook runs on every save (not just publish). The extraction function is synchronous and walks the document's JSON tree, so it should be fast for typical documents. If performance becomes a concern, we can add timing instrumentation or restrict the hook to publish-only as a future enhancement.

#### Restricting lookups to configured blocks

When walking a richText field's Lexical AST, the function builds a Map<slug, Block> from the `BlocksFeature`'s `blocks` array (i.e. the allowed blocks). i.e.:

```
lexicalEditor({
  features: ({ rootFeatures }) => [
    ...rootFeatures,
    BlocksFeature({ blocks: [MediaBlock, CalloutBlock, ButtonBlock] }),
  ]
})
```

For each `type: 'block'` node in the AST, it looks up `fields.blockType` in the map. If the blockType isn't in the allowed blocks, it's skipped (the "unknown blockType" edge case tests). If it IS found, the Block's fields schema is used to recursively extract references from that block's data.

This feels like the correct decision but I felt like it was worth noting in case someone sees an issue with this logic.


src/hooks/populateDocumentReferences.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Generic beforeChange hook that populates the `documentReferences` field
+ * by walking the full document and extracting all relationship/upload references.
+ *
+ * Works for any collection — uses the collection's resolved field config
+ * so it handles all field types including richText with BlocksFeature at any
+ * nesting depth.
+ */
+export const populateDocumentReferences: CollectionBeforeChangeHook = ({ data, collection }) => {
```
stevekuznetsov: Is it possible for there to be a cycle in the references, and, if so, what happens?

busbyk: I don't believe it's possible for this to cycle. This is a before change hook so we can modify data here and that does not re-trigger hooks. 

Cycling is probably more of a concern in the next PR #1026 but that's also not re-triggering hooks because the revalidation system is only calling Next.js' `revalidatePath` which does not change any data in the db or interact with the Payload Local API or Rest API besides find queries. 

Good thought though and we should ensure this isn't possible. So I added a test for that in #1026: https://github.com/NWACus/web/pull/1026/changes/ce5e8d82c2a04c3a4074698e4476f163ccfe3d9b (that PR is still in draft atm fyi so hold off on full review until I mark ready).

src/utilities/extractDocumentReferences.ts
```diff
@@ -0,0 +1,363 @@
...
+ * extracting all relationship and upload references into a flat array.
+ *
+ * Handles: relationship, upload, blocks, richText (Lexical), group, array,
+ * row, collapsible, and tabs fields at any nesting depth.
+ */
+export function extractDocumentReferences(fields: Field[], data: DataObject): DocumentReference[] {
+  const refs: DocumentReference[] = []
+  walkFields(fields, data, '', null, refs)
+
+  // Deduplicate on collection + docId (keep first occurrence)
```
rchlfryn: The dedup key is `collection:docId`, which means if the same document is referenced from two different blocks/fields, only the first occurrence's `blockType` and `fieldPath` are kept. For example, if the same media appears in a `HeroBlock` and a `ContentBlock`, only the HeroBlock reference is kept.

This is fine if we are only checking "does doc A reference doc B?", but I think we are shooting for "which blocks reference doc B?".  If the latter is true, keying on `collection:docId:blockType:fieldPath` or something would capture all the blocks as intended.

src/fields/documentReferencesField.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Reusable field definition for tracking all document references within a collection.
+ * Intended to be populated by the populateDocumentReferences beforeChange hook.
+ * Hidden in admin UI — visible in API responses.
+ */
+export const documentReferencesField = (): Field => ({
+  name: 'documentReferences',
+  type: 'array',
+  admin: {
+    disabled: true,
```
rchlfryn, line 12: Should this be `hidden:true`? Or are you keeping for now for the other PR/testing?

src/migrations/20260414_030626_add_documentReferences_field.ts
rchlfryn, line 1: Limited knowledge on this but the reverse lookup #1026 will need `"find all documents that reference doc X" — queries WHERE collection = ? AND doc_id = ?`. Might be worth adding indexes on these columns (`collection`, `doc_id`), otherwise, those queries will scan the table of every `*_document_references table`.

Worth adding the indexes in this migration or a follow-up before #1026 lands:

```ts
CREATE INDEX `home_pages_document_references_collection_doc_id_idx`
  ON `pages_document_references` (`collection`, `doc_id`);
```
(And the same for posts, events, pages, and their `_v` counterparts.)

src/utilities/getBlocksFromConfig.ts
```diff
@@ -35,7 +35,7 @@ export async function getBlocksFromConfig() {
   // Find blocks used in Events collection (in richText lexical editor)
   const eventsCollection = config.collections?.find((collection) => collection.slug === 'events')
   const eventsBlocks = extractBlocksFromRichTextBlocksFeature(eventsCollection?.fields || [])
-  const eventsBlockMappings = extractBlockMappings(postsBlocks)
+  const eventsBlockMappings = extractBlockMappings(eventsBlocks)
```
rchlfryn, line 38: Was this done in https://github.com/NWACus/web/pull/1023? Might just be https://github.com/NWACus/web/pull/1024 is out of sync

src/hooks/populateDocumentReferences.ts
rchlfryn: Do we want to run this on publish only? I could see this being a possible performance hit?

src/hooks/populateDocumentReferences.ts
```diff
@@ -0,0 +1,20 @@
...
+  const documentReferences = extractDocumentReferences(collection.fields, data)
+
+  return {
+    ...data,
+    documentReferences,
+  }
+}
```
rchlfryn: I think we might want to pass a merged object of `originalDoc`  and `data` here.

Per the payload docs https://payloadcms.com/docs/hooks/collections#beforechange

> On update operations, data contains only the fields being changed. **It may omit** id and **any unchanged fields**....

`data` is typed as `Partial<T>` and contains only the incoming request payload, not the full stored document — Payload passes `originalDoc` separately and does not merge the two. Only using `data` might pass incomplete data into `extractDocumentReferences`. 

This does not explain why your loom worked.

```suggestion
export const populateDocumentReferences: CollectionBeforeChangeHook = ({
  data,
  originalDoc,
  collection,
}) => {
  const merged = { ...originalDoc, ...data }
  const documentReferences = extractDocumentReferences(collection.fields, merged)
  return { ...data, documentReferences }
}
```

rchlfryn (changes requested):
Friendly reminder to regenerate migrations to keep the JSON files in sync.

src/fields/documentReferencesField.ts
busbyk: I considered that. Having `disabled: true` means this does not appear in the admin panel, accomplishing the same thing. There's a slight difference in that `hidden: true` would still render a hidden input whereas `disabled: true` does not. But they both mean that the field doesn't show so it doesn't make sense to use both. 

Actually it's silly to use `readOnly: true` here for the same reason. So I'll remove that. Was just using the same pattern I used in `blocksInContent` but I think I hadn't realized the overlap in effect: https://github.com/NWACus/web/blob/revalidation/src/collections/Events/index.ts#L184-L185

src/utilities/extractDocumentReferences.ts
busbyk: Ahh yes good thought. Hmm as mentioned in the demo and the PR description, I'm only including `blockType` and `fieldPath` for potential debugging purposes. I don't actually have a reason to keep them around.

So you're totally right that for debugging purposes, this would miss multiple instances of the same document. 

I think I'm inclined to either:
1. Change the shape to be: 
```
{
  collection: '',
  docId: '',
  instances: [{ blockType: '', fieldPath: '' }, { blockType: '', fieldPath: '' }]
}
```
2. Remove `fieldPath` and `blockType` completely 

Leaning towards 1.

src/utilities/getBlocksFromConfig.ts
busbyk: I think this was because I'd cherry picked those changes out of this branch. I just rebased so these changes shouldn't be showing up anymore.

src/hooks/populateDocumentReferences.ts
busbyk: Yea definitely a possible performance hit. We could add monitoring to see what impact this has and restricting to only running on publish could be a future enhancement?

src/hooks/populateDocumentReferences.ts
busbyk: Ah good call. The reason my Loom worked is because the admin panel sends the entire document in it's REST API request so all of the data is present. But if we only sent a couple fields in an update request to the rest API in some custom code, we'd run into the case you're describing. Good catch!

I'll update.

src/hooks/populateDocumentReferences.ts
busbyk: Ehh I'm leaning the other way now. Walking an average document should be quite quick and adding performance logging would add a lot of logs IMO. I think let's see if we start to see some slow saves / users complaining.

src/fields/documentReferencesField.ts
busbyk: [23527ff](https://github.com/NWACus/web/pull/1025/commits/23527ffdcf6fc69744c7140f97ee8965e2fffb5a) and then [1383c8b](https://github.com/NWACus/web/pull/1025/commits/1383c8b02748d0077afe473b888617e89021338f) to avoid a new table (json field)

src/utilities/extractDocumentReferences.ts
busbyk: [6716caf](https://github.com/NWACus/web/pull/1025/commits/6716caf4161bcda411cd9e84d863f67079f3fc1a)

src/hooks/populateDocumentReferences.ts
busbyk: [46a4d8d](https://github.com/NWACus/web/pull/1025/commits/46a4d8dc9820b5b69501109745ab98e2300df5c3)

src/hooks/populateDocumentReferences.ts
rchlfryn: Let's just keep an eye on it.

src/migrations/20260414_030626_add_documentReferences_field.ts
rchlfryn: Looks like this was taken care of in [c976e80](https://github.com/NWACus/web/pull/1025/commits/c976e804179c65712daf50ab83d364f5b6c2a370)

rchlfryn (approved):
All changes look good!

src/migrations/20260414_030626_add_documentReferences_field.ts
busbyk: Yes! Forgot to reply here. Thanks for noticing.

PR #1026, opened 2026-04-06
Unified Revalidation

### LINKED ISSUE #455: Deeply nested block revalidation

busbyk:
Make sure to check out https://github.com/NWACus/web/pull/1025 first as this PR is dependent on/based on that as it's first phase. 

## Description

Replaces the two parallel revalidation subsystems (block reference tracking and relationship reference tracking) with a single unified system that discovers all document references at any nesting depth. This fixes confirmed bugs where events had
wrong block mappings, events were missing from relationship queries entirely, and deeply nested references (e.g., a SponsorsBlock inside a ContentBlock's richText) were invisible to the old system.

The old system required per-collection tracking fields, per-collection population hooks, and hand-maintained config mapping functions. The new system uses Payload's collection config to automatically discover all relationship, upload, and block
references in any document — no manual wiring needed when adding new collections or blocks.

With full on-demand revalidation coverage, the ISR safety-net interval is increased from 600s (10 min) to 3600s (1 hour).

## Related Issues

Closes #455 

## Key Changes

**New unified system (3 files):**
- `src/utilities/findDocumentsWithReferences.ts` — Queries the `documentReferences` JSON field across all routable collections
- `src/utilities/revalidateDocumentReferences.ts` — Orchestrates find → revalidate, replacing both `revalidateBlockReferences` and `revalidateRelationshipReferences`
- `src/hooks/revalidateForm.ts` — Updated to use unified system

**Updated revalidation hooks (8 files):**
- All collection revalidation hooks (`revalidatePost`, `revalidateEvent`, `revalidateBiography`, `revalidateMedia`, `revalidateDocuments`, `revalidateSponsors`, `revalidateTag`, `revalidateTeam`, `revalidateForm`) now call
`revalidateDocumentReferences` instead of the old dual `revalidateBlockReferences` + `revalidateRelationshipReferences`

**Schema changes:**
- Removed `blocksInContent` from Posts and Events, `blocksInHighlightedContent` from HomePages

**Removed (10 files, ~1,370 lines):**
- `revalidateBlockReferences.ts`, `revalidateRelationshipReferences.ts`
- `findDocumentsWithBlockReferences.ts`, `findDocumentsWithRelationshipReferences.ts`
- `getBlocksFromConfig.ts`, `getRelationshipsFromConfig.ts`, `extractBlockReferencesFromLexical.ts`
- `populateBlocksInContent.ts` (Posts + Events copies), `populateBlocksInHighlightedContent.ts`

**ISR interval:**
- Increased from `revalidate = 600` to `revalidate = 3600` across all 5 routable page components (this has to be statically analyzed so we can't extract this into a constant)

**Tests:** New tests covering `extractDocumentReferences`, `findDocumentsWithReferences`, `revalidateDocumentReferences`, and `revalidateDocument` (cycle-safety invariant)

## How to test

1. `pnpm tsc` — no type errors
2. `pnpm lint` — no lint errors
3. `pnpm test` — all tests pass
4. Functional: edit a document in the admin that is referenced inside a block within a richText field on another page → the referencing page should revalidate immediately

## Screenshots / Demo video

https://www.loom.com/share/9f6de2da56c24be8976a95223dbe30d7

## Migration Explanation

- **`20260406_212933_remove_old_block_tracking_fields`** — Drops 6 tables: `posts_blocks_in_content`, `events_blocks_in_content`, `home_pages_blocks_in_highlighted_content` + their version tables. **Intentional data loss** — this data has been fully replaced by the `documentReferences` field backfilled in a prior migration on the `revalidation` branch. See `_explanation.md` alongside the migration.

- Let's monitor for stale content reports for 1-2 weeks post-deploy. If stable, let's consider increasing ISR interval further (4h, 24h).


src/utilities/revalidateDocument.ts
```diff
@@ -19,29 +22,21 @@ export interface RevalidationReference {
 }
 
 export interface DocumentForRevalidation {
-  collection: 'pages' | 'posts' | 'homePages' | 'events'
+  collection: string
```
rchlfryn, line 21: Can we add the typing (`'pages' | 'posts' | 'homePages' | 'events'`) back in here?  Or what was the reason for removing it?

Typing here would then allow us to get rid of the warn at the bottom of this file

src/collections/Biographies/hooks/revalidateBiography.ts
```diff
@@ -3,46 +3,24 @@ import type { CollectionAfterChangeHook, CollectionAfterDeleteHook } from 'paylo
...
 
   try {
-    // Also revalidate teams that contain this biography (i.e. pages/posts with TeamBlocks)
-    payload.logger.info(`Checking for cascading team references for biography ID ${biographyId}`)
-
     const teamsWithBiographyRes = await payload.find({
       collection: 'teams',
       where: {
-        members: { equals: biographyId },
+        members: { contains: biographyId },
```
rchlfryn: Does `equals` no longer work here?

src/utilities/findDocumentsWithReferences.ts
```diff
@@ -0,0 +1,70 @@
...
+  const payload = await getPayload({ config: configPromise })
+  const results: DocumentForRevalidation[] = []
+
+  const allSlugs = new Set(payload.config.collections.map((c) => c.slug))
+
+  const collectionsWithReferences = payload.config.collections
+    .filter((c) => c.fields.some((f) => 'name' in f && f.name === 'documentReferences'))
+    .map((c) => c.slug)
+
+  for (const collectionSlug of collectionsWithReferences) {
```
rchlfryn: The collection queries are independent of each other — this could use `Promise.allSettled` to run them concurrently instead of sequentially. `allSettled` preserves the existing error-resilience behavior (one collection failing doesn't block the others).

src/utilities/findDocumentsWithReferences.ts
```diff
@@ -0,0 +1,70 @@
...
+      const res = await payload.find({
+        collection: collectionSlug,
+        where: {
+          and: [
+            { _status: { equals: 'published' } },
+            { 'documentReferences.collection': { equals: reference.collection } },
+            { 'documentReferences.docId': { equals: reference.id } },
+          ],
+        },
+        depth: 1,
```
yuliadub, line 44: Claude tells me that: Only id, slug, and tenant are used, but depth: 1 and no select means full documents (including large layout block arrays and rich text content) are fetched and populated. Adding select: { id: true, slug: true, tenant: true } would significantly reduce payload (pun intended) for pages with heavy content.

yuliadub (approved):
reviewed with help of Claude, looks good to me. The only suggestion I agreed with I put in the comments. Thanks for getting this figured out!

src/utilities/findDocumentsWithReferences.ts
busbyk: Oo nice! I like it [0ff7d36](https://github.com/NWACus/web/pull/1026/commits/0ff7d3672ae890c45542a812f359c83797f4f932)

src/collections/Pages/hooks/revalidatePage.ts
```diff
@@ -65,6 +66,8 @@ export const revalidatePage: CollectionAfterChangeHook<Page> = async ({
     })
 
     revalidatePageTags(tenant.slug)
+
+    await revalidateDocumentReferences({ collection: 'pages', id: doc.id })
```
rchlfryn: `revalidateDocumentReferences` is called in the "currently published" block but not in the "previously published but now unpublished or slug changed" block below (lines 75-87). So when a page is unpublished or its slug changes, other docs that embed it via collection-relation blocks won't revalidate. do we want that?

Posts and Events avoid this by putting the reference-revalidation call inside a shared `revalidate()` helper that runs in both code paths. Can we do the same here?

src/utilities/revalidateDocumentReferences.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  try {
+    const documentsToRevalidate = await findDocumentsWithReferences(reference)
+
+    payload.logger.info(
+      `Found ${documentsToRevalidate.length} documents referencing ${reference.collection} ID ${reference.id}`,
+    )
+
+    for (const doc of documentsToRevalidate) {
+      await revalidateDocument(doc)
```
rchlfryn: Do we want a `try/catch` here? If `revalidateDocument` throws on one doc, the outer try/catch aborts the loop and the remaining docs never get revalidated. Since the errors are swallowed and only logged, a single transient failure silently drops the rest of the batch.

rchlfryn (changes requested):
Minor changes. Overall looks good. Used claude since there were so many files in this one

src/collections/Biographies/hooks/revalidateBiography.ts
busbyk: I believe it wasn't working before because members is an array!

src/utilities/revalidateDocument.ts
busbyk: Keeping `'pages' | 'posts' | 'homePages' | 'events'` as the type here causes a type error on the                                   
[collectionSlug](https://github.com/NWACus/web/blob/unified-revalidation/src/utilities/findDocumentsWithReferences.ts#L56) in `findDocumentsWithRerences` because we're determining collections that could be pushed to the results array by dynamically looking up the collections with `documentReferences` fields:                                                                                 
```                                                                                                                                  
const collectionsWithReferences = payload.config.collections                                                                         
    .filter((c) => c.fields.some((f) => 'name' in f && f.name === 'documentReferences'))                                             
    .map((c) => c.slug)                                                                                                              
```                                                                                                                                  
                                                                                                                                     
I chose to do this because I didn't want to have to maintain a list of collections that we put the `documentReferences` field on. It just felt too easy to forget to update it if we ever made another collection a "routable" collection or otherwise put a `documentReferences` field on it. But looking at this again is making me realize that's silly because we'll need to remember to update these if statements if we ever add a routable collection. 

So I'm just going to use a `ROUTABLE_COLLECTIONS` array and we can use that for the type here.

busbyk: [4a28d79](https://github.com/NWACus/web/pull/1026/commits/4a28d793bc0e40c8989d8150dabaf49817fc3043)

src/utilities/findDocumentsWithReferences.ts
busbyk: I like it [90f7b56](https://github.com/NWACus/web/pull/1026/commits/90f7b56375547640fc26a67524d71310de151bf8)

src/collections/Pages/hooks/revalidatePage.ts
busbyk: Good call! [c920b11](https://github.com/NWACus/web/pull/1026/commits/c920b114fb175a3eb91f5a797de2f1ca24fc17ee)

src/utilities/revalidateDocumentReferences.ts
busbyk: Good call [92ef251](https://github.com/NWACus/web/pull/1026/commits/92ef251a8b6146b592f1783bbb1ae08e33c22de2)

rchlfryn (approved):
Changes look good

PR #1037, opened 2026-04-14
Recursive revalidation through intermediate collections

busbyk:
## Description

Fixes a gap in the revalidation system where changes to sub-resources of non-routable collections don't propagate to the pages displaying them.

**The problem:** When a media image used as a sponsor's photo is updated, pages displaying that sponsor via `SponsorsBlock` are not revalidated. The revalidation system only tracked references one level deep — `findDocumentsWithReferences` only queries collections with a `documentReferences` field (routable collections like Pages, Posts, Events, HomePages). Non-routable collections like Sponsors sit between Media and Pages in the reference chain, creating a blind spot:

```
Media (image changes) → Sponsor (has photo field) → Page (has SponsorsBlock)
                         ↑ invisible to revalidation
```

The same gap affects Biographies (photo → media) and Teams (members → biographies). We had previously been handling this gap with some custom code in the Biographies revalidation hooks but we weren't for other collections that had the same intermediate position. 

**The fix:** Give non-routable intermediate collections their own `documentReferences` field, then make `revalidateDocumentReferences` recursive — when it finds a non-routable intermediate document, it follows the chain through it to reach routable documents. A `visited` set prevents cycles and deduplicates.

After the fix, the full chain works:
```
Media photo changes
  → revalidateDocumentReferences({ collection: 'media', id: 123 })
    → finds Page ID 1 (directly uses the image) → revalidatePath ✓
    → finds Sponsor ID 5 (photo field) → non-routable, recurse
      → finds Page ID 7 (SponsorsBlock) → revalidatePath ✓
    → finds Biography ID 3 (photo field) → non-routable, recurse
      → finds Team ID 2 (members field) → non-routable, recurse
        → finds Page ID 15 (TeamBlock) → revalidatePath ✓
```

## Related Issues

- #1025

## Key Changes

1. **Added `documentReferencesField` + `populateDocumentReferences`** to Sponsors, Biographies, and Teams — these now track their own references to media/biographies
2. **Made `revalidateDocumentReferences` recursive** — non-routable results trigger further lookups instead of being ignored. `ROUTABLE_COLLECTIONS` is now exported from `revalidateDocument.ts` as the single source of truth
3. **Simplified `revalidateBiography`** — removed manual team cascading logic since the generic recursion handles biography → team → page automatically
4. **Fixed `findDocumentsWithReferences` `_status` filter** — only applies `_status: published` filter to collections with drafts enabled (Sponsors, Biographies, Teams don't have versioning and errored on this query)
5. **Added developer guidance** in `docs/coding-guide.md` on when to add `documentReferencesField` to intermediate collections
6. **Backfill migration** to populate `documentReferences` on existing Sponsor, Biography, and Team documents

## How to test

1. `pnpm seed` to get a fresh database with the new fields populated
2. Save a Sponsor block on a routable collection document
3. In the admin panel, edit a Sponsor's photo (swap to a different image)
4. Check server logs — should see recursive revalidation finding the page(s) that display the sponsor
5. Verify the page showing that sponsor reflects the updated image without a manual rebuild

Automated tests cover:
- Recursive resolution through non-routable intermediate collections
- Mixed routable + non-routable results
- Deduplication (same page found via two paths, only revalidated once)
- Cycle protection (contrived cycle terminates without infinite recursion)
- Deep chains (media → biography → team → page)
- `_status` filter applied conditionally based on drafts

## Screenshots / Demo video

https://www.loom.com/share/9bd26892e4b24f72b41304aaa8c046da

## Migration Explanation

**`20260424_180242_intermediate_collection_document_references`** — Creates `documentReferences` sub-tables for sponsors, biographies, and teams. Purely additive (no existing tables modified). The `ON DELETE cascade` foreign key clauses flagged by the safety check are standard Payload array sub-table behavior — when a parent row is deleted, its reference tracking rows are cleaned up.

**`20260424_180419_backfill_intermediate_document_references`** — Data migration that triggers a no-op update on every existing Sponsor, Biography, and Team document to populate the new `documentReferences` field via the `populateDocumentReferences` beforeChange hook.

- The `ROUTABLE_COLLECTIONS` set in `revalidateDocument.ts` is needed and nice for typing correctly but I'd like this to be moved into a set of utilities for getting application paths / canonical urls from collection configs. Future PR - just something I've been thinking about.


src/collections/Biographies/index.ts
```diff
@@ -1,8 +1,10 @@
 import { accessByTenantRole } from '@/access/byTenantRole'
 import { filterByTenant } from '@/access/filterByTenant'
 import { contentHashField } from '@/fields/contentHashField'
+import { documentReferencesField } from '@/fields/documentReferencesField'
```
rchlfryn, line 4: Since Sponsors, Biographies, and Teams are commonly readable by the public (they render on public pages), every public fetch of a sponsor will now include the full back-reference list of pages that use it.

This isn't a security hole, but might bloats response bodies for high-traffic endpoints, and it leaks "page X has a draft sponsor block referring to me" if those references include unpublished pages. Maybe add to `documentReferencesField`:

```js
access: { read: ({ req: { user } }) => Boolean(user) }
```

src/migrations/20260424_180419_backfill_intermediate_document_references.ts
```diff
@@ -0,0 +1,36 @@
...
+export async function up({ payload }: MigrateUpArgs): Promise<void> {
+  for (const collection of INTERMEDIATE_COLLECTIONS) {
+    const docs = await payload.find({
+      collection,
+      limit: 0,
+      depth: 0,
+    })
+
+    for (const doc of docs.docs) {
+      await payload.update({
```
rchlfryn: do we want a try/catch around this? One bad doc aborts the entire migration mid run

rchlfryn (changes requested):
This is sweet - too minor comments

src/migrations/20260424_180419_backfill_intermediate_document_references.ts
busbyk: Good call [fa3dd2b](https://github.com/NWACus/web/pull/1037/commits/fa3dd2bae59e61108d620bcdf32c29747142748c)

src/collections/Biographies/index.ts
busbyk: Sponsors, Biographies, and Teams all use `accessByTenantRole` so they are not readable by the public - there must be an authenticated user and that user must have a tenant role giving permission to that collection. 

We use the local API for our server components which bypasses these permissions anyways (see [overrideAccess](https://payloadcms.com/docs/local-api/overview#:~:text=array%20of%20locales.-,overrideAccess,-Skip%20access%20control)) which makes sense because for the most part we're using `force-static` so we're not sending data payloads directly. I believe the RSC payload has the risk of being sent along with it although in my local testing I was seeing `documentReferences` as undefined (really it was a placeholder value that I'm inferring was a placeholder for undefined based on other places it was used). 

For /blog and /events, you can easily find `documentReferences` in the page document payload. But for both cases - statically or dynamically rendered - do we really have any issues with the `documentReferences` field being publicly visible (genuine question)? It's just ids of documents with the collection and the field path (in the instances array). This is data that's technically already publicly available because it's rendered in publicly accessible pages. 

Using that field level access control for read doesn't affect what's sent in the RSC payloads because the local API overrides access by default (and we want this for effective page rendering). We could use `defaultPopulate` on each collection that has a `documentReferences` field but we'd need to be careful about making sure we include all fields that consumers are using and that could get difficult to maintain. But if we did that we could then use `select` clauses to explicitly request the `documentReferences` field in revalidation functions. 

That feels riskier than the value. If we agree that `documentReferences` values are safe to be publicly discoverable in the RSC payloads then I suggest we lave this as is. I'm not too concerned about response size bloat since `documentReferences` items are pretty lightweight even if their instances get up to the tens of items.

rchlfryn: Thank you for looking into this and explaining this. You're right on both fronts. My mental model was wrong.

These collections are gated by `accessByTenantRole`, so there's no public REST fetch path. I conflated "renders on public pages" with "publicly readable via the API," which isn't the case. Also, I was thinking about this as a back-reference list ("pages that use this sponsor"), but `documentReferences` on a sponsor is a "forward" list. The page→sponsor reference lives on the page's `documentReferences`, which has been there since the original revalidation PR.


> do we really have any issues with the documentReferences field being publicly visible (genuine question)?

I don't think we have an issue with this. I think we can leave it as is. The data (collection name, ID, field path) for published pages is inferable from what's rendered, and draft pages aren't reachable by the public anyway.

rchlfryn (approved):

PR #1058, opened 2026-05-05
Fix: single migration for recent changes

busbyk:
## Description

The `src/migrations/20260414_030934_backfill_document_references.ts` migration failed when [deploying dev](https://github.com/NWACus/web/actions/runs/25341499017/job/74299622393) because it uses the Payload Local API with it's config from the current version of the code which includes `documentReferences` on sponsors, teams, and biographies. It also has a `beforeChange` hook which queries `sponsors_document_references`, `biographies_document_references`, and `teams_document_references` — but those tables aren't created until a later migration `src/migrations/20260424_180242_intermediate_collection_document_references.ts`. 

The obvious solution would be to either combine `20260414_030934_backfill_document_references` and `20260424_180242_intermediate_collection_document_references` and also combine `20260414_030934_backfill_document_references` and `20260424_180419_backfill_intermediate_document_references`. But in order to regenerate the JSON snapshots correctly using `pnpm payload migrate:create {name}`, we'd need to git checkout a specific commit with the Payload config in the right state and regenerate the `20260417_202409_add_dev_mode` and `20260417_225756_remove_old_block_tracking_fields` migrations. 

So instead, I combined several migrations into one and ensured they resulted in the same changes by generating a .diff file using our `pnpm migrate:diff` utility. 

The last migration run in production is `20260319_152951_add_text_wrap_to_image_text_block` so it's fine for us to muck around with these migrations in `main` before they land in `release`. 

It's a lot in a single migration but it felt more reliable to fix this issue this way rather than trying to checkout git commits with the right Payload config state just so we could keep the same migration file separation. 

## Related Issues

N/A

## Key Changes

- Combined `20260414_030934_backfill_document_references`, `20260424_180242_intermediate_collection_document_references`, `20260417_202409_add_dev_mode`, and `20260417_225756_remove_old_block_tracking_fields` migrations into a single migration.
- Recreated `backfill_document_references` migration using a single migration that no-op updates all routable collections and intermediate collections. 
- Renamed `20260403_204010` to `20260403_204010_mcp_server`. This one also hasn't landed in production yet and it was bothering me that it didn't have a descriptive migration name (this was my bad). 

## How to test

Follow `### Testing a migration locally` in `docs/migration-safety.md`

## Screenshots / Demo video

N/A

## Migration Explanation

Explained above

None


rchlfryn (approved):

rchlfryn:
🚀 This is included in version [v1.13.0](https://github.com/NWACus/web/releases/tag/v1.13.0)

Issue #836, opened 2026-01-07
Allow all file types in Documents except for images/video


busbyk:
Allow all file types in Documents _except_ for images and videos. The Media collection is for images and videos. 

The DocumentBlock should be updated to have two layout variants:
1. Embed in an iframe (current behavior)
1. Show a download link

We might want logic to only allowing embedding in an iframe for some mime types - or we could just trust users to make this decision (and maybe default to the download link layout option).

There isn't an ideal way to implement "all but images are allowed" logic for the Documents collection because the mimeTypes field is an allow list: https://payloadcms.com/docs/upload/overview#mimetypes

We could accomplish that "all but images are allowed" logic using a hook with correct error messaging to users in a toast (Payload's error notification pattern) but we lose the nice UX of only allowing users to select file types in the file upload modal (i.e. the system file picker that pops up from the html file input) that they are actually allowed to upload. That's probably something we can accept though.

We could also see about opening a PR to use a deny list for mime types in Payload.

busbyk:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Let the Documents collection accept all file types *except* images and videos (which belong in Media), and give the Document block a download-link layout in addition to the current iframe embed.

**Current behavior:**
The Documents upload config uses an **allow-list** `mimeTypes` (`application/pdf`, php, xml, octet-stream, kml). That both over-restricts (only those types upload) and can't express "anything but images/video." The Document block only embeds documents in an iframe.

**Desired behavior:**
- Documents accepts **any file type except `image/*` and `video/*`** (those go in the Media collection). Because Payload's `mimeTypes` is allow-only, implement the exclusion with an upload **validation hook** that rejects `image/*` and `video/*` with a clear user-facing error (Payload's toast/error pattern), and broaden/remove the allow-list accordingly.
- The Document block gets two layout variants selectable by the editor: (1) **iframe embed** (current) and (2) **download link**. Default to the download-link layout. Optionally restrict the iframe option to embeddable types (e.g. PDF), or trust the editor — keep it simple.

**Key interfaces:**
- Documents collection `upload` config + a validation hook (`beforeOperation`/`beforeValidate`) that inspects the uploaded file's mime type and rejects images/videos with a helpful message.
- The Document block config (add a layout/`displayMode` select) and its render component (render an iframe or a download link accordingly).

**Acceptance criteria:**
- [ ] Uploading an image or video to Documents is rejected with a clear error; other file types upload successfully.
[...]

PR #1072, opened 2026-05-09
Issue-836 Document Types

scotthillson:
## Description
Allow all file types in Documents except for images and videos. The Media collection is for images and videos. See issue for more detail.

## Related Issues
https://github.com/NWACus/web/issues/836

## Key Changes
Adds a layout field to the Document block so editors can choose between embedding a PDF in an iframe or showing a download link button. Also removes the hardcoded mimeTypes allowlist from the Documents collection and replaces it with a validation hook that blocks only images and videos, allowing any other file type (PDFs, KML, XML, etc.) to be uploaded without needing to enumerate every MIME type.
The iframe title was generalized from "Document PDF" to "Document" since non-PDF files can now be embedded.

## How to test
Upload a PDF to the Documents collection — should succeed
Try to upload an image file into documents, you'll see a toast error

## Migration Explanation
Adds a layout column (nullable select) to the blocks_document table. No data is destroyed; existing rows get null which the component treats as embed.


src/blocks/Document/config.ts
```diff
@@ -11,5 +11,14 @@ export const DocumentBlock: Block = {
       relationTo: 'documents',
       required: true,
     },
+    {
+      name: 'layout',
```
busbyk: Can we name this `displayAs` instead? I think that might be clearer to editors.

src/blocks/Document/Component.tsx
```diff
@@ -5,24 +5,46 @@ import { getMediaURL } from '@/utilities/getURL'
...
   const src = getMediaURL(document.url, null, getHostnameFromTenant(tenant))
+  const filename = document.filename ?? 'Download'
+
+  // Treat missing layout as 'embed' so existing blocks keep their current behavior
+  const resolvedLayout = layout ?? 'embed'
+
+  if (resolvedLayout === 'download') {
+    return (
+      <div className={cn('my-4', { container: isLayoutBlock })}>
+        <a
```
busbyk: Clicking this triggers the `nextjs-toploader` loader:

Image

If you add `target="_blank"` we can avoid that.

src/migrations/index.ts
busbyk, line 1: I just added to our `docs/coding-guide.md` to outline our preferences for migration naming and resolving migration conflicts when merging in `main` has new migrations. Can you please follow those instructions and update this PR's migrations accordingly? https://github.com/NWACus/web/pull/1077

src/scripts/check-migrations.ts
```diff
@@ -20,7 +20,8 @@ const DANGEROUS_PATTERNS = [
     message: 'TRUNCATE keyword detected - review for data loss',
   },
   {
-    pattern: /\bALTER\b/i,
+    // ALTER TABLE ... ADD is always safe (additive); only flag destructive variants
+    pattern: /\bALTER\b(?!.*\bADD\b)/i,
```
busbyk: Oo, I think this makes sense. But could you break this out into a separate PR or remove?

If the only reason for this change was to be able to commit your changes, we suggest using `--no-verify` after reviewing to make sure your migration isn't actually destructive. Not perfect, but our goal was to be overly cautious in flagging potentially destructive migrations and then require a dev to review before committing. 

https://github.com/NWACus/web/blob/c0b1916e13718da388660fa676376aaac5e6d480/docs/migration-safety.md?plain=1#L64

src/blocks/Document/Component.tsx
```diff
@@ -5,24 +5,46 @@ import { getMediaURL } from '@/utilities/getURL'
...
+          <span>{filename}</span>
+        </a>
+      </div>
+    )
+  }
 
   return (
     <div className={cn('my-4', { container: isLayoutBlock })}>
-      <iframe src={src} width="100%" height="600px" title="Document PDF" />
+      <iframe src={src} width="100%" height="600px" title="Document" />
```
busbyk: If there is no browser viewer for a given file type and the layout is set to `embed`, the file is automatically downloaded when visiting the page that the DocumentBlock is on: 

Image

I can't find the behavior documented anywhere but it seems that if we set an iframe's src attribute to a file with a MIME type that it does not have a built-in viewer for, the file is automatically downloaded when the iframe is rendered. We definitely don't want this behavior so we need to account for this. 

It seems that PDF, HTML, Plain text, and XML/KML can all be rendered in an iframe. So we need conditional logic in the collection that only allows setting the `displayAs` field to `embed` when the file's MIME type is one of `['application/pdf', 'text/html', 'text/plain', 'text/xml', 'application/xml',                           
  'application/vnd.google-earth.kml+xml', '.kml']`. If we discover there are other MIME types that browsers can reliably render, we could add to that list.

src/blocks/Document/Component.tsx
```diff
@@ -5,24 +5,46 @@ import { getMediaURL } from '@/utilities/getURL'
...
 
   if (!isValidRelationship(document) || !document.url) {
     return null
   }
 
   const src = getMediaURL(document.url, null, getHostnameFromTenant(tenant))
+  const filename = document.filename ?? 'Download'
+
+  // Treat missing layout as 'embed' so existing blocks keep their current behavior
+  const resolvedLayout = layout ?? 'embed'
```
busbyk: This was a good assumption to make but I think we should fallback to `download` due to the browser behavior that might result in a file being automatically downloaded if the browser can't render it in an iframe. See [my other comment](https://github.com/NWACus/web/pull/1072/changes#r3236395404).

We only have one page with DocumentBlocks in production, so this would be easy enough to go and update manually after we deploy this.

busbyk (changes requested):
Thanks for the contribution @scotthillson!

I've requested a few changes. Let me know if you have any questions.

rchlfryn:
@scotthillson Checking in to see if you need any help. Let us know!

scotthillson:
Hey! @busbyk I think this is ready for another pass. Fingers crossed I have this knocked out in just the 1-2 passes!

src/collections/Documents/index.ts
```diff
@@ -38,18 +39,9 @@ export const Documents: CollectionConfig = {
   ],
   upload: {
     staticDir: path.resolve(dirname, '../../../public/documents'),
-    mimeTypes: [
```
rchlfryn, line 41: This won't allow HTML and re-breaks the old allowlisted types.

Uploading HTML fails with Restricted file type detected `-- set 'allowRestrictedFileTypes'` to true. That's Payload core (`checkFileRestrictions`), not the hook. When mimeTypes is absent, it enforces a built-in blocklist (text/html, php, js, exe…) and throws before the hook runs. Removing the allowlist also drops the pdf/php/xml/kml types the old list permitted.

 Fix — keep an allowlist instead of removing it:
```ts
 upload: {
   staticDir: path.resolve(dirname, '../../../public/documents'),
   mimeTypes: [
     'application/pdf',
     'application/xml',
     'text/xml',
     'application/octet-stream',
     'application/vnd.google-earth.kml+xml',
     '.kml',
     'text/html',
     'text/csv',
   ],
 },
```

A set mimeTypes bypasses the blocklist, so HTML passes while unlisted executables stay blocked.  `validateNotImageOrVideo` can be dropped (image/video just aren't listed).

The alternative (remove mimeTypes + `allowRestrictedFileTypes: true`) matches the issue literally but also permits .exe/.js on public URLs. Given staff-only uploads of static docs, the allowlist seems safer.

src/blocks/Document/Component.tsx
```diff
@@ -5,24 +5,58 @@ import { getMediaURL } from '@/utilities/getURL'
...
 
   if (!isValidRelationship(document) || !document.url) {
     return null
   }
 
   const src = getMediaURL(document.url, null, getHostnameFromTenant(tenant))
+  const filename = document.filename ?? 'Download'
+
+  const isEmbeddable = document.mimeType != null && EMBEDDABLE_MIME_TYPES.has(document.mimeType)
+  const resolvedDisplay = displayAs === 'embed' && isEmbeddable ? 'embed' : 'download'
```
rchlfryn: Existing Document blocks render as embedded iframes today. After this PR they flip to download links. The migration backfills `display_as` to 'download', and this line resolves anything not explicitly 'embed' to download. So every published PDF becomes a download button on deploy. I think we can just pass `displayAs` as is instead of creating a place for this to silently fail. Remove this and update to check `displayAs` below.
```suggestion

```

Also adding `required: true` will remove the possibility of this being null

src/blocks/Document/config.ts
```diff
@@ -11,5 +11,14 @@ export const DocumentBlock: Block = {
...
       required: true,
     },
+    {
+      name: 'displayAs',
+      type: 'select',
+      defaultValue: 'download',
+      options: [
+        { label: 'Download Link', value: 'download' },
+        { label: 'Embed (iframe)', value: 'embed' },
+      ],
```
rchlfryn, line 22: Let's add `required: true` here to remove the possibility of `null`

src/blocks/Document/Component.tsx
```diff
@@ -5,24 +5,58 @@ import { getMediaURL } from '@/utilities/getURL'
...
+  'text/plain',
+  'text/xml',
+  'application/xml',
+  'application/vnd.google-earth.kml+xml',
+])
 
 type Props = DocumentBlockProps & {
   isLayoutBlock: boolean
+  // displayAs is present in the block config but absent from generated types until pnpm generate:types is run
+  displayAs?: 'download' | 'embed' | null
```
rchlfryn: This should be passed in `DocumentBlockProps` so we don't need to add it here
```suggestion
```

rchlfryn (changes requested):
Apologies for the delay in the review. Other priorities kept getting in the way. Because of the delay you will need to merge/rebase main and rerun the migration to get the latest.

PR #1077, opened 2026-05-13
Add migration naming and merge workflow docs

busbyk:
## Summary
- Add a **Migrations** section to the coding guide documenting two practices:
  - Always provide a descriptive name when running `pnpm payload migrate:create`
  - How to properly resolve migration conflicts when merging `main` into a feature branch (delete your migrations, recreate them so they're ordered after main's)
- Update `CLAUDE.md` migration references to include the descriptive name argument and link to the new docs section

### DIFF
--- a/docs/coding-guide.md
+++ b/docs/coding-guide.md
@@ -77,15 +77,19 @@ Descriptive names make it easy to understand what each migration does when scann
 
 ### Merging `main` When There Are New Migrations
 
-When you merge `main` into your feature branch and `main` has new migrations, you will get a conflict in `src/migrations/index.ts`. This is expected. Follow this process:
+When you merge `main` into your feature branch and `main` has new migrations, you **must recreate your branch's migrations** — even if there is no merge conflict. The critical reason: each migration's `.json` snapshot must reflect the full schema state at that point, including migrations from `main`. If you skip this step, the snapshot will be missing `main`'s migrations and won't accurately represent the database state.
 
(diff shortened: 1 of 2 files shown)


docs/coding-guide.md
```diff
@@ -59,9 +59,34 @@ Start by copying `src/collections/collection-template.ts` into a new file in `sr
...
+```bash
+pnpm payload migrate:create add_sponsors_block
+pnpm payload migrate:create remove_embed_block_height
+```
+
+Descriptive names make it easy to understand what each migration does when scanning `src/migrations/index.ts` or the migration files on disk.
+
+### Merging `main` When There Are New Migrations
+
+When you merge `main` into your feature branch and `main` has new migrations, you will get a conflict in `src/migrations/index.ts`. This is expected. Follow this process:
```
rchlfryn: Also worth noting that even if there is not a conflict with `main`, you still need to make sure the json contains the new migration. That will include rerunning it the same way. The only important piece to keep from the rerun is the json file, but might as well keep both anyways.

rchlfryn (changes requested):
One note.

docs/coding-guide.md
busbyk: Yea good call [070c73a](https://github.com/NWACus/web/pull/1077/commits/070c73a1a88ae9be5ddec512110628b63f3fc1b5)

rchlfryn (approved):

PR #1090, opened 2026-05-26
Fix column layout null validation and error display

### LINKED ISSUE #1086: Cannot set column layout when value is set to null
Original error report: https://www.loom.com/share/f738a139aef5427dabe1190eade56cee

Workaround: https://www.loom.com/share/df76747fbf3d4442ae64d7e4f7be225f

This appears to be an issue with pages created during our initial production seed (assumption) which have `null` values for their content blocks' layout fields. That causes a validation error on publish and for some reason clicking on the column layout does not change the value from null. 

We should figure out how to display error highlighting for our column layout field while we're fixing this, btw. The error toast shows which field it is but the Payload pattern is to also highlight the offending field and our custom component does not hook into that correctly. 

Sentry Issue: [AVY-WEB-NE](https://nwac.sentry.io/issues/7477053076/?referrer=github_integration)

```
h: The following field is invalid: Content > Layout > Block 1 (Content) > Column layout
  File "web/./node_modules/.pnpm/payload@3.81.0_graphql@16.10.0_typescript@5.7.3/node_modules/payload/dist/fields/hooks/beforeChange/index.js", line 40, in g
    throw new ValidationError({
  File "node:internal/process/task_queues", line 104, in process.processTicksAndRejections
  File "web/./node_modules/.pnpm/payload@3.81.0_graphql@16.10.0_typescript@5.7.3/node_modules/payload/dist/collections/operations/utilities/update.js", line 153, in s
    let result = await beforeChange(beforeChangeArgs);
  File "web/./node_modules/.pnpm/payload@3.81.0_graphql@16.10.0_typescript@5.7.3/node_modules/payload/dist/collections/operations/updateByID.js", line 114, in x
    let result = await updateDocument({
  File "web/./node_modules/.pnpm/payload@3.81.0_graphql@16.10.0_typescript@5.7.3/node_modules/payload/dist/collections/endpoints/updateByID.js", line 9, in bl.handler
    const doc = await updateByIDOperation({
...
(5 additional frame(s) were not displayed)
```

busbyk:
## Description

Pages created during the initial production seed had `null` values for their content block layout fields. This caused a validation error on save, and the custom `ColumnLayoutPicker` component didn't display the error or allow fixing the value.

## Related Issues

Fixes #1086

## Key Changes

- **Self-healing hook** (`src/blocks/Content/hooks/healColumnLayout.ts`): A `beforeChange` field hook that auto-fixes null/invalid layout values to a sensible default matching the current column count
- **Error display** (`src/components/ColumnLayoutPicker/index.tsx`): The custom component now uses `showError`/`errorMessage` from Payload's `useField` hook to display validation errors inline
- **UI auto-heal**: Added a `useEffect` that detects null values on mount and auto-selects a valid layout option matching the column count
- Removed the TODO comment about error display since it's now fixed
- Added 9 unit tests for the healing hook

## How to test

1. If you have a page with a null column layout (from early seed data), open it in the admin panel
2. The layout picker should auto-select a valid option
3. Save — should succeed without validation error
4. Run `pnpm test -- __tests__/server/healColumnLayout.server.test.ts`

## Screenshots / Demo video

https://www.loom.com/share/d8d56a158c404736b0a8adea17a15dea

## Migration Explanation

No migration needed — hook and UI logic only.

### DIFF
--- a/__tests__/server/healColumnLayout.server.test.ts
+++ b/__tests__/server/healColumnLayout.server.test.ts
@@ -1,0 +1,94 @@
+import { healColumnLayout } from '@/blocks/Content/hooks/healColumnLayout'
+
+// Minimal mock for FieldHook args
+const createArgs = (overrides: Record<string, unknown>) =>
+  ({
+    collection: {} as never,
+    context: {},
+    data: {},
+    field: {} as never,
+    global: null,
+    indexPath: [],
+    operation: 'update',
+    originalDoc: {},
+    overrideAccess: false,
+    path: [],
+    previousSiblingDoc: {},
+    previousValue: undefined,
+    req: {} as never,
+    schemaPath: [],
+    siblingData: {},
+    siblingDocWithLocales: {},
+    siblingFields: [],
+    value: null,
+    blockData: undefined,
+    ...overrides,
+  }) as Parameters<typeof healColumnLayout>[0]
+
+describe('healColumnLayout', () => {
+  it('passes through valid layout values', () => {
+    const result = healColumnLayout(
+      createArgs({ value: '2_11', siblingData: { columns: [{}, {}] } }),
+    )
+    expect(result).toBe('2_11')
+  })
+
+  it('heals null to default layout', () => {
+    const result = healColumnLayout(createArgs({ value: null, siblingData: {} }))
+    expect(result).toBe('1_1')
+  })
+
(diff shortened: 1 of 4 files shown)


rchlfryn (approved):

rchlfryn:
@busbyk merging this in

PR #1122, opened 2026-06-18
Filter out content with future publishedAt dates from frontend

### LINKED ISSUE #1092: Filter out content with future publishedAt dates from frontend
## Description

Currently, content with a `publishedAt` date set in the future is still displayed on the frontend. We should respect the `publishedAt` field and filter out content that hasn't reached its publish date yet.

This would enable editors to schedule content for future publication by setting a `publishedAt` date in the future.

## Collections affected

- Posts (used for blog sorting and date display)
- Pages
- HomePages

## What needs to happen

1. Frontend queries should filter by `publishedAt <= now` (or `publishedAt` is null/not set)
2. Consider whether this should also apply to ISR/caching — content with future dates would need revalidation when the publish date arrives
3. Consider adding a "Scheduled" status indicator in the admin UI for content with future publishedAt dates

## Context

Raised while working on #878 (publishedAt field standardization). The Pages collection previously had a misleading description claiming publishedAt "affects the page's visibility and can be used for scheduling future publications" — but this was never actually implemented.

rchlfryn:
## Description

Published Posts, Pages, and HomePages with a `publishedAt` date in the future were being rendered on the frontend. This adds a shared `publishedFilter()` where clause that requires `_status === 'published'` AND a `publishedAt` that is not in the future (documents without a `publishedAt` are treated as published immediately), and applies it across all frontend queries.

## Related Issues

Fixes #1092

## Key Changes

- Add shared `publishedFilter()` where clause in `src/utilities/publishedFilter.ts`
- Apply it to all frontend Posts, Pages, and HomePages queries, sitemaps, and `generateStaticParams`
- Exclude future-published items from `filterValidPublishedRelationships` (related/static posts)
- Add server tests for `publishedFilter()` and a client test for the relationship filter

## How to test

1. Create a Post/Page with `_status: published` and a `publishedAt` set in the future.
2. Confirm it does not render on the frontend, does not appear in sitemaps, and is not in related/static post lists.
3. Confirm published content with a past or missing `publishedAt` still renders.

## Screenshots / Demo video

N/A

## Migration Explanation

No migration — query-level change only.

None.

### DIFF
--- a/__tests__/client/utilities/relationships.client.test.ts
+++ b/__tests__/client/utilities/relationships.client.test.ts
@@ -44,6 +44,26 @@ describe('isValidPublishedRelationship', () => {
     expect(isValidPublishedRelationship({ id: 1, _status: 'draft' })).toBe(false)
   })
 
+  it('returns true for a published object with a past publishedAt', () => {
+    expect(
+      isValidPublishedRelationship({
+        id: 1,
+        _status: 'published',
+        publishedAt: '2020-01-01T00:00:00.000Z',
+      }),
+    ).toBe(true)
+  })
+
+  it('returns false for a published object with a future publishedAt', () => {
+    expect(
+      isValidPublishedRelationship({
+        id: 1,
+        _status: 'published',
+        publishedAt: '2999-01-01T00:00:00.000Z',
+      }),
+    ).toBe(false)
+  })
+
   it('returns false for a number (unresolved ID)', () => {
     expect(isValidPublishedRelationship(42)).toBe(false)
   })
--- a/__tests__/server/publishedFilter.server.test.ts
+++ b/__tests__/server/publishedFilter.server.test.ts
@@ -1,0 +1,27 @@
+import { publishedFilter } from '@/utilities/publishedFilter'
+
+describe('publishedFilter', () => {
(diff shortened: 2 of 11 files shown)


busbyk:
@rchlfryn I think this introduces a revalidation system issue. 

When a user set's a future publishedAt date on a post (or other document but I'll use post for this example), revalidation runs and the path for that post is not statically generated (because now it's not published anymore due to publishedAt being in the future). Visiting that post's path would 404. 

But when the current clock time passes the publishedAt date, no event triggers revalidation to revalidate that path. That path would still 404. I don't think SWR would solve for this either because the 404 is cached so I don't _think_ the next user would see the correct content. Since we have time-based revalidation, the path would revalidate after 1hr but that means our scheduled publishing will always be 1hr behind. 

One way we could get around this is using [Payload's scheduled publish feature](https://payloadcms.com/docs/versions/drafts#scheduled-publish) which uses their [Jobs Queue feature](https://payloadcms.com/docs/jobs-queue/overview) to modify the scheduled document at publish time which _would_ fire our revalidation system. 

I think I'm going to want to use something similar for the native AFP product pages. So we can probably make a note and handle this during that effort; set up the Jobs Queue then.

PR #1127, opened 2026-06-23
Issue 02: per-product source adapter, normalized model, two-control split

busbyk:
## Description

Issue 02 of the native-product-pages effort: the **data layer** every native product page consumes — a **per-product source adapter** returning a **normalized, API-shape-agnostic model**, plus the **two controls** that govern it (rollout flag; data-source config). This is reconcile/refactor, not greenfield: the fetchers, zod types, and components already exist; this reorganizes them behind an adapter seam and adds the two config seams.

> ⚠️ **Stacked PR.** Base is `forecast-date-picker` (issue 09), which is itself stacked on `native-forecast-reconcile` (issue 01, #1121). Linear stack 01 → 09 → 02; this auto-retargets up the stack as the lower PRs merge. Draft until the stack lands.

## Related Issues

- `.scratch/native-product-pages/issues/02-data-layer-adapter-and-controls.md`
- PRD: `.scratch/native-product-pages/PRD.md` · ADR: "Native product page architecture" (to land as `docs/decisions/018-*`)
- Depends on 01 (reconciliation) and is stacked on 09 (date picker)

## Key Changes

- **Normalized model** (`services/nac/model/forecast.ts`) — API-agnostic types the components depend on. Owns the top-level product types (`Forecast`/`Summary`/`Warning`/`WarningProduct`/…) and reuses the leaf domain enums/types from the v2 wire schema. "No active warning" is plain `null`, not v2's null-object, so consumers stop special-casing the miss representation (the duplicated `toWarningProduct` helpers are gone).
- **Source adapter** (`services/nac/sources/`) — `ForecastSource`/`WarningSource` interfaces, a v2 implementation wrapping the existing fetchers with **pure, unit-tested `mapV2*` mappers**, and `getForecastSource`/`getWarningSource` resolvers. Components/pages fetch through the adapter, never a raw fetcher.
- **Control 2 — data source** (`sources/config.ts`) — per-product `v2`/`v3` selection via **code/env config**, uniform across tenants, defaulting to v2, with a per-center v3 canary allowlist. Not a Setting. (`NAC_{FORECAST,WARNING}_SOURCE`, `NAC_{FORECAST,WARNING}_V3_CANARY_CENTERS`.)
- **Control 1 — rollout** — `Settings.useNativeForecasts` becomes a per-product `nativeProducts: { forecast, warning }` group; new `getNativeProductFlag(center, product)` reader. Per-product (not all-or-nothing) so a center can go native on forecast while later products stay on the widget.
[...]

PR #1126, opened 2026-06-22
Forecast date picker (historical browse / season replay)

busbyk:
## Description

Adds a danger-colored calendar date picker to the native forecast page so readers can browse a zone's published forecast history ("season replay") and jump to any past forecast. Implements `.scratch/native-product-pages/issues/09-forecast-date-picker.md`.

> **Stacked PR.** Base is `native-forecast-reconcile` (issue 01, #1121), which is not yet merged into `native-product-pages`. This PR's diff is only the date-picker work; it will auto-retarget to `native-product-pages` once #1121 lands. **Do not merge before #1121.**

> Also rides along on this branch: a small docs-only commit (ADR 017 + DOMAIN_CONTEXT "Forecast Glossary" terms) capturing the forecast-glossary design decided separately — it belongs to issues 06/10, not the date picker, but was committed here with the author's okay.

## Related Issues

- `.scratch/native-product-pages/issues/09-forecast-date-picker.md` (this PR)
- Stacked on #1121 (issue 01 — reconciliation)
- Docs commit relates to issues 06 / 10 (forecast glossary)

## Key Changes

- **New nested route** `forecasts/avalanche/[zone]/[date]/page.tsx` (`date=YYYY-MM-DD`): on-demand server render, `revalidate=2592000` (~immutable), `dynamicParams=true`, `robots noindex`.
- **Shared presentational `NativeForecastView`** extracted from `NativeForecastPage` (warning optional — historical views pass null) so live + dated pages render identically.
- **Client `ForecastDatePicker.client.tsx`**: shadcn `Calendar` (react-day-picker v9) in a Popover, days background-colored by `danger_rating` via `dangerScale.ts`; lazy-loads older months on demand.
- **Route handler `GET /api/[center]/forecast-archive?zone&from&to`** → `{ dates: [{ date, productId, productType, dangerRating }] }` for lazy month color loading; the page never ships the full archive.
- **Pure `services/nac/archiveDates.ts`** (noon valid-date rule in center tz, dedupe-by-date latest-wins) + unit tests.
- Nav via real Next `<Link>`s so the app's `nextjs-toploader` bar shows (it only triggers on anchor clicks, not `router.push`).

## How to test

1. `pnpm dev`, open a NWAC zone forecast (e.g. `nwac.localhost:3000/forecasts/avalanche/west-slopes-north`).
2. Open the date picker — current month's days are danger-colored; page older months and colors lazy-load.
[...]

PR #1121, opened 2026-06-17
Native forecast page: reconcile from replace-widgets + initial polish

busbyk:
## Description

Adds **native Next.js avalanche-forecast pages** (single-zone + all-zones grid) behind the per-tenant `useNativeForecasts` Settings flag: a center that enables it renders native server components; with it off, the routes render the legacy NAC widget exactly as before (instant per-center rollback).

The page was built earlier on the unmerged `replace-widgets` branch (14 commits ahead / ~420 behind `main`); this PR reconciles that work onto current `main`, regenerates the DB migration, and adds review polish.

**Why this targets `native-product-pages`, not `main`:** `native-product-pages` is a long-lived integration branch for this multi-PR feature — each slice (data adapter, forecast freshness, observations, warnings, E2E…) lands as its own small PR into it for incremental review and a single shared feature preview, and the whole branch merges to `main` as a unit when the feature is complete. This is the foundational first slice.

## Related Issues

First slice of the **native-product-pages** feature. Subsequent slices (data-layer adapter, forecast freshness, historical date picker, observations, warnings, E2E) land as their own PRs into `native-product-pages`.

No linked GitHub issues yet. I just have this locally on my machine. It probably makes sense for me to convert these to GH issues once the tooling PRs (#1117, #1118, #1119) land. So I'll plan to do that once those are merged.  

## Key Changes

- **Flag gating** — both forecast routes branch on `getUseNativeForecasts(center)`; the widget path is untouched for flag-off centers (the main no-regression check).
- **Settings migration** — regenerated against the current schema (the brought-forward branch's snapshot predated recent migrations). Additive column; details below.
- **Sanitizer swap `isomorphic-dompurify` → `sanitize-html`** — forecast HTML is sanitized in **server** components, and DOMPurify needs jsdom, which breaks Next's server bundle at runtime. `sanitize-html` is pure-JS. External (other-domain) links now open in a new tab with `rel="noopener noreferrer"`.
- **4 merge-conflict files** — `services/nac/nac.ts`, the two forecast routes, `migrations/index.ts`; all additive, keeping `main`'s newer route config (ISR + OG image).
- **UI** — site-standard `container` width, flattened + fully-clickable zone cards, center-timezone dates (from NAC metadata), "Issued" labels.

## How to test

1. `pnpm seed && pnpm dev`
2. Admin → Settings → **Features** → enable `useNativeForecasts` for a center (e.g. NWAC).
[...]

PR #1129, opened 2026-06-30
Native forecast pages — combined stack (reconcile + date picker + data layer + forecast page)

busbyk:
Combined review PR for the native forecast page work. Folds in the previously-stacked PRs so the interdependent changes are reviewed together:

- **#1121** — reconcile forecast page from `replace-widgets` + initial polish
- **#1126** — forecast date picker (historical browse / season replay)
- **#1127** — per-product source adapter, normalized model, two-control split
- **this branch** — native forecast page parity + freshness
- **+ pinch-zoom** in the forecast media lightbox (`ZoomablePhoto`, reusing the Gallery block's `react-zoom-pan-pinch` config)

**Base:** `native-product-pages` (the long-lived integration branch, kept current with `main`). Green on `pnpm tsc` / `lint` / `test` (619) / `drift:check`.

> Note: the two Payload migration snapshots (`src/migrations/*.json`, ~50k lines) dominate the diff — skip them in review.

Issue: `.scratch/native-product-pages/issues/03-native-forecast-page.md` · Build plan: `.scratch/native-product-pages/handoffs/03-native-forecast-page-handoff.md`

## Done
- [x] **ISR / unfreeze** — all-zones route moved off bare `force-static` to `revalidate=300`; single-zone lowered 1800→300. Forecast is never frozen at build time.
- [x] **Bottom Line above danger** — leads the view, matching the widget.
- [x] **Avalanche Problems (N)** heading + `Problem #i: {NAME}` cards with the four labeled columns (Problem Type / Aspect-Elevation / Likelihood / Size) and the example photo inline-right with its caption.
- [x] **Danger section parity** — today/tomorrow side-by-side **date-headed** columns (stack ≤375px), per-band `"N - Name"` labels with icons, **Elevation Band Descriptions** disclosure, and the **danger-scale legend** (1–5 strip + expandable definitions + explainer link).
- [x] **Validity-date banner** — archived on dated views, expired on live (absolute-instant comparison); withdrawn deferred (no v2 model flag).
- [x] Sanitization already wired everywhere; all-zones grid already complete.
- [x] **Media lightbox pinch-zoom** — pinch/scroll/click-to-zoom on forecast photos; carousel drag disabled so panning a zoomed photo doesn't trigger swipe (arrow keys / prev-next still navigate).

## Remaining (next pass — carry live-data validation)
- [ ] **Weather Summary inline** — Weather product schema/model/fetcher/v2 source (verbatim port from `../avy`), `WeatherTable`/`WeatherTableV1` shape-detected components, mount between discussion and media (live page only). *Gate: validate `/v2/public/product/{id}` for both table formats.*
- [ ] **Revalidate-on-view freshness** — `fetchForecastConditional` (no-store, If-None-Match, tolerate 304) + fix the `next.tags` spread bug + tag `fetchForecast`; `getForecastConditional` with server-derived fingerprint ETag; `api/[center]/forecast-freshness` route handler; `ForecastFreshness.client` (live-only) → `router.refresh()`; no-skip invariant. *Gate: validate live ETag/304 behavior.*

## Notes / decisions
- Bottom Line ordering = above danger by default (matches widget); revertable block-swap pending NWAC/Aaron confirmation.
- Glossary tooltips deferred to issue 06 (weather-table help is static, not glossary).
- Migrations recreated per `docs/coding-guide.md` after merging main: the two Settings-flag migrations are combined into `20260701_224854_native_products_flags` with a main-inclusive snapshot.


busbyk:
**Self-review:** ran an adversarial multi-agent review over the full diff (4 dimensions — freshness/caching, weather rendering, danger/problems parity, React/Next — each finding independently refuted-or-confirmed). It surfaced **3 real issues, all in the revalidate-on-view freshness route**, now fixed in `eca53494`:

1. **[med] Transient upstream error blanked the forecast.** A 502 / bad JSON made the fresh fetch return `null`, which was treated as a change → purged the cache and refreshed the viewer into the "unable to load forecast" error page (worse than plain ISR). → `null` fresh is now a no-op `304`; ISR backstops it.
2. **[med] Unauthenticated cache-purge abuse.** The purge was gated on the client `If-None-Match`, so a bot could force `revalidateTag` every request. → purge is now **server-authoritative** (revalidate only when the fresh product differs from the cached one); the client header only decides whether that one viewer refreshes. Bounded to ~one upstream refetch per real correction.
3. **[low] In-place weather corrections weren't reflected** on a forecast-triggered refresh (`fetchWeatherProduct` was untagged). → tagged + revalidated alongside the forecast.

One raw finding (weather `selectTable` fallback misattributing a zone) was **refuted** by the verification pass and left as-is.

**Known low-severity limitations (by design):** weather-only in-place corrections and genuine forecast withdrawals fall back to the 300s ISR window rather than on-view (keeps the freshness path API-considerate and never blanks last-known-good on a transient blip). A live end-to-end `router.refresh()` smoke on a real correction is unexercised (off-season — no live corrections to trigger it).

Issue #1106, opened 2026-06-09
NWAC Data Portal


rustynwac:
Ensure NWAC Data Portal functionality exists

PR #1145, opened 2026-07-08
Add NWAC weather station data pages

rchlfryn:
## Description

Adds native weather-station data pages to AvyWeb for NWAC, recreating the nwac.us `/weatherdata/<station>/now/` tables (currently a WordPress plugin). Data-first React tables backed by SnowObs, fetched server-side so the API token never reaches the browser.

## Related Issues

Part of #1106 

## Key Changes

- **SnowObs service** (`src/services/snowobs`): server-side `wx/v1/station/data/timeseries` fetch (token via `SNOWOBS_TOKEN` env, required), zod-validated response, and a pure transform → render-ready tables — cumulative precip, cross-station full-outer-join on timestamp, UTC→`America/Vancouver`, short sensor/unit labels. Unit-tested.
- **Station registry** (`src/constants/weatherStations.ts`): 32 station groups ported from the legacy `station_group_tables_config` — `slug`, `legacySlug` (for future redirects), `region`, `stids`, `columns`. Integrity-tested.
- **Detail route** `/weather/stations/[station]`: ISR (10-min revalidate), **24h / 7-day range toggle**, latest-observation timestamp + stale flag (>2h), per-station error isolation, gated to the NWAC tenant (`notFound()` for other centers).
- **Index** `/weather/stations`: region-grouped directory (up to 4 columns) + station picker.
- **`StationNowTable`**: shadcn table — zebra rows, sticky time column, tightened/centered columns, stacked label/unit/elevation headers.
- **NWAC-only "Weather Data" nav item** in the Weather dropdown (seed + provisioning).

## How to test

- `pnpm dev`, then load `nwac.localhost:3000/weather/stations` and `/weather/stations/hurricane-ridge`.
- Toggle **Last 24 Hours / Last 7 Days**; spot-check values against nwac.us.
- Confirm other centers (dvac/sac/snfac) `404` on these routes.
- `pnpm test` — covers the SnowObs transform and the registry.

## Screenshots / Demo video

https://github.com/user-attachments/assets/fdf8fb05-aacb-407f-8303-c89a2affe6f8

## Migration Explanation
None needed - can add page via Built In Pages

- Region "area" pages with historical date-range browsing (prototyped separately, deferred).
- Data Portal graphs (legacy D3 → React charts).
- nwac.us cutover: redirects from `/weatherdata/<slug>/now/` → new URLs (deferred; `legacySlug` is stored for this).


src/services/snowobs/constants.ts
```diff
@@ -0,0 +1,48 @@
...
+  '%': '%',
+  'W/m**2': 'W/m²',
+  volt: 'V',
+}
+
+// Variable name of the computed running-total precipitation column.
+export const PRECIP_HOURLY = 'precip_accum_one_hour'
+export const PRECIP_CUMSUM = 'precip_cumsum'
+
+export const DISPLAY_TIMEZONE = 'America/Vancouver'
```
busbyk: I assume this is hardcoded because we're only using this service for NWAC. But I wonder if we would use this service for other centers in the future and this would become an issue? 

Does the SnowObs timeseries endpoint only return stations for NWAC?

src/services/snowobs/snowobs.ts
```diff
@@ -0,0 +1,103 @@
...
+
+type FetchOptions = {
+  windowHours?: number
+  revalidate?: number
+}
+
+function buildTimeseriesUrl(stids: string[], windowHours: number, bucketSeconds: number): string {
+  const token = process.env.SNOWOBS_TOKEN
+  if (!token) {
+    throw new SnowObsError('SNOWOBS_TOKEN environment variable is not set')
```
busbyk, line 37: I could see the argument for this being a generic `Error` instead of a `SnowObsError` specifically but I assume that the consumer of `fetchStationTimeseries` expects a `SnowObsError` so this probably makes sense. Just thinking out loud.

src/services/snowobs/tableHelpers.ts
busbyk, line 1: I wasn't sure what this file would be from the name `transform.ts`. That's not necessarily a problem since yea the functions in this file are pretty much transforming data. But they're also all table helpers. Something along the lines of table helpers might make more sense. But I don't feel strongly about it.

busbyk (commented):
I focused on the SnowObs service like you requested and I think it looks great!

src/services/snowobs/constants.ts
rchlfryn: Good question, the endpoint itself isn't NWAC-specific. Every request we build is explicitly scoped to NWAC. We pass `source=nwac` plus the exact station id (stid) list from the station registry, and `SNOWOBS_TOKEN` is NWAC's token. So we only ever get back the stations we ask for. 
 
The hardcoding of the timezone is needed for this PR. From the legacy nwac.us pages, and `America/Vancouver` is what the old plugin used. I've renamed the NWAC-specific constants: `NWAC_SOURCE`, `NWAC_DISPLAY_TIMEZONE`, `NWAC_WEATHER_STATION_GROUPS`, and `NWAC_STATION_REGIONS`. The service, transform, and components are all center-agnostic, so if another center wants these pages later the refactor is easy, promote those constants to per-tenant config keyed off the tenant slug (the routes are already tenant-gated) and add `SNOWOBS_TOKEN` to the tenant's settings. I'd rather not build that abstraction until a second center actually needs it.

src/services/snowobs/tableHelpers.ts
rchlfryn: I like tableHelpers better. Updated

src/services/snowobs/constants.ts
rchlfryn: Opened #1169 to track making these pages center dynamic.

rustynwac: For sure want to think about this for other centers. I'll comment on #1169

Issue #1179, opened 2026-08-07
Convert Claude GitHub Actions workflows to subscription auth (CLAUDE_CODE_OAUTH_TOKEN)


busbyk:
## Summary

Our Claude-powered GitHub Actions currently authenticate with a token-usage-billed `ANTHROPIC_API_KEY`. We'd rather run them on the organization's Claude subscription via a Claude Code OAuth token (`claude setup-token` → `CLAUDE_CODE_OAUTH_TOKEN` secret), which the Claude Code CLI and `claude-code-action` both support. Subscription tokens only work through Claude Code — they cannot authenticate raw Messages API calls — which shapes the work below.

## Affected workflows

There are three Claude consumers in `.github/workflows/`:

1. **`claude.yml`** (`@claude` mentions) — trivial conversion. `anthropics/claude-code-action@v1` accepts `claude_code_oauth_token` in place of `anthropic_api_key`.
2. **`pr-visual-recap.yml` → `size-gate` job** (added in #1178) — needs a rework. It currently calls the Messages API directly with `claude-sonnet-5` to judge borderline PRs. Convert it to shell out to the Claude Code CLI instead (`npx -y @anthropic-ai/claude-code@2 -p "<gate prompt>" --model claude-sonnet-5 --output-format json`) with `CLAUDE_CODE_OAUTH_TOKEN` in the env, keeping the existing structured recap/reason output and the numeric fallback when the call fails.
3. **`pr-visual-recap.yml` → `visual-recap` job** — blocked upstream. It calls the reusable workflow `BuilderIO/agent-native/.github/workflows/pr-visual-recap-reusable.yml@main`, whose secrets contract only declares `ANTHROPIC_API_KEY`: its gate refuses to run when that secret is empty, and the Claude step env only passes that variable to the CLI. Reusable workflows can't receive secrets outside their declared interface. Options:
   - **(preferred)** File an issue/PR on [BuilderIO/agent-native](https://github.com/BuilderIO/agent-native) to add an optional `CLAUDE_CODE_OAUTH_TOKEN` secret (pass it in the Claude step env; accept it as an alternative to `ANTHROPIC_API_KEY` in the gate's secret-presence check). Small patch, keeps us on the auto-updating `@main` reference.
   - **(fallback)** Vendor the full workflow copy into this repo and patch it ourselves, at the cost of losing upstream fixes.

## Tasks

[...]

PR #1180, opened 2026-08-07
Authenticate the Claude Code workflow with a subscription OAuth token

busbyk:
## Description

Switches `.github/workflows/claude.yml` from the token-usage-billed `ANTHROPIC_API_KEY` to `claude_code_oauth_token`, so `@claude` mentions run on the organization's Claude subscription instead. `anthropics/claude-code-action@v1` declares `claude_code_oauth_token` as an explicit alternative to `anthropic_api_key` and forwards it to the CLI as `CLAUDE_CODE_OAUTH_TOKEN`, so this is a one-line swap.

**This cannot merge until the `CLAUDE_CODE_OAUTH_TOKEN` repo secret exists** — it does not today. Mint it with `claude setup-token` from an account on the org subscription (prefer a shared account over a personal one, so the secret survives personnel changes) and add it first, or `@claude` will fail with no credentials.

## Related Issues

Part of #1179. The other two consumers named in that issue are deferred: the `size-gate` rework has nothing to apply to until #1178 merges, and the `visual-recap` job is blocked upstream because `BuilderIO/agent-native`'s reusable workflow declares no `CLAUDE_CODE_OAUTH_TOKEN` secret. Both are documented with the specific upstream touchpoints in [this comment](https://github.com/NWACus/web/issues/1179#issuecomment-5221044965).

## Key Changes

- `.github/workflows/claude.yml` — `anthropic_api_key` → `claude_code_oauth_token`
- `docs/claude-github-issues.md` — troubleshooting section now names the new secret
- `drift.lock` — binds that doc to `claude.yml` so drift flags it if the workflow's auth changes again

## How to test

Note that `issues` / `issue_comment` workflows only ever run from the **default branch**, so this branch's `claude.yml` cannot be exercised by an `@claude` mention pre-merge. Test in this order:

1. Validate the token out-of-band before it goes anywhere near CI: `env -u ANTHROPIC_API_KEY CLAUDE_CODE_OAUTH_TOKEN='<token>' claude -p 'reply with OK'`. Catches a wrong account, an expired token, or a missing subscription.
2. Add the `CLAUDE_CODE_OAUTH_TOKEN` repo secret.
3. Optional end-to-end check before merge: temporarily add a `push`-triggered scratch workflow on this branch (push events *do* run workflows from the pushed ref) that invokes `anthropics/claude-code-action@v1` with the same input, then delete it before merge.
4. After merge, `@claude` on a throwaway issue and confirm the run authenticates in the Actions tab.

Low risk either way — the failure mode is "`@claude` stops responding," not a production impact, and revert is one commit.

`ANTHROPIC_API_KEY` is intentionally left in repo secrets: #1178's `pr-visual-recap.yml` still references it and has no other credential until the upstream piece lands. One known minor behavior change — the action's inline-comment classifier reads only `ANTHROPIC_API_KEY`, so unconfirmed buffered comments now post unclassified instead of being filtered. Comments Claude posts with `confirmed: true` (the normal review path) are unaffected, and the fallback is the action's own documented pre-buffering behavior.


busbyk:
Ran the step-3 smoke test on this branch and it passed, so the subscription token is verified end-to-end from CI.

A temporary `push`-triggered workflow (added in 64d7336c, removed in 10ba6f18 — not in the net diff) invoked the Claude Code CLI with `CLAUDE_CODE_OAUTH_TOKEN` as the **only** credential in the step env. [Run 31213153669](https://github.com/NWACus/web/actions/runs/31213153669) went green in 9s with a real model response:

```
subtype          success
is_error         False
result           OAUTH OK
num_turns        1
duration_ms      1907
```

That establishes the two things worth knowing before merge: the token authenticates a real model call from a Blacksmith runner, and it does so with no `ANTHROPIC_API_KEY` present. Since `action.yml` forwards the `claude_code_oauth_token` input straight to that same env var, the remaining risk on this PR is close to nil.

Still post-merge-only: the actual `@claude` trigger path, since `issues` / `issue_comment` workflows run from the default branch. Worth an `@claude` on a throwaway issue right after merge.

Issue #1169, opened 2026-07-27
Make weather station data pages center-dynamic


rchlfryn:
## Context

#1145 added native weather-station data pages (`/weather/stations` and `/weather/stations/[station]`), currently gated to the NWAC tenant. The SnowObs service, table transform (`tableHelpers.ts`), and all `WeatherStations` components are center-agnostic; the NWAC coupling lives entirely in deliberately named constants. Follow-up from review discussion: https://github.com/NWACus/web/pull/1145#discussion_r3657858600

## What's hardcoded today

- `NWAC_SOURCE = 'nwac'` (`src/services/snowobs/snowobs.ts`) — the `source` param sent to SnowObs
- `NWAC_DISPLAY_TIMEZONE = 'America/Vancouver'` (`src/services/snowobs/constants.ts`)
- `NWAC_WEATHER_STATION_GROUPS` / `NWAC_STATION_REGIONS` (`src/constants/weatherStations.ts`) — the station registry: 32 station groups with slugs, regions, SnowObs `stid`s, and column configs
- `STATIONS_TENANT_SLUG = 'nwac'` — routes `notFound()` for other tenants
- `SNOWOBS_TOKEN` — single env token (NWAC's)

## Proposed direction

Promote the per-center pieces to tenant-scoped config:

- [ ] Store `snowobsSource` and display timezone per tenant (don't assume tenant slug == SnowObs source name)
- [ ] Move the station registry (groups, regions, columns) to per-tenant config — likely a Payload collection so centers can manage their own stations, or per-tenant constants as an interim step
- [ ] Decide token strategy: shared `SNOWOBS_TOKEN` vs per-tenant secrets (depends on how SnowObs scopes tokens/accounts)
- [ ] Replace the `STATIONS_TENANT_SLUG` gate with "tenant has station config" (404 when none)
- [ ] `generateStaticParams` over all tenants with station config

## Non-goals / notes

- No behavior change for NWAC; its pages should render identically after the refactor.
- Not needed until a second center actually wants these pages — the current hardcoding is deliberate (1:1 port of legacy nwac.us pages).
- Legacy redirect work (`legacySlug`) is tracked separately in #1106.

rchlfryn:
Design pass on #1195 (Dennis's station graph/table feedback) pulled this forward — there are now concrete NWAC-side needs for the config this issue scopes, not just the "second center wants these pages" trigger in the non-goals.

Four things from #1195 need admin-editable config and are deferred here:

[...]

PR #1202, opened 2026-08-21
POC: Add weather station collections

rchlfryn:
## Description

Models the station registry currently living in `src/constants/weatherStations.ts` as Payload collections, so the pages can become center-dynamic. **Nothing reads these collections yet** — this is inert on its own, and the page cutover follows separately.

The registry describes **pages, not stations**: 17 of the 32 groups cover more than one logger, and the layout is a property of the combination rather than of any one station (Alpental lists Summit, Mid and Base temperatures together before moving on to humidity). So groups stay editorial and carry the table layout, while stations sync from SnowObs — which has no concept of a page, a slug or a column order.

## Related Issues

First of three for #1169. Then the page cutover, then alerting.

## Key Changes

- **`stations`** — synced from SnowObs `station/metadata`, fully read-only. Includes `weatherStationPartner` (who owns the site), which Django carried and we'd otherwise have dropped.
- **`stationGroups`** — the pages: slug, legacy slug, region, the stations the page covers, and the table layout as **one row per reading** naming the loggers that report it. That is where the 275 hand-ordered `[stid, variable]` pairs go, transposed into 242 rows — `air_temp → Summit, Mid, Base` instead of three separate entries. No station is on two pages, so nothing is configured twice.
- **`stationRegions`** — index headings, ranked north to south. A collection rather than a select because that order matters; Django models the same thing as `Area.order`.
- **A SnowObs tab on `Settings`** — source, token, a sync button and the pages switch. `weatherPagesEnabled` will replace the hardcoded `STATIONS_TENANT_SLUG` gate. Settings is already the per-tenant singleton, so this avoids a one-row collection that reads oddly in the admin. The token is encrypted with `payload.encrypt`, the same mechanism holding the MCP plugin's API keys, and field access is gated to global roles.
- **`/api/cron/sync-stations`** — hourly, per tenant, upserting on `(tenant, source, stid)`. Never deletes: groups point at these rows, and a station going quiet is a fault to report rather than a reason to forget it.
- **`POST /api/settings/:id/sync-stations`** — the same sync on demand, behind the button, so a new source or token can be checked the moment it's saved instead of an hour later.
- **`pnpm seed:weather`** — local-only. See below.

## How to test

```bash
pnpm seed                     # tenants + content, schema pushed
pnpm seed:weather             # regions, stations, groups, SnowObs config
pnpm dev                      # Weather → Station Groups → Alpental Ski Area
```

Verified locally against live SnowObs:

- **30 of 32 groups reproduce their legacy column layout exactly**, checked by seeding every group through `payload.create` and reading it back at `depth: 2`. The two that differ hold the identical columns in a different order: Alpental and Crystal Green Valley interleaved a pair of snow readings per station, which one row per reading cannot express.
- 242 column rows across the 32 pages, down from 275 flat `[stid, variable]` pairs.
- Sync on a seeded database: `created 0, updated 0, unchanged 60` in 410ms — no-op writes are skipped.
- Corrupting a station's name and elevation, then re-running: `updated 1, unchanged 59`, values restored.
- The sync endpoint returns 401 unauthenticated, 403 for another center's settings, and 400 with a readable message when the source or token is missing.
- Token confirmed stored as ciphertext, not plaintext.

## Migration Explanation

One migration creates the station tables, adds three `snowobs_*` columns to `settings`, and seeds 14 regions, 60 stations and 32 groups. Additive throughout — `down()` drops the tables and the columns.

It holds its own copy of the seed rather than importing the constants, matching `20260505_045200_backfill_nav_builtin_pages`, so it keeps working once they're deleted. The station rows are a point-in-time snapshot; the sync corrects anything stale on its first run. If the `nwac` tenant is absent it logs and skips rather than failing the deploy.

**That last point is why `pnpm seed:weather` exists.** Locally the schema is pushed rather than migrated, and migrations that do run execute before `/next/seed` creates any tenant — so the backfill finds no `nwac` and skips. The script does the same work against a pushed database, building regions and groups from the constants and fetching stations through the real `syncStations()`. It's idempotent. Deployed environments are unaffected; the tenant already exists there.

- **Next:** page cutover — `generateStaticParams` and the index read from Payload, `NWAC_WEATHER_STATION_GROUPS` deleted. Then alerting.
- **60 stations, not 54.** SnowObs holds seven that no page shows: stid 15 (retired 2019) and six "5 minute" variants of Mission Ridge and Mt Hood Meadows. They report no battery voltage, so they're inert — but they're real and now visible.
- **The SnowObs tab is not seeded on deploy** — it holds a credential, so NWAC's is filled in by hand in the admin. The sync skips any center without a source and token.
- **Two pages reorder a column or two** (above). Worth a look from Dennis before the page cutover, since these are read daily.
- **`seed:weather` reads the legacy constants**, which the page cutover deletes. It'll need repointing then — most likely at a dump of the real data.
- Sync failures return 500 so a broken run surfaces rather than sitting in a log. Wiring that to Sentry is worth doing when alerting lands.
- The 27k-line migration `.json` is Payload's generated schema snapshot, the same as every other migration in the repo.


rchlfryn:
Closing - going another direction since not all of these collections are needed

PR #1231, opened 2026-08-26
Replace the release commenter with a batched script

rchlfryn:
## Description

The `🚀 This is included in version …` comments didn't post for v1.16.0. The `post-release` job failed 5 seconds in with:

```
##[error]HttpError: You have exceeded a secondary rate limit.
```

`apexskier/github-release-commenter` walks **every commit** in a release and fires a GraphQL query per commit to find its associated PRs. v1.16.0 had 197 commits since v1.15.1, so ~200 rapid-fire queries tripped GitHub's secondary rate limit during enumeration — before it posted a single comment.

It tracks commit count, not PR count, and the only two failures in the workflow's history are our two biggest releases:

| Release | Commits | PRs | Result |
| --- | --- | --- | --- |
| v1.13.0 | 242 | 20 | ❌ |
| v1.16.0 | 197 | 23 | ❌ |
| v1.12.0 | 124 | 11 | ✅ |
| v1.15.0 | 81 | 8 | ✅ |
| v1.14.0 | 60 | 11 | ✅ |
| v1.15.1 | 24 | 5 | ✅ |

Clean break between 124 and 197, and our releases keep growing. A retry wouldn't help — the burst is the same size on every attempt. This replaces the action with a script that asks GitHub for the PR list **once**, so the request count scales with PRs (~48 calls for v1.16.0) instead of commits.

## Related Issues

None.

## Key Changes

- **Ask once instead of per-commit.** `POST /releases/generate-notes` returns every PR in the release in a single request (it picks the previous release as the base automatically).
- **Still comments on linked issues.** The old action followed each PR to the issues it closes — for v1.16.0 that's 19 PRs **+ 9 linked issues = 28 targets**. The script follows `closingIssuesReferences` so we don't silently lose those.
- **Idempotent.** Each target is checked for an existing mention of the release URL before commenting, so re-runs and backfills can't double-post. The old action had no such guard.
- **Backfillable.** Added a `workflow_dispatch` trigger with a `tag` input, so a past release can be commented after the fact. The `release: published` event only fires once and can't be replayed, which is why v1.13.0 stayed empty for four months.
- Preserves the existing `skip-label: dependencies` behavior and the comment text verbatim.

Dropping the action also drops an unmaintained dependency — `apexskier/github-release-commenter` is at v1.4.1 (Jan 2026), has no throttle or retry input, and still targets Node 20 (already warning as deprecated in our logs).

Following that Node 20 thread: a second commit bumps **every** action in `.github/workflows/` to its lowest Node 24 major (two, `peter-evans/find-comment@v2` and `create-or-update-comment@v3`, were still on Node 16), verified against the live Blacksmith runner version 2.336.0.

## How to test

The logic was verified end-to-end against both failed releases by running the step body locally:

- v1.16.0 → 19 PRs → 28 comment targets
- v1.13.0 → 20 PRs → 26 comment targets

**Both releases have already been backfilled** — those comments are live now, posted from a local run, so they're authored by @rchlfryn rather than `github-actions[bot]`. Every other release's comments stay bot-authored.

To verify the guard rails on this branch:

```bash
# extract the step body, then:
REPO=NWACus/web TAG=v1.16.0 GH_TOKEN=$(gh auth token) bash step.sh
# every line should read "#N already mentions v1.16.0." — no new comments
```

Once merged, `gh workflow run post-release.yml -f tag=<tag>` backfills any release ( `workflow_dispatch` only appears for workflows on the default branch).

## Screenshots / Demo video

N/A — CI-only change.

## Migration Explanation

None needed — no schema changes.

- **One unverified detail:** the job now needs `contents: read` for the `generate-notes` call, which the previous `permissions:` block didn't grant. GitHub's REST docs don't state the fine-grained permission for that endpoint, so if the first real run 403s, bump it to `contents: write`. Everything else in the script is proven against live data.
- The next release is the real test. If it posts clean, this is settled; if not, the failure will be loud and specific rather than a silent secondary-limit abort.

---
<a href="https://app.blacksmith.sh/NWACus/codesmith/web/pr/1231"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790362697&installation_model_id=426131&pr_number=1231&repository=NWACus%2Fweb&return_to=https%3A%2F%2Fgithub.com%2FNWACus%2Fweb%2Fpull%2F1231&signature=fe60030051132ce146a1ce0c2118aa2c3f957151dca37a488a3c8a3195d371e8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -3,16 +3,78 @@ on:
   release:
     types:
       - published
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Release tag to comment for (backfill a past release, e.g. v1.13.0)
+        required: true
+        type: string
 
 jobs:
   post_release:
     permissions:
+      contents: read
       issues: write
       pull-requests: write
     runs-on: blacksmith-2vcpu-ubuntu-2404
     steps:
-      - uses: apexskier/github-release-commenter@v1
-        with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          comment-template: |
-            🚀 This is included in version {release_link}
+      - name: Comment on the PRs and issues in this release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          TAG: ${{ github.event.release.tag_name || inputs.tag }}
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          release_url="https://github.com/$REPO/releases/tag/$TAG"
+          comment="🚀 This is included in version [$TAG]($release_url)"
+          owner="${REPO%/*}"
+          name="${REPO#*/}"
+
+          # The PRs in this release, straight from GitHub's own release-notes
+          # generator (it picks the previous release as the base). One request
+          # for the whole list -- the action this replaced walked every commit
+          # individually and tripped the secondary rate limit on big releases.
+          prs=$(gh api "repos/$REPO/releases/generate-notes" -f tag_name="$TAG" --jq .body \
+            | grep -oE '/pull/[0-9]+' | grep -oE '[0-9]+' | sort -un || true)
+
+          if [ -z "$prs" ]; then
+            echo "No pull requests found for $TAG."
+            exit 0
+          fi
+
+          # Comment on each PR and on the issues it closes. A number that isn't
+          # a PR resolves to null and is skipped.
+          targets=""
+          for pr in $prs; do
+            meta=$(gh api graphql -f query="
+              query {
+                repository(owner: \"$owner\", name: \"$name\") {
+                  pullRequest(number: $pr) {
+                    labels(first: 20) { nodes { name } }
+                    closingIssuesReferences(first: 20) { nodes { number } }
+                  }
+                }
+              }" --jq '.data.repository.pullRequest')
+
+            [ "$meta" = "null" ] && continue
+
+            if echo "$meta" | jq -e '.labels.nodes[] | select(.name == "dependencies")' >/dev/null; then
+              echo "Skipping #$pr (dependencies)."
+              continue
+            fi
+
+            targets="$targets $pr $(echo "$meta" | jq -r '.closingIssuesReferences.nodes[].number')"
+          done
+
+          for target in $(echo "$targets" | tr ' ' '\n' | sort -un); do
+            # Idempotent, so a re-run or a backfill can't double-comment.
+            if gh api "repos/$REPO/issues/$target/comments" --paginate --jq '.[].body' \
+              | grep -qF "$release_url"; then
+              echo "#$target already mentions $TAG."
+              continue
+            fi
+            gh api "repos/$REPO/issues/$target/comments" -f body="$comment" --silent
+            echo "Commented on #$target."
+          done
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,11 +19,11 @@ jobs:
     runs-on: blacksmith-2vcpu-ubuntu-2404
     steps:
       - name: 🏗 Setup repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
       - name: 🏗 Setup pnpm
-        uses: pnpm/action-setup@v4
+        uses: pnpm/action-setup@v5
       - name: 🏗 Setup Node
-        uses: actions/setup-node@v4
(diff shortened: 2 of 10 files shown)


.github/workflows/post-release.yml
```diff
@@ -3,16 +3,78 @@ on:
...
+          comment="🚀 This is included in version [$TAG]($release_url)"
+          owner="${REPO%/*}"
+          name="${REPO#*/}"
+
+          # The PRs in this release, straight from GitHub's own release-notes
+          # generator (it picks the previous release as the base). One request
+          # for the whole list -- the action this replaced walked every commit
+          # individually and tripped the secondary rate limit on big releases.
+          prs=$(gh api "repos/$REPO/releases/generate-notes" -f tag_name="$TAG" --jq .body \
+            | grep -oE '/pull/[0-9]+' | grep -oE '[0-9]+' | sort -un || true)
```
busbyk: The `|| true` swallows failures from `gh` itself, not just an empty result. I stubbed a 403 on `generate-notes` and the job prints "No pull requests found for v1.17.0." and exits 0 — green, no comments posted. That's the same symptom we're fixing here, so if the `contents: read` guess is wrong we won't find out from the run status.

Splitting the fetch from the parse keeps the loud failure and still handles a genuinely empty release:

```bash
notes=$(gh api "repos/$REPO/releases/generate-notes" -f tag_name="$TAG" --jq .body)
prs=$(printf '%s' "$notes" | grep -oE "$REPO/pull/[0-9]+" | grep -oE '[0-9]+$' | sort -un || true)
```

Scoping the grep to `$REPO` also stops a PR title that links some other repo's `/pull/123` from turning into a target here.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/post-release.yml
```diff
@@ -3,16 +3,78 @@ on:
...
+              continue
+            fi
+
+            targets="$targets $pr $(echo "$meta" | jq -r '.closingIssuesReferences.nodes[].number')"
+          done
+
+          for target in $(echo "$targets" | tr ' ' '\n' | sort -un); do
+            # Idempotent, so a re-run or a backfill can't double-comment.
+            if gh api "repos/$REPO/issues/$target/comments" --paginate --jq '.[].body' \
+              | grep -qF "$release_url"; then
```
busbyk: This can post duplicates. `grep -q` exits on first match, so if `gh` writes anything after that it dies on SIGPIPE, `pipefail` propagates 141, the `if` goes false and we comment on a target that already has the comment. Reproduced with a stub whose comment list already contained the release URL — it posted anyway.

It's a race, so it passes on small threads (which is why your v1.16.0 run showed "already mentions" every line) and breaks when `gh --paginate` fetches a second page, i.e. anything over 30 comments. Long-lived issues that collect a comment per release are exactly the ones that hit it.

```bash
existing=$(gh api "repos/$REPO/issues/$target/comments" --paginate --jq '.[].body')
if grep -qF "$release_url" <<<"$existing"; then
```

---
_Generated by [Claude Code](https://claude.ai/code)_

busbyk (changes requested):
Couple issues found via Claude review

.github/workflows/post-release.yml
rchlfryn: Fixed in [e619501](https://github.com/NWACus/web/pull/1231/commits/e6195014ac9bfcc897c810e4691005ade040a952)

.github/workflows/post-release.yml
rchlfryn: Fixed in [e619501](https://github.com/NWACus/web/pull/1231/commits/e6195014ac9bfcc897c810e4691005ade040a952)

busbyk (approved):

